PR #720, opened 2016-04-20
Switching back to using tags in email alert payload

christophwong:
[TRELLO](https://trello.com/c/ploRXLJV)
As of now email-alert-api does not support a way without using "tags" to get a
multi-document subscriber list (subscribed to all /cma-cases changes as
opposed to one specific document as seen with travel advise publisher's
usage) from content-ids/links.

Leaving links intact because a.) it doesn't affect the payload in initial testing, b.) tags are being deprecated and we hope to use just links in the future.

I have tested this change in payload format by using 

```
curl -X POST -H "Content-Type: application/json" -d <actual payload> https://email-alert-api.integration.publishing.service.gov.uk/notifications
```

The payload where marked <actual payload> in the above example is as follow with the HTML replaced with <shortened> (but for context the actual html string was used). Focus is on the `"tags"` attribute

```
{"subject": "test with b", "body": "<shortened>", "links": {"topic": ["e033acc7-9034-4f01-90f4-ca6a49952194"]}, "tags": {"format": "cma_case", "opened_date": "2016-04-10", "case_type": "ca98-and-civil-cartels", "case_state": "open", "market_sector":["something"]}, "document_type": "cma_case", "header": "<shortened>", "footer": "<shortened>"}
```

In the Tech Debt session of email-alert-api govuk docs, regarding the new links architechure, one of the future goals is "Resolving how specialist document alerts would work alongside such an architecture"
Reference: https://gov-uk.atlassian.net/wiki/display/TECH/Email+Alert+API

### DIFF
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -25,16 +25,9 @@ private
   end
 
   def tags
-    { format: document.publishing_api_document_type }.deep_merge(metadata)
-  end
-
-  def metadata
-    merged_fields = document.format_specific_fields.map { |f|
-      {
-        f => document.send(f)
-      }
-    }.reduce({}, :merge)
-    merged_fields.reject { |_k, v| v.blank? }
+    {
+      format: document.publishing_api_document_type
+    }.deep_merge(document.format_specific_metadata.reject { |_k, v| v.blank? })
   end
 
   def standard_body(template_path)
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -9,7 +9,7 @@ class EmailAlertPresenter
     {
       subject: document.title + " updated",
       body: body,
-      links: { topics: [document.content_id] },
+      tags: tags,
       document_type: document.publishing_api_document_type,
     }.merge(extra_options)
   end
@@ -22,6 +22,19 @@ private
     else
       standard_body("email_alerts/publication")
     end
+  end
+
+  def tags
+    { format: document.publishing_api_document_type }.deep_merge(metadata)
+  end
+
+  def metadata
+    merged_fields = document.format_specific_fields.map { |f|
+      {
+        f => document.send(f)
+      }
+    }.reduce({}, :merge)
+    merged_fields.reject { |_k, v| v.blank? }
   end
 
   def standard_body(template_path)
--- a/spec/models/cma_case_spec.rb
+++ b/spec/models/cma_case_spec.rb
@@ -68,8 +68,12 @@ RSpec.describe CmaCase do
 
   let(:email_alert_payload) do
     {
-      "links" => {
-        "topics" => [cma_cases[0]["content_id"]]
+      "tags" => {
+        "format" => "cma_case",
+        "opened_date" => "2014-01-01",
+        "case_type" => "ca98-and-civil-cartels",
+        "case_state" => "open",
+        "market_sector" => ["energy"]
       },
       "document_type" => "cma_case"
     }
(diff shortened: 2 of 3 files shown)


mathildathompson:
👍 Looks good to me. Think it would be useful for someone (maybe @elliotcm ) with a bit more context of email-alert-api to double check.

app/presenters/email_alert_presenter.rb
```diff
@@ -24,6 +25,19 @@ def body
     end
   end
 
+  def tags
+    { format: document.publishing_api_document_type }.deep_merge(metadata)
+  end
+
+  def metadata
```
benilovj: I find this method a bit cryptic. Do the fields have to be dynamic here, or are only a few used (eg `format`, `case_type`)?

elliotcm:
I'd recommend removing the links because while I think it'll be safe, we work under the assumption that it'll be one or the other when triggering notifications.  Having both in subscriptions is OK.

app/presenters/email_alert_presenter.rb
christophwong: The reason I went with a dynamic approach was because for each document type, the tags (with the exception of format) are different. So I copied this pattern from document presenter.
`case_type` is only used for cma_cases, and for each document type, a different combinations of tags are used.

benilovj: IMHO, if it has to be dynamic, then this is fine, but if it could be static, an explicit list of fields would read better. @elliotcm any thoughts? I don't really have enough context here.

christophwong:
@elliotcm links has been removed.

app/presenters/email_alert_presenter.rb
elliotcm: You're both right!  Pulling them from the objects looks to be the way it's got to be done, but it is quite a hard to parse method.  @christophwong could you spend a short time seeing if it can be made clearer?

elliotcm:
@christophwong Thanks.

app/presenters/email_alert_presenter.rb
christophwong: Gladly, if we find a solution to this, we can also clean up the SearchPresenter and DocumentPresenter too.

I think I will use this [method](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/app/models/document.rb#L92) in my first step, (although it still has the same problem of lacking clarity, I still believe "metadata" belongs in the document model rather than the presenter.)

app/presenters/email_alert_presenter.rb
```diff
@@ -24,6 +24,12 @@ def body
     end
   end
 
+  def tags
+    {
+      format: document.publishing_api_document_type
+    }.deep_merge(document.format_specific_metadata.reject { |_k, v| v.blank? })
```
christophwong, line 30: @benilovj @elliotcm this is the refactoring (it has caused the previous discussion to be hidden) but the problem of clarity is still here.
However, I am leaning towards dynamic approach towards the metadata, as writing all the fields out for each specific format seemed messy. 

Also, in this system the attributes that are grouped into "format_specific_metadata" seems semantically accurate to me. I believe our app only needs to know them as a group of metadata, and there isn't specific information or behavioural difference between the specific fields inside this metadata group.

christophwong, line 30: similar pattern:
- [Document Presenter](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/app/presenters/document_presenter.rb#L51-L63)
- [SearchPresenter for rummager](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/app/presenters/search_presenter.rb#L16)

elliotcm, line 30: Seems OK to me.

christophwong, line 30: ok to merge?

elliotcm, line 30: @benilovj?

PR #725, opened 2016-04-22
Display 'published' for live specialist-documents on index page

christophwong:
Paired with @Rosa-Fox 
Previous fix for the show page calls `document#live?` in the `state` helper
method. However, the index page loads OpenStruct objects instead of
Document objects and will always return nil for `document#live?`.

Before:
pubstate_before

After:
publishstate_after


issyl0:
This still makes the "live" => "published" change work on the show pages, so it looks fine to me.

PR #728, opened 2016-04-25
Use pundit gem for authorisations

christophwong:
We wanted to use this gem because it allows us to format authorisation
levels in policies objects, which might be an improvement on readability
of permission rules.

This commit replaces the original "permitted?" method used in DocumentController with Pundit code.

Another benefit is that we are moving away from using the "helper methods" in application controller (#formats_user_can_access) and it helps clean up the code some more.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -6,17 +6,24 @@ class DocumentsController < ApplicationController
   include ActionView::Helpers::TextHelper
 
   before_action :fetch_document, only: [:edit, :show, :publish, :update]
-  before_action :permitted?, if: :document_type
+  before_action :check_authorisation, if: :document_type
+
+  after_action :verify_authorized
+
+  def check_authorisation
+    if current_format
+      authorize document_klass
+    else
+      flash[:danger] = "That format doesn't exist. If you feel you've reached this in error, contact your SPOC."
+      redirect_to manuals_path
+    end
+  end
 
   def index
     page = filtered_page_param(params[:page])
     per_page = filtered_per_page_param(params[:per_page])
-    if current_format
-      @response = document_klass.all(page, per_page)
-      @paged_documents = PaginationPresenter.new(@response, per_page)
-    else
-      redirect_to manuals_path
-    end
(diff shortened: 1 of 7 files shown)


app/controllers/documents_controller.rb
```diff
@@ -6,19 +6,26 @@ class DocumentsController < ApplicationController
   include ActionView::Helpers::TextHelper
 
   before_action :fetch_document, only: [:edit, :show, :publish, :update]
-  before_action :permitted?, if: :document_type
+  before_action :check_authorisation, if: :document_type
 
-  def index
-    page = filtered_page_param(params[:page])
-    per_page = filtered_per_page_param(params[:per_page])
+  after_action :verify_authorized
```
benilovj, line 11: We should enforce this in `ApplicationController` as suggested in the [pundit docs](https://github.com/elabs/pundit#ensuring-policies-and-scopes-are-used) to avoid forgetting to authorise...

app/policies/document_policy.rb
```diff
@@ -0,0 +1,52 @@
...
+  def gds_editor
+    user.gds_editor?
+  end
+
+  def index?
+    gds_editor || departmental_editor || writer
+  end
+
+  def new?
+    gds_editor || departmental_editor || writer
```
benilovj: You can condense the repetition by using `alias_method` (see the [signon code](https://github.com/alphagov/signonotron2/blob/master/app/policies/api_user_policy.rb) for an example).

benilovj:
This is looking encouraging...

app/controllers/documents_controller.rb
christophwong, line 11: This is correct, however, that will enforce pundit for Attachments, manuals, manuals attachments, manuals sections, and I'm not sure if we want to implement permission levels for all those other endpoints in this PR.

From offline discussion, I will implement this in the next PR.

benilovj, line 11: Ok, but then could we then do this on a subsequent PR? Having this enforcement and having a mechanism which catches forgotten authorisation is one of the biggest benefits of a framework like Pundit.

PR #732, opened 2016-04-28
Don't check for specific success status code when pushing to Publishing API and Rummager

davidbasalla:
As of https://github.com/alphagov/rummager/commit/805647a49d3a530338657bbfd6cb2ff1f432550d, synchronous publishing with Rummager has been removed, so always returns a 202 on success. This commit updates the checks in the publish methods as well as the test helpers to stub rummager. Presumably these were put in place for the development environment where synchronous rummager publishing was enabled.

We update the `publish!` method for drug safety updates in a separate PR (https://github.com/alphagov/specialist-publisher-rebuild/pull/727)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -207,13 +207,12 @@ class Document
       presented_links = DocumentLinksPresenter.new(self)
 
       begin
-        item_request = publishing_api.put_content(self.content_id, presented_document.to_json)
-        links_request = publishing_api.patch_links(self.content_id, presented_links.to_json)
+        publishing_api.put_content(self.content_id, presented_document.to_json)
+        publishing_api.patch_links(self.content_id, presented_links.to_json)
 
-        item_request.code == 200 && links_request.code == 200
+        true
       rescue GdsApi::HTTPErrorResponse => e
         Airbrake.notify(e)
-
         false
       end
     else
@@ -228,8 +227,8 @@ class Document
 
     begin
       update_type = self.update_type || 'major'
-      publish_request = publishing_api.publish(content_id, update_type)
-      rummager_request = rummager.add_document(
+      publishing_api.publish(content_id, update_type)
+      rummager.add_document(
         search_document_type,
         base_path,
         indexable_document.to_json,
@@ -239,7 +238,7 @@ class Document
         email_alert_api.send_alert(EmailAlertPresenter.new(self).to_json)
       end
 
-      publish_request.code == 200 && rummager_request.code == 200
+      true
     rescue GdsApi::HTTPErrorResponse => e
       Airbrake.notify(e)
       false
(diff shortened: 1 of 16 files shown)


app/models/document.rb
```diff
@@ -239,7 +239,7 @@ def publish!
         email_alert_api.send_alert(EmailAlertPresenter.new(self).to_json)
       end
 
-      publish_request.code == 200 && rummager_request.code == 200
+      publish_request.code == 200 && rummager_request.code == 202
```
benilovj: Being so prescriptive on the expected return type feels very brittle (eg if Rummager or Publishing API suddenly starts returning another 2XX status code, this will break). I believe that `gds-api-adapters` will raise an exception if there's a non-200 return code, so so is this line is even needed?

Thoughts? @danielroseman @tijmenb

tijmenb: Yes, feels very brittle. It would be better to rely on the exceptions here (assuming they're raised).

h-lame: GDS Api Adapters does behave like this so the check does feel redundant.  The real question is that now Rummager returns 202 "Thanks, I'll get round to it later" what API does it give us for checking that there weren't errors with that?  We're accepting the 202 and telling users that the document is published, but something might happen with that eventual search processing that means it's not in the search index - how do we let users know about this?  The same question could be levelled at publishing-api too as although it responds with a 200, it's also really a 202 because everything is async now isn't it?  

Should publishing-api respond with a 202?  Should rummager respond with a 200? Should there be a "what's the state of my thing?" api for both?  Is there a pattern for this sort of thing?

Apols for hi-jacking this PR somewhat.

tijmenb: Note that rummager has returned `202` in production for years (https://github.com/alphagov/rummager/pull/124 - July 2013). I've removed the special development behaviour recently, because I suspected it could cause problems for people using and testing things in dev (https://github.com/alphagov/rummager/pull/629). 

Seeing as above code would've suddenly and mysteriously failed in production (because `rummager_request.code == 200`), I'm happy about such a quick return on investment :)

As for the questions: I think we haven't properly discussed how to give feedback to users about the state of the publishing pipeline.

In an ideal world I would like to see every content item in publisher apps have a status indicator 🚦 for all async side-effects:
- published to content-store
- indexed in search
- email alerts sent

danielroseman: When we originally prototyped the phase 2 stuff @elliotcm did some work on async notifications, but we haven't had time to take it any further. It's definitely in the long-term roadmap; happy to have a discussion to thresh out how it should work.

elliotcm: @tijmenb 
async_publishing_prototype

tijmenb: @elliotcm @danielroseman That looks really really cool! Had no idea that existed.

h-lame: That does indeed seem like a very useful thing, thanks for sharing @elliotcm!  Glad this sort of stuff is on a roadmap, and for the avoidance of doubt I'm totally onboard with the suggested fix of removing the status code checks.

benilovj:
This looks good – could we possibly squash the commits into one though? The first commit is no longer relevant and the last two don't have to be distinct IMHO.

PR #733, opened 2016-04-29
Fix jenkins schema testing

tijmenb:
This probably fixes persistent errors on ci:

https://ci.dev.publishing.service.gov.uk/job/govuk_specialist-publisher-rebuild_schema_tests/50/console

This app should not be setting `GOVUK_APP_DOMAIN` in jenkins at all - it’s using a really old pattern here. See this for example: https://github.com/alphagov/collections-publisher/pull/128/commits/fbd02be7adc9ce616b537efc408a1c8a7e37fbc9


davidbasalla:
👍

PR #734, opened 2016-04-29
Ensure headers navigation is rendered for specialist-documents

davidbasalla:
Headers data is now sent to the publishing-api to ensure that headers navigation renders on specialist-frontend.

Paired with @davidbasalla.

`remove_empty_headers` method has been implemented to match specifications within content schema which enforce that ‘headers’ values must not be empty (https://github.com/alphagov/govuk-content-schemas/blob/master/formats/specialist_document/publisher/details.json#L72). Previously the GovSpeak `structured_headers` method was generating a key `headers` which would be an empty array unless it contained a nested header, causing contract tests to fail.

[Trello](https://trello.com/c/RO6QRqw9/96-headings-missing-on-specialist-documents-medium)

### DIFF
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -41,6 +41,19 @@ private
       change_history: change_history
     }.tap do |details_hash|
       details_hash[:attachments] = attachments if document.attachments
+      details_hash[:headers] = headers if !headers.empty?
+    end
+  end
+
+  def headers
+    headers = Govspeak::Document.new(document.body).structured_headers
+    remove_empty_headers(headers.map(&:to_h))
+  end
+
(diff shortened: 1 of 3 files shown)


app/presenters/document_presenter.rb
```diff
@@ -37,13 +37,26 @@ def to_json
   def details
     {
       body: GovspeakPresenter.present(document.body),
+      headers: headers,
```
h-lame: Should this be in the `.tap` so that we don't add it if there are no `headers` at all - I don't think we have spec for that scenario.

Rosa-Fox: Thanks @h-lame, very good point. I have made the changes and amended.

PR #736, opened 2016-05-03
Move specialist document titles into the document model, simplify ApplicationController

benilovj:
This means that more of the format-specific logic lives in the document
model, not centrally.

This PR also includes:
- a correction to 2 document format titles (EAT Decisions, ET decisions) from plural to singular, to fit the pattern of the other formats.
- a simplification of `ApplicationController` to reduce unnecessary indirection

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,33 +26,15 @@ private
   end
 
   def current_format
-    document_types.fetch(params.fetch(:document_type, nil), nil)
+    @current_format ||= document_types.detect { |format| format.document_type == document_type }
   end
 
   def formats_user_can_access
-    document_types.select { |_, v| policy(v.klass).index? }
+    document_types.select { |format| policy(format).index? }
   end
 
-  # This Struct is for the document_types method below
-  FormatStruct = Struct.new(:klass, :document_type, :format_name, :title, :organisations)
-
   def document_types
-    # For each format that follows the standard naming convention, this
-    # method takes the model class and generates a FormatStructure.
-    #
-    # eg GdsReport will become this:
-    #
-    # {
-    #   "gds-reports" => FormatStruct.new(
-    #     klass: GdsReports, # This is the class name of the model
-    #     document_type: "gds-reports", # This is internally used for building urls
-    #     format_name: "gds_report", # This is used for fetching the params of the format
-    #     title: "GDS Report", # Rendered as the format name to the User, taken from GdsReport.title
-    #     organisations: ["a-content-id"], # Content IDs for the Orgs fetched from the schema
-    #   )
-    # }
-
-    document_classes = [
+    [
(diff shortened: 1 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -42,48 +42,44 @@ def formats_user_can_access
...
-          v.organisations,
+        title.downcase.parameterize.pluralize => FormatStruct.new(
+          document_class,
+          title.downcase.parameterize.pluralize,
+          document_class.to_s.underscore,
+          title,
+          document_class.organisations,
         )
       }
     }.reduce({}, :merge)
```
h-lame: While we're tidying this up how about also:

1) memoizing it; this method isn't free to run and is probably used a bunch
2) getting rid of the `map` + `reduce` dance by using `.map.with_object`:

``` ruby
document_classes.map.with_object({}) do |document_class, everything|
  title = document_class.title
  thing = # build the FormatStruct instance
  everything[thing.document_type] = thing
end
```

When I read the `map` block and see it's followed by the `reduce` I have to spend a moment working out what it's doing and why, whereas it'd be upfront in the `map.with_object` case.  YMMV obvs.

I'm more interested in memoizing it than this TBH.

benilovj: I've done away with the entire `FormatStruct` business in 3ac24f8. Do you think memoizing is still necessary?

app/controllers/application_controller.rb
```diff
@@ -26,66 +26,27 @@ def document_type
   end
 
   def current_format
-    document_types.fetch(params.fetch(:document_type, nil), nil)
+    document_types.detect { |format| format.document_type == document_type }
   end
```
h-lame: Old `current_format` impl. was a hash lookup so not much to be gained by memoisation, but now it's iterating through an array to find something, so there will be tiny benefits to be gained for all requests for `VehicleRecallsAndFaultsAlert` documents ;)

app/controllers/application_controller.rb
```diff
@@ -26,66 +26,27 @@ def document_type
...
 
   def formats_user_can_access
-    document_types.select { |_, v| policy(v.klass).index? }
+    document_types.select { |format| policy(format).index? }
   end
```
h-lame, line 35: Might not hurt to memoize this while we're at it too (not sure how much it's used though, if it's just in the drop down for the header maybe it's of less use).

app/controllers/application_controller.rb
benilovj: done

app/controllers/application_controller.rb
benilovj, line 35: I don't think it's worth doing here, as it's only used in the dropdown.

h-lame, line 35: Fair enough.

PR #739, opened 2016-05-04
Add search box to document indexes

danielroseman:
Passes the q parameter on to publishing-api for filtering by title/slug.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -162,17 +162,8 @@ class Document
     @public_updated_at = Time.parse(timestamp.to_s) unless timestamp.nil?
   end
 
-  def self.all(page, per_page)
-    # The current version of this method is a result of the Publishing API
-    # returning the `details` field as an empty hash. As such, we get the
-    # content_id of all `specialist_document`s, then request the individual
-    # payload for each, which allows us to construct the real object.
-
-    # When the Publishing API is fixed and `details` is returned, this method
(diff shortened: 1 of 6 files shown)


app/models/document.rb
```diff
@@ -182,9 +173,10 @@ def self.all(page, per_page)
         :publication_state,
       ],
       page: page,
-      per_page: per_page
-    ).to_ostruct
```
h-lame, line 186: One less `OpenStruct` in the world! 🎏 🎆

spec/models/cma_case_spec.rb
```diff
@@ -117,6 +117,13 @@ def cma_case_content_item(n)
     it "returns all CMA Cases" do
       expect(described_class.all(page, per_page).results.length).to be(cma_cases.length)
     end
+
+    it "passes query parameter if supplied" do
+      expect(described_class.publishing_api).to receive(:get_content_items)
+        .with(hash_including(q: "foo"))
+        .and_return(double(to_ostruct: {}))
+      described_class.all(page, per_page, q: "foo")
+    end
```
h-lame, line 126: Do we also want to have a feature spec here to make sure the UI -> model part works? (I'm sure it does, but...).

benilovj, line 126: :+1:

danielroseman:
Added feature spec.

PR #740, opened 2016-05-04
Make publish! report failure for drug safety updates

boffbowsh:
If an error is raised while publishing a drug safety update, the
`publish!` method should return false in line with other `Document`
subclasses


benilovj:
Cheers. We really should extract the begin / rescue / notify airbrake behaviour from across the many places in the app.

PR #741, opened 2016-05-05
Wrap remote publishing calls

benilovj:
This helps have more consistent behaviour when pushing data to remote apps
(Publishing API, Rummager, Email Alert API).

/cc @boffbowsh

### DIFF
--- a/app/helpers/publishing_helper.rb
+++ b/app/helpers/publishing_helper.rb
@@ -1,0 +1,11 @@
+module PublishingHelper
+  def handle_remote_error(&block)
+    begin
+      block.call
+      true
+    rescue GdsApi::HTTPErrorResponse => e
+      Airbrake.notify(e)
+      false
+    end
+  end
+end
--- a/config/initializers/services.rb
+++ b/config/initializers/services.rb
@@ -3,6 +3,33 @@ require 'gds_api/rummager'
 require 'gds_api/asset_manager'
 require 'gds_api/email_alert_api'
 
+module Services
+  def self.publishing_api
+    GdsApi::PublishingApiV2.new(
+      Plek.new.find('publishing-api'),
+      bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example',
+    )
+  end
+
+  def self.rummager
+    GdsApi::Rummager.new(Plek.new.find('search'))
+  end
+
+  def self.asset_api
+    GdsApi::AssetManager.new(
+      Plek.current.find('asset-manager'),
+      bearer_token: ENV['ASSET_MANAGER_BEARER_TOKEN'] || '12345678'
+    )
+  end
+
+  def self.email_alert_api
+    GdsApi::EmailAlertApi.new(
+      Plek.current.find('email-alert-api'),
+      bearer_token: ENV['EMAIL_ALERT_API_BEARER_TOKEN'] || 'example123'
+    )
+  end
+end
+
 module SpecialistPublisher
   def self.register_service(name, service)
     @services ||= {}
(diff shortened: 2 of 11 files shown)


config/initializers/services.rb
```diff
@@ -10,8 +10,12 @@ def self.register_service(name, service)
...
   end
 
-  def self.services(name)
-    @services[name] || raise(ServiceNotRegisteredException.new(name))
+  def self.services(name = nil)
+    if name
+      @services[name] || raise(ServiceNotRegisteredException.new(name))
+    else
+      @services
+    end
```
h-lame, line 45: Not super keen on the method over-riding here, particularly as calling `services` without an argument services exposes the internal hash to the rest of the world, which we've gone to some effort to hide with `register_service`.  How about:

``` ruby
module SpecialistPublisher
  class Services
    def initialize
      @services = {}
    end
    def register_service(name, service)
      @services[name] = service
    end
    def service(name)
      @services[name] || raise(ServiceNotRegisteredException.new(name))
    end
  end

  def self.register_service(name, service)
     services.register_service(name, service)
  end

  def self.service(name)
    services.service(name)
  end

  def self.services
    @services ||= Services.new
  end
end

module PublishingHelper
  def remote_publish!(&block)
    begin
      block.call(services)
      true
    rescue GdsApi::HTTPErrorResponse => e
      Airbrake.notify(e)
      false
    end
  end

  def services
    SpecialistPublisher.services
  end
end

class Document
  def publish
    # other publish code
    remote_publish! do |services|
      services.service(:publishing_api).put_content(self.content_id, presented_section)
      services.service(:publishing_api).patch_links(self.content_id, presented_section_links)
    end
  end
end
```

TBH while that no longer exposes the internal storage, I can't say I'm a huge fan of `services.service(:service_name)`. I'd be in favour of storing our services in something that looks more like:

``` ruby
module Services
  def self.publishing_api
    GdsApi::PublishingApiV2.new(
      Plek.new.find('publishing-api'),
      bearer_token: ENV['PUBLISHING_API_BEARER_TOKEN'] || 'example',
    )
  end

  def self.rummager
    GdsApi::Rummager.new(Plek.new.find('search'))
  end

  def self.asset_api
    GdsApi::AssetManager.new(
      Plek.current.find('asset-manager'),
      bearer_token: ENV['ASSET_MANAGER_BEARER_TOKEN'] || '12345678'
    )
  end 

  def self.email_alert_api
    GdsApi::EmailAlertApi.new(
      Plek.current.find('email-alert-api'),
      bearer_token: ENV['EMAIL_ALERT_API_BEARER_TOKEN'] || 'example123'
    )
  end
end
```

Mostly on the grounds that method calls are better than hash key lookups, but also on the grounds that it looks more like other services storage in other govuk apps (see [contacts-admin](https://github.com/alphagov/contacts-admin/blob/master/lib/services.rb) and [collections](https://github.com/alphagov/collections/blob/master/lib/services.rb) for example).  This allows remote publishing to look like:

``` ruby
class Document
  def publish
    # other publish code
    remote_publish! do
      Services.publishing_api.put_content(self.content_id, presented_section)
      Services.publishing_api.patch_links(self.content_id, presented_section_links)
    end
  end
end
```

Which feels marginally nicer to use (note lack of yielding, but we could leave that in if we wanted to hide Services).

benilovj, line 45: agreed – I've changed `SpecialistPublisher#services` to `Services#publishing_api` etc, as you suggested.

This should be ready for another look.

app/helpers/publishing_helper.rb
```diff
@@ -0,0 +1,11 @@
+module PublishingHelper
+  def remote_publish!(&block)
```
boffbowsh: This name feels overloaded - it's not just being used for publishing (in terms of talking to the publishing api) but also for talking to Rummager and email-alert-api.

benilovj: I agree, although I'm kinda struggling to find something better... maybe something like `remote_write!` but again that feels very generic.

boffbowsh: The method isn't actually doing any remote anything, so maybe something like:

``` ruby
handle_remote_error do
```

benilovj: updated as suggested.

PR #742, opened 2016-05-05
Display bulk_published label value

Rosa-Fox:
Displays an associated boolean value alongside `bulk_published` labels for all formats.

Tests check that:
- the correct labels display.
- publishing a new edition of a document will not change the `bulk_published` field.

Please note - functionality to bulk publish documents has not yet been implemented within specialist-publisher-rebuild

[Trello](https://trello.com/c/GMrhLspk/79-missing-bulk-publication-field-for-specialist-documents-small)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -33,6 +33,10 @@ class Document
     end
   end
 
+  def bulk_published
+    @bulk_published || false
+  end
+
   def base_path
     @base_path ||= "#{finder_schema.base_path}/#{title.parameterize}"
   end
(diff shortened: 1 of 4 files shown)


benilovj:
Looks good. One minor point: it'd be good to have an assertion in a feature test somewhere which checks for `bulk published: false` when the flag isn't set...

Rosa-Fox:
Thanks @benilovj. Have added [here](https://github.com/alphagov/specialist-publisher-rebuild/blob/bulk-publish-label-values/spec/features/creating_a_cma_case_spec.rb#L49).

benilovj:
👍

PR #747, opened 2016-05-09
Incredibly basic DFID finder

rgarner:
- Add DFID to the list of finders we know about
- Add features that let us test that the finder does indeed appear and
  we can create things with it
- Specify a very basic schema that has a single facet: `country`
- Allows us to create and publish drafts

Finder menu item
New output
Finder-frontend list
Specialist-frontend item


lib/documents/schemas/dfid_research_outputs.json
```diff
@@ -0,0 +1,835 @@
...
+    {
+      "key": "published_date",
+      "name": "Published",
+      "short_name": "Published",
+      "type": "date",
+      "display_as_result_metadata": true,
+      "filterable": false
+    }
+  ]
+}
```
benilovj, line 836: missing newline here

benilovj:
This all seems pretty sensible.

benilovj:
Would you be able to attach some screenshots to the PR description?

lib/documents/schemas/dfid_research_outputs.json
rgarner, line 836: Oh, bugger, that's as [`rake patch:countries`](https://github.com/alphagov/dfid-transition/blob/research-outputs-import/lib/tasks/patch/countries.rake) wrote it. I guess I'll have to add it to the `Patch` classes, but make a small commit here for now.

rgarner:
Screenshots added, newline inserted, thanks @benilovj

benilovj:
Actually, I think the schema needs `"pre_production": true` added, as for [other pre-production schemas](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/lib/documents/schemas/tax_tribunal_decisions.json#L62), otherwise it's going to be published to production.

app/views/layouts/application.html.erb
```diff
@@ -19,7 +19,7 @@
         <% end %>
         <span class="caret"></span>
       </a>
-      <ul class="dropdown-menu" role="menu">
+      <ul class="dropdown-menu finders" role="menu">
```
Rosa-Fox: What was the reason for adding this?

benilovj: This class is used further down in a feature spec to find this element.

Rosa-Fox: Thanks. Grabbing it with `'.dropdown-menu li'` also works but I guess `finders` does better describe its functionality.

rgarner: We didn't want to make the bet that there wouldn't ever be another `dropdown-menu`. I'm mildly conflicted on adding stuff like this just for tests, but I think it's just about better than that bet.

Davidslv: You could always try to find the first `dropdown-menu` if someone adds another one before then your test will fail, but that might be as it will be meant to and possibly force the next person to add another test for it's own `dropdown-menu`

I can think of two possible options to do that `:first-child` or `.dropdown-menu:nth-of-type(1)`

PR #749, opened 2016-05-09
Publish button permissions for specialist documents

benilovj:
This PR hides the Publish button from specialist document writers (these users aren't allowed to publish anyway).

image


christophwong:
Looks good 👍

PR #750, opened 2016-05-09
Publish button permissions on manuals

benilovj:
Writers should not see the publish button (since they can't publish) but other roles can.

Note that publishing itself doesn't actually work yet. I'm raising this PR as a counterpart to https://github.com/alphagov/specialist-publisher-rebuild/pull/749 so that we don't forget permissions later.

There's slightly different behaviour between manuals and documents (this is carried across from Specialist Publisher v1), namely:
- writers don't see any button for manuals
- writers are told "You don't have permission to publish this document" for specialist documents

### DIFF
--- a/spec/features/publishing_a_manual_spec.rb
+++ b/spec/features/publishing_a_manual_spec.rb
@@ -1,0 +1,62 @@
+require 'spec_helper'
+
+RSpec.feature "Publishing a Manual", type: :feature do
+  let(:manual_content_item) { Payloads.manual_content_item }
+  let(:manual_links) { Payloads.manual_links }
+  let(:manual_organisation_content_id) { manual_links["links"]["organisations"].first }
+  let(:section_content_items) { Payloads.section_content_items }
+  let(:section_links) { Payloads.section_links }
+  let(:fields) { %i[content_id description title details public_updated_at publication_state base_path update_type] }
+  before do
+    publishing_api_has_content([manual_content_item], document_type: "manual", fields: fields, per_page: 10000)
+    publishing_api_has_content(section_content_items.map { |section| { content_id: section["content_id"] } }, document_type: "manual_section", fields: fields, per_page: 10000)
+
+    content_items = [manual_content_item] + section_content_items
+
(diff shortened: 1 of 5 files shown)


spec/features/publishing_a_manual_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    expect(page).to have_content("First section")
+    expect(page).to have_content("Second section")
+
+    expect(page).to have_selector(:button, 'Publish')
+
+    # the publishing itself hasn't been implemented yet
+  end
+
+  scenario "writers don't see a Publish button" do
+    log_in_as_editor(:cma_writer, organisation_content_id: manual_links["links"]["organisations"].first)
```
christophwong: Although I quite like passing in options, the other patterns used for User Factories have been creating a user type that encapsulates all the necessary attribute, is there a danger in diverging patterns?

sihugh:
Looks good to me, just the question raised by @christophwong

spec/features/publishing_a_manual_spec.rb
benilovj: I've updated the spec to be a bit clearer.

In this particular case, it's important that the editor belongs to the same organisation that the manual is associated with. In my mind, it's better for this relationship to be explicitly defined in the spec, rather implicitly matching in the factory setup, so overriding the defaults is done consciously here.

Generally speaking, I've not seen test factories scale well when they don't provide an easy way to override the defaults. What tends to happen is many very similar factory methods that are hard to maintain and reuse (we're already starting to see this with the `Payloads` approach), so I tend to favour a small number of focussed factories and setting the specifics inside the specs themselves.

benilovj:
This should be ready for another review.

spec/features/publishing_a_manual_spec.rb
christophwong: That's a very good point. I do agree with prioritising clear over condensed tests 👍

christophwong:
LGTM 👍

PR #751, opened 2016-05-09
Add max_cache_time to specialist-document content-items

mathildathompson:
- Reduce cache to 10 seconds for specialist-document content-items; 
- This will ensure website content is in sync to date with email-notification content;
- 10 seconds is also used for Travel Advice Publisher and has been confirmed by infrastructure team;
- [Trello Card](https://trello.com/c/WGlWkWT5/70-reduce-cache-on-specialist-publisher-medium)

### DIFF
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -38,7 +38,8 @@ private
     {
       body: GovspeakPresenter.present(document.body),
       metadata: metadata,
-      change_history: change_history
+      change_history: change_history,
+      max_cache_time: 10,
     }.tap do |details_hash|
       details_hash[:attachments] = attachments if document.attachments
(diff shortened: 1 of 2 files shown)


christophwong:
Would like to see what the team thinks of the duplicated attributes in payloads, if we extract the common fields out we can avoid having to add multiple repeating lines when a common attribute such as `max_cache_time` to each document payload. Or do we think it's less readable?

Otherwise, good to merge.

mathildathompson:
@christophwong @sihugh and I discussed this issue. We thought it was more readable to create another sub_method, but agree that it is less DRY. @benilovj what is your opinion on this?

benilovj:
I think that generally, the `Payloads` approach doesn't scale very well and would advocate another approach to managing the test data (maybe using FactoryGirl). However I think that's a separate piece of work.

PR #752, opened 2016-05-11
Fix email text

Rosa-Fox:
[Trello](https://trello.com/c/1QTTPUR1/106-emails-sent-for-new-content-make-clear-that-the-content-is-new-and-not-updated-small)
- Subject now only appends the text 'updated' if a document has actually been updated (previously it was also displaying when a document was 1st published)
- Name of the organisation displays within the H1 as opposed to "Documents" which was previously hard-coded into the h1 in V2
- 'Further information' paragraph dynamically informs the user if the document is published or updated aka 'For further information on this published case:' as opposed to 'For further information on this updated document' which was previously hardcoded into v2

This is an email from V1 (which this PR aims to replicate): 

v1-create-new-document

This is how a V2 email looked before this PR:

v2-create-new-document

This is what a V2 email looks (with the changes from this PR) when a document is 1st published:

v2-published-subject-line

v2-published-email

This is what a V2 email looks (with the changes from this PR) when a document is updated:

v2-updated-subject-line

v2-updated-email

Currently email information is set for all formats (except `drug-safety-updates` which are sent out manually each month). This may change to only being set for certain formats at a later date. This decision will be made after further investigation into email alert behaviour, which is detailed on this [Trello card](https://trello.com/c/S3PiWkE8/123-discovery-email-1-day).

Currently there are 2 commits on this PR, which I am planning to squash together after review. Originally I defined the text in the models, but then switched to using hashes within the `email_alert_presenter`. Please let me know which way is preferable. I prefer the hash version but can revert back if preferred.

### DIFF
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -7,7 +7,7 @@ class EmailAlertPresenter
 
   def to_json
     {
-      subject: document.title + " updated",
+      subject: subject,
       body: body,
       tags: tags,
       document_type: document.publishing_api_document_type,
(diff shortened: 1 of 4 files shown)


app/presenters/email_alert_presenter.rb
```diff
@@ -39,10 +39,25 @@ def standard_body(template_path)
...
+    publication_state ? document.title + " updated" : document.title
+  end
+
+  def updated_or_published
+    publication_state ? "updated" : "published"
+  end
+
+  def publication_state
+    document.publication_state == "redrafted"
+  end
```
tuzz: Would it be worth changing the name of this method to something like `#redrafted?` or flip the conditional and have `#published?`. I'd expected a method called `#publication_state` to return a string, whereas this is binary. What do you think?

app/views/email_alerts/medical_safety_alerts/publication.html.erb
```diff
@@ -22,7 +22,7 @@
       <% end %>
 
       <p style="color: #0B0C0C; font-family: sans-serif; font-size: 19px; line-height: 1.315789474; margin: 15px 0;">
-        For further information on this updated document:
+        For further information on this <%= updated_or_published %> <%= document_noun %>:
```
tuzz, line 25: 👍

spec/presenters/email_alert_presenter_spec.rb
```diff
@@ -3,21 +3,31 @@
...
+        email_alert_presenter_redrafted = EmailAlertPresenter.new(redrafted_cma_case)
+        presented_data_redrafted = email_alert_presenter_redrafted.to_json
+
         expect(presented_data[:subject]).to include(cma_case_payload["title"])
+        expect(presented_data[:subject]).not_to include("updated")
+        expect(presented_data_redrafted[:subject]).to include("updated")
         expect(presented_data[:body]).to include(cma_case_payload["description"])
         expect(presented_data[:body]).to include(cma_case_payload["title"])
+        expect(presented_data[:body]).to include("For further information on this published case")
+        expect(presented_data_redrafted[:body]).to include("For further information on this updated case")
```
tuzz, line 30: 👍

tuzz:
Looks good. 🎉 

One minor comment about a method name. Regarding your refactor question:

> Originally I defined the text in the models, but then switched to using hashes within the `email_alert_presenter`. Please let me know which way is preferable. I prefer the hash version but can revert back if preferred.

I'm in favour of your refactored version. 👍  I think it's cleaner to keep email-related fields out of the models and include them in the presenter.

Rosa-Fox:
Thanks @tuzz. Totally agree in regards to renaming the method to `redrafted?`. I have pushed that change and squashed the commits.

tuzz:
🚢

PR #753, opened 2016-05-11
Stop sending organisations to Rummager, add a SearchPresenter spec

rgarner:
A retrievable-by-author-only restriction in the publishing-api (thanks @tuzz):
https://github.com/alphagov/publishing-api/blob/master/app/controllers/v2/content_items_controller.rb#L64-L68

means that `get_content_items` always returned an empty organisations
list. This means that Rummager never got to relate indexable content to organisation
slugs.

@danielroseman pointed out we could use `get_linkables`
instead, so we stubbed that and added tests for everything else (there was no
 spec for `SearchPresenter`).

Subsequently, @tijmenb pointed out that this will shortly not be necessary,
since 'rummager will start looking up the organisations in the
publishing-api soon-ish'.

As a result, we removed the misleading specs and assertions and left an obvious
placeholder in the `SearchPresenter` spec from which to address this at a time 
when upstream work is complete – but the `get_linkables` commit remains here should
we subsequently need to use it.

Then we removed even that placeholder. We'll wait for Rummager.

### DIFF
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -11,25 +11,17 @@ class SearchPresenter
       description: document.summary,
       link: document.base_path,
       indexable_content: indexable_content,
-      organisations: organisation_slugs,
+      # FIXME: This line was dependent on a method called #organisation_slugs,
+      # which in turn used publishing_api.get_content to map org content_ids to
+      # organisation slugs. It was stubbed in many places in spec/, but that method
+      # call always returned a blank list in the real world.
+      organisations: [],
       public_timestamp: document.public_updated_at.to_datetime.rfc3339,
     }.merge(document.format_specific_metadata).reject { |_k, v| v.blank? }
   end
 
   def indexable_content
     document.body
-  end
-
-  def organisation_slugs
-    response = publishing_api.get_linkables(document_type: "organisation")
-
-    organisations = response.select do |organisation|
-      document.organisations.include?(organisation["content_id"])
-    end
-
-    organisations.map do |org|
-      org["base_path"].gsub("/government/organisations/", "").gsub("/courts-tribunals/", "")
-    end
   end
 
 private
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -21,10 +21,15 @@ class SearchPresenter
   end
 
   def organisation_slugs
-    response = publishing_api.get_content_items(document_type: "organisation", fields: [:content_id, :base_path])
+    response = publishing_api.get_linkables(document_type: "organisation")
 
-    orgs = response.results.select { |o| document.organisations.include?(o["content_id"]) }
-    orgs.map { |o| o["base_path"].gsub("/government/organisations/", "") }.map { |o| o.gsub("/courts-tribunals/", "") }
+    organisations = response.select do |organisation|
+      document.organisations.include?(organisation["content_id"])
+    end
+
+    organisations.map do |org|
+      org["base_path"].gsub("/government/organisations/", "").gsub("/courts-tribunals/", "")
+    end
   end
 
 private
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -11,11 +11,6 @@ class SearchPresenter
       description: document.summary,
       link: document.base_path,
       indexable_content: indexable_content,
-      # FIXME: This line was dependent on a method called #organisation_slugs,
-      # which in turn used publishing_api.get_content to map org content_ids to
-      # organisation slugs. It was stubbed in many places in spec/, but that method
-      # call always returned a blank list in the real world.
-      organisations: [],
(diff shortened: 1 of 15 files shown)


spec/features/publishing_a_cma_case_spec.rb
```diff
@@ -86,7 +85,7 @@ def withdrawn_item
     log_in_as_editor(:cma_editor)
 
     publishing_api_has_content([cma_case, minor_update_item, live_item, withdrawn_item], document_type: CmaCase.publishing_api_document_type, fields: fields, page: page_number, per_page: per_page)
-    publishing_api_has_content([cma_org_content_item], document_type: 'organisation', fields: [:base_path, :content_id])
+    publishing_api_has_linkables([cma_org_content_item], document_type: 'organisation')
```
benilovj: is this still needed?

rgarner: Eek. I missed one. Good spot.

app/presenters/search_presenter.rb
```diff
@@ -11,7 +11,11 @@ def to_json
       description: document.summary,
       link: document.base_path,
       indexable_content: indexable_content,
-      organisations: organisation_slugs,
+      # FIXME: This line was dependent on a method called #organisation_slugs,
+      # which in turn used publishing_api.get_content to map org content_ids to
+      # organisation slugs. It was stubbed in many places in spec/, but that method
+      # call always returned a blank list in the real world.
+      organisations: [],
```
benilovj: Since we're working off the assumption that [Rummager will index the organisations from Publishing API](https://trello.com/c/RrN7B6xQ/593-rework-the-way-rummager-indexes-links) very soon, do we need the FIXME and to send the organisations array at all? We can always revert this PR if the assumption turns out to be incorrect for whatever reason.

spec/presenters/search_presenter_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      it 'includes format-specific metadata' do
+        expect(json[:country]).to eql(['GB'])
+      end
+
+      it 'does not include blank values' do
+        expect { json.fetch(:blank_value) }.to raise_error(KeyError)
+      end
+
+      # This needs thought.
+      it 'prepares organisation details *somehow*'
```
benilovj: Since we're working off the assumption that [Rummager will index the organisations from Publishing API](https://trello.com/c/RrN7B6xQ/593-rework-the-way-rummager-indexes-links) very soon, I don't think we need this.

rgarner: Ah, I misunderstood. Looking at that card, "notifier rather than source of truth" seems to suggest Rummager will just go off and find stuff for itself based on, say, just `content_id` or `base_path`?

app/presenters/search_presenter.rb
rgarner: Do we know what the real user impact of not sending this now is? The card you link to seems to suggest the work is a little way off yet.

Happy to take the `organisations: []` line (and the FIXME) out, that was just trying to be explicit (it's meaningless as anything but a placeholder as it gets taken out by the `reject` below).

rgarner:
That's a better title, thanks @benilovj

spec/presenters/search_presenter_spec.rb
benilovj: > "notifier rather than source of truth" seems to suggest Rummager will just go off and find stuff for itself based on, say, just content_id or base_path?

Yes. My understanding is that:
- the publishing tool will publish links to the Publishing API
- Publishing API puts a message on the message queue
- Rummager sees the message on the fanout
- Rummager fetches the links from Publishing API and indexes them

app/presenters/search_presenter.rb
benilovj: There won't be any user impact, based on the assumption that the Finding Things work will go to production before this tool. If this assumption is no longer true, we will revert this change. I'd prefer to not leave any placeholders in, as that would necessitate another change to clean them out later.

benilovj:
👍  from me, @tijmenb have we done the right thing here?

tijmenb:
@rgarner @benilovj yes, looks good!

app/presenters/search_presenter.rb
rgarner: I guess I meant "what would the impact be if this were to go live without the Finding Things work?" In theoretical terms I'm curious about what purpose the inflamed appendix we just took out was _for_, but in practical terms I suppose I should just be happy the patient isn't dead. To use an uncomfortably _medical_ metaphor...

tijmenb: Rummager uses the organisations tag to 1) allow filtering by organisation in search and 2) display a metadata thingy in the search result:

screen shot 2016-05-12 at 12 45 48

rgarner: Thanks @tijmenb!

PR #755, opened 2016-05-12
Disable RSpec monkey-patching

tuzz:
It is usually discouraged to monkey patch RSpec methods
into Object. This change enables the RSpec option that
disables monkey patching. More information here:

http://rspec.info/blog/2013/07/the-plan-for-rspec-3/#zero-monkey-patching-mode

Happy to discuss this change if people prefer to use things
like `describe` without the prefixed `RSpec` constant.

### DIFF
--- a/spec/helpers/url_helper_spec.rb
+++ b/spec/helpers/url_helper_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe UrlHelper, type: :helper do
+RSpec.describe UrlHelper, type: :helper do
   describe "#view_on_website_link_for" do
     it "returns document's public link" do
       manual = Manual.new.tap { |doc| doc.base_path = "/guidance/abc" }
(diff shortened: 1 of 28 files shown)


benilovj:
👍  from me.

rgarner:
Personally I tend to feel that the main purpose of RSpec as a DSL is not to resemble code quite to much as to upload intent into a reader's brain as quickly as possible. I suppose it doesn't matter too much if you've only got one top-level `describe`, but I tend to use `describe '#some_method` a fair bit, and I'm a huge `context` fan. Adding an `RSpec.` wart to the beginning of everything in, say, a spec like [this](https://github.com/alphagov/dfid-transition/blob/master/spec/lib/dfid/patch/specialist_publisher/regions_spec.rb) would bother me. Not only because it's ugly, but also because that's five fewer characters out of the 80 we're supposed to stick to in the styleguide ;)

It's a :-1: from me, I'm afraid.

christophwong:
make sense to move forward with the updated syntax

PR #756, opened 2016-05-13
Refactor pre-production finder feature flag code

benilovj:
This change moves the feature flag to an explicit `SpecialistPublisher`
module, where it's more explicitly defined. The new location is shorter,
meaning it can be referenced directly, without helper methods.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -28,4 +28,10 @@ module SpecialistPublisher
 
     config.autoload_paths << Rails.root.join('lib')
   end
+
+  mattr_accessor :publish_pre_production_finders
+
+  def self.should_publish_pre_production_finders?
+    publish_pre_production_finders.present?
+  end
 end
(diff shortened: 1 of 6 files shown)


lib/specialist_publisher.rb
```diff
@@ -0,0 +1,7 @@
+module SpecialistPublisher
+  mattr_accessor :publish_pre_production_finders
+
+  def self.should_publish_pre_production_finders?
+    publish_pre_production_finders.present?
+  end
+end
```
tuzz: There's already a `SpecialistPublisher` module defined in `config/application.rb`. Perhaps this could go there? It might be easier to find than in `lib/`? It'd need to go inside the module, but outside the `class Application`.

tuzz:
Looks good 👍  one minor suggestion.

lib/specialist_publisher.rb
benilovj: Updated as per suggestion.

PR #758, opened 2016-05-13
Consistent error messages

Rosa-Fox:
[Trello](https://trello.com/c/SSLDHS0j/81-consistent-error-message-styling-and-wording-small)

Paired with @fofr on this.

Restyle and reword error messages.
- Reword error messages to keep consistent with v1
- Restyle error messages in the style of [govuk-elements](http://govuk-elements.herokuapp.com/errors/#example-ni-number)

V2 error messages before:

v2-before

V2 error messages after:

v2-after

### DIFF
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -1,3 +1,15 @@
+.elements-error-summary {
+  border: 5px solid #b10e1e;
+  padding: 15px;
+  color: #b10e1e;
+  font-weight: bold;
+
+  h4 {
+    margin-top: 0;
+    color: black;
+  }
+}
+
 form {
 
   .select2 {
@@ -9,7 +21,7 @@ form {
   }
 
   .elements-error {
-    border-left: 3px solid #b10e1e;
+    border-left: 5px solid #b10e1e;
     padding-left: 15px;
 
     label {
@@ -22,7 +34,8 @@ form {
     }
 
     input, textarea {
-      border-width: 3px;
+      border-width: 2px;
+      border-radius: 0;
       border-color: #b10e1e;
     }
 
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -8,6 +8,26 @@ form {
     height: 50px;
   }
 
+  .elements-error {
+    border-left: 3px solid #b10e1e;
+    padding-left: 15px;
+
+    label {
+      color: black;
+    }
+
+    .elements-error-message {
+      color: #b10e1e;
+      font-weight: bold;
+    }
+
+    input, textarea {
+      border-width: 3px;
+      border-color: #b10e1e;
+    }
+
+  }
+
   .error-list {
     h2 {
       font-weight: bold;
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -1,7 +1,9 @@
+$elements-red: #b10e1e;
+
 .elements-error-summary {
-  border: 5px solid #b10e1e;
+  border: 5px solid $elements-red;
   padding: 15px;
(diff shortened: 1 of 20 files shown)


app/assets/stylesheets/_forms.scss
```diff
@@ -1,3 +1,15 @@
+.elements-error-summary {
+  border: 5px solid #b10e1e;
```
Davidslv: There's a scss file for colours (which comes from using govuk_frontend_toolkit)
For example, that colour is specified here: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss#L21

app/views/shared/_form_fields.html.erb
```diff
@@ -1,12 +1,11 @@
...
+<% end %>
+
+<%= render layout: "shared/form_group", locals: { f: f, field: :summary } do %>
   <%= f.text_area :summary, class: 'form-control short-textarea' %>
-</div>
-<div class="form-group">
-  <%= f.label :body %>
+<% end %>
+
+<%= render layout: "shared/form_group", locals: { f: f, field: :body } do %>
```
Davidslv, line 9: It feels like what you looking for is `render partial: ...` instead of `render layout: ...`

app/assets/stylesheets/_forms.scss
fofr: At the moment the use of `govuk_frontend_toolkit` is limited to previewing content – I'd like to avoid mixing toolkit and admin template variables. I'd prefer `govuk_admin_template` to pull in the colour variables from toolkit, and make them available to admin apps.

For now I think it's best to define this as a colour variable within the app.

app/views/shared/_form_fields.html.erb
fofr, line 9: `render layout:` felt like the cleanest way of passing a yield-able block to a template. Unless that can be done in a cleaner way with partials?

Davidslv, line 9: Ah, I see! In that case I agree with you.

app/assets/stylesheets/_forms.scss
Davidslv: I'm not sure if I understand. I haven't found any colour variables in `govuk_admin_template` which makes me wonder that is safe to use the colours provided by `govuk_frontend_toolkit` since you are using the gem already.

I think that for `govuk_admin_template` to use the colours from `govuk_frontend_toolkit` it would require extracting that into another gem that would have common attributes between those two and then make it a requirement into both. Then yes, there's 1 source of truth for any commonalities.

I would prefer reusing something that has been already defined, decisions change all the time, maybe today red is `#b10e1e` but tomorrow is `#ff0e1e` and it's just harder to find and change many files than just one.

fofr: @Davidslv The colour variables come directly from Bootstrap.
`govuk_admin_template` defines how we should use Bootstrap and is responsible for keeping the look and feel consistent between apps.

Red in Bootstrap is one colour, red in toolkit is another. This is the first example where we are using elements over Bootstrap, hence the slight weirdness.

Davidslv: Ah! I see! Sorry for the confusion.

christophwong:
When I tested this locally, I got the old styles for error messages displayed on the `edit` form.
editing_doc
Steps to reproduce:
1. click on existing cma_case
2. click 'edit document' button
3. empty out input box for `title`
4. click 'save' and see error message

New document looks right 👍 
creating_doc

christophwong:
deleted my previous comment about a bug, I retried it and cannot reproduce it. Now that I think about it, it might have just been a cache on the style sheets. Looks good for both edit & new forms. 👍

benilovj:
👍

PR #759, opened 2016-05-16
[do not merge] Payloads refactoring spike

tuzz:
Superseded by https://github.com/alphagov/specialist-publisher-rebuild/pull/766


tuzz:
Very much in favour of providing a mechanism to aid test data setup. This looks good. 👍 

One thing I was thinking about is what these calls would do in a typical rails app:
1) `FactoryGirl.build(:aaib_report)`
2) `FactoryGirl.create(:aaib_report)`
3) `FactoryGirl.attributes_for(:aaib_report)`

This pull request seems most like 3) in that it builds a hash of attributes and leaves it up to the test to decide what to do with them.

I was wondering whether we could also have something like 1) and 2) as well. We could make `build` return an instance of the `AaibReport` model with its attributes populated and `create` return the model, but also set up the WebMock stubs against Publishing API to reflect this.

I don't know if this is going to too far, or maybe it's something we could consider later. What do you think?

PR #761, opened 2016-05-16
Send text to Rummager for Specialist Publisher bodies

benilovj:
Before this change, the app was sending Govspeak to Rummager as the document body. This isn't quite right, it's better to send text (like [Whitehall does](https://github.com/alphagov/whitehall/blob/89573063e31af11bf664c588ca560dd980e05c68/app/models/about_page.rb#L25), for instance).

(hat tip to @tijmenb)


tijmenb:
👍

PR #762, opened 2016-05-16
Change draft label colour for document index

christophwong:
worked with @Rosa-Fox and @brenetic 

The code was there to change the colour but because the index controller
passes OStruct instead of document objects to the view,
`Document#draft?` in the ApplicationHelper was returning nil and assigned the wrong style-class to the element.

[Trello](https://trello.com/c/gug9gFfl/125-draft-label-is-different-colour-in-v2-to-v1-small)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -6,7 +6,7 @@ module ApplicationHelper
   def state(document)
     state = document.publication_state == "live" ? "published" : document.publication_state
 
-    if document.draft?
+    if document.publication_state == "draft"
       classes = "label label-primary"
     else
       classes = "label label-default"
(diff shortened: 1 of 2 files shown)


benilovj:
The `ApplicationHelper` method having to have this knowledge about the document feels like a code smell – this logic wouldn't be necessary if there were a `draft?` method on the `OStruct` (or something that wraps the `OStruct`), but that may be overkill...

Anyone else have thoughts on this?

tuzz:
I personally don't think we should be using `OStruct`s. I think they're an anti-pattern. It'd have been nicer if one of our tests had raised a `NoMethodError` when this code was added to the application. I think this should be addressed separately, though.

`draft?` is probably a bit nicer than checking the `publication_state` as @benilovj suggests. The document can better encapsulate what it means for something to be in a draft state, which means that `ApplicationHelper` doesn't need to know that unnecessary detail.

We could add this to the `Document` model. What do you think?

tuzz:
Had a quick poke around and it's weird that `Document.all` returns openstructs. It doesn't instantiate instances of the class you've called it on, which is what I'd expect it to do. That means my suggestion above to add a `draft?` method to `Document` won't work. The `ActiveRecord`-esque abstraction seems a bit leaky! Because of this, I'm happy to let this slide for now and we'll address these openstruct things another day. What do you think? @benilovj

benilovj:
I'm OK to defer fixing this until later, but we should actually fix it because the structs have bitten us a couple of times already...

PR #766, opened 2016-05-19
Push unit tests down to document specs

benilovj:
Supersedes https://github.com/alphagov/specialist-publisher-rebuild/pull/759

The document model specs have a lot of duplication between them; however the models themselves (except `DrugSafetyUpdate`) don't actually contain the logic under test – the `Document` class does.

This PR moves the specs into `document_spec` and slims the specs down to a single shared spec. The shared spec checks that saving sends an appropriate payload to Publishing API, which is valid against the content schema. This step isn't possible to test in `document_spec`, because there isn't a schema to validate against for the dummy document subclass used in document specs.

Out of scope here (will look to do these on subsequent PRs):
- converting the `Payloads` module to use FactoryGirl
- tidying up `drug_safety_update_spec`

### DIFF
--- a/spec/models/countryside_stewardship_grant_spec.rb
+++ b/spec/models/countryside_stewardship_grant_spec.rb
@@ -25,18 +25,6 @@ RSpec.describe CountrysideStewardshipGrant do
     Timecop.freeze("2015-12-18 10:12:26 UTC")
   end
 
-  describe ".find" do
-    it "returns a Countryside Stewardship Grant" do
-      content_id = countryside_stewardship_grants[0]["content_id"]
-      countryside_stewardship_grant = described_class.find(content_id)
-
-      expect(countryside_stewardship_grant.base_path).to eq(countryside_stewardship_grants[0]["base_path"])
-      expect(countryside_stewardship_grant.title).to eq(countryside_stewardship_grants[0]["title"])
-      expect(countryside_stewardship_grant.summary).to eq(countryside_stewardship_grants[0]["description"])
-      expect(countryside_stewardship_grant.body).to eq(countryside_stewardship_grants[0]["details"]["body"][0]["content"])
-    end
-  end
-
   describe "#save!" do
     it "saves the Countryside Stewardship Grant" do
       stub_any_publishing_api_put_content
--- a/spec/models/countryside_stewardship_grant_spec.rb
+++ b/spec/models/countryside_stewardship_grant_spec.rb
@@ -15,16 +15,6 @@ RSpec.describe CountrysideStewardshipGrant do
     )
   end
 
-  let(:indexable_attributes) {
-    {
-      "title" => "Example Countryside Stewardship Grant 0",
-      "description" => "This is the summary of example Countryside Stewardship Grant 0",
-      "link" => "/countryside-stewardship-grants/example-countryside-stewardship-grant-0",
-      "indexable_content" => "Header " + (["This is the long body of an example Countryside Stewardship Grant"] * 10).join(" "),
-      "public_timestamp" => "2015-11-16T11:53:30+00:00",
-    }
-  }
-
   let(:countryside_stewardship_grants) { 10.times.map { |n| countryside_stewardship_grant_content_item(n) } }
 
   before do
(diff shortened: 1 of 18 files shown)


benilovj:
This is ready for review.

spec/models/a_document_subclass.rb
```diff
@@ -0,0 +1,33 @@
...
+        "details" => {
+          "change_history" => [
+            {
+              "public_timestamp" => timestamp,
+              "note" => "First published.",
+            }
+          ]
+        }
+      )
+    end
```
tuzz: Should we use this method in the `document_spec` too? There's a similar assertion about `First published`.

spec/models/document_spec.rb
```diff
@@ -39,4 +45,216 @@ def self.title
...
+        metadata: {
+          field1: "2015-12-01",
+          field2: "open",
+          field3: %w(x y z),
+          document_type: "my_document_type",
+          bulk_published: true,
+        }
+      })
+  }
+  let(:document) { MyDocumentType.from_publishing_api(payload) }
```
tuzz, line 84: 👍

spec/models/countryside_stewardship_grant_spec.rb
```diff
@@ -1,93 +1,7 @@
...
-
-      countryside_stewardship_grant = described_class.find(countryside_stewardship_grants[0]["content_id"])
-      expect(countryside_stewardship_grant.publish!).to eq(true)
-
-      assert_publishing_api_publish(countryside_stewardship_grant.content_id)
-      assert_rummager_posted_item(indexable_attributes)
-    end
-  end
+  let(:payload) { Payloads.countryside_stewardship_grant_content_item }
+  it_behaves_like "a Document subclass"
```
rgarner: How is a `Document` subclass expected to behave? I went and looked. Effectively, it looks like we mean "it saves only documents valid against the schema". Could we say that here instead?

spec/models/document_spec.rb
```diff
@@ -39,4 +45,216 @@ def self.title
       MyDocumentType.all(1, 20, q: "foo")
     end
   end
+
+  let(:finder_schema) {
+    {
+      base_path: "/my-docyment-types",
```
rgarner: typo

spec/spec_helper.rb
```diff
@@ -63,4 +63,10 @@
   # the seed, which is printed after each run.
   #     --seed 1234
   config.order = "random"
+
+  def write_payload(document)
+    document.delete("updated_at")
+    document.delete("publication_state")
+    document
+  end
```
tuzz: This is also in `document_spec`

spec/support/payloads/document.rb
```diff
@@ -303,7 +303,7 @@ def self.drug_safety_update_content_item(attrs = {})
     }.deep_merge(attrs)
   end
 
-  def self.employment_tribunal_decision_content_item(attrs)
+  def self.employment_tribunal_decision_content_item(attrs = {})
```
tuzz, line 306: What do you think about making these methods child factories of `parent`?

spec/models/document_spec.rb
```diff
@@ -39,4 +45,216 @@ def self.title
...
+      publishing_api_has_item(payload)
+      Timecop.freeze(Time.parse("2015-12-18 10:12:26 UTC"))
+    end
+
+    it "saves document" do
+      stub_any_publishing_api_put_content
+      stub_any_publishing_api_patch_links
+
+      c = MyDocumentType.find(payload["content_id"])
+      expect(c.save!).to eq(true)
```
rgarner, line 170: Is there a `context` in which it wouldn't? Both `save!` and `publish!` come with a bang, which suggests either mutable state (core Ruby style) or a variant of a method that would raise an error (ActiveRecord style, qv `find` vs `find!`, `save` vs `save!`). If it's neither of these styles, should we lose the `!`?

EDIT: I had a little look in both `Document#save!` and `Document#publish!`. It appears the first is "noisy" – will raise a `RecordNotSaved` error if anything goes wrong, and the second is wrapped in a handler block that returns `true` or `false`. So I guess we could just say `c.save!` here and drop the expectation. `publish!`, on the other hand, probably needs to be stripped of its bang or made to behave consistently :)

spec/support/payloads/document.rb
benilovj, line 306: Yes, but could we defer work on the factories to another PR? I did the minimum to keep things working here but would like to revisit the test factories wholesale on another PR.

tuzz, line 306: 👍

tuzz:
Really great changes. A few minor comments. 👍

rgarner:
This is a hugely 👍  worthwhile PR, and anything I've added above is minor. It's shown up some other stuff that doesn't need to be dealt with here.

spec/models/document_spec.rb
benilovj, line 170: I agree – `save!` doesn't behave correctly when there's a network issue. I have a fix for it locally but I'll raise a separate pull request because the fix requires a change to the production code, whereas I'd prefer to keep this as a test-only change.

spec/models/a_document_subclass.rb
benilovj: I've refactored `document_spec` to use this method.

spec/models/countryside_stewardship_grant_spec.rb
benilovj: I've renamed the shared examples to something with a bit more meaning.

spec/models/document_spec.rb
benilovj: fixed.

spec/spec_helper.rb
benilovj: I've gotten rid of the duplication.

benilovj:
I think I've addressed the feedback now – this is ready for another review.

tuzz:
Happy for this to be merged. Anything from @rgarner before this goes in?

rgarner:
All good here.

PR #768, opened 2016-05-23
"Path to live" spike

tuzz:
This is now set up at http://specialist-publisher-rebuild-standalone.integration.publishing.service.gov.uk/


tuzz:
These changes have been seen by the Mob™️ so I will merge now.

PR #769, opened 2016-05-23
Add a healthcheck specific to this application

tuzz:
When the "frankenstein" app is running in production with
both v1 and v2, it will be useful to check that the
routing to v2 is working correctly. We will add this rule
to the Nginx config, too.


tuzz:
Mobmerge.

PR #773, opened 2016-05-24
Fix root path

christophwong:
The root path redirects to manuals#index. However, this action makes a lot of api calls and currently times-out on integration, making product review quite troublesome.
Seeing that it is not in high priority to work on the manuals controller, this is a quick work around to lower the number of manuals loaded on the index page so to avoid the 504 errors from happening.

Alternative takes:
1. I've considered changing the root path so it redirects to a specialist-document endpoint. However, there isn't a document type that is open to all user (across different permissions), making it a bad choice as the root page.
1. Also, I attempted to create a stub-landing-page. This breaks some integration tests regarding what the index (root) page should display.

### DIFF
--- a/app/models/manual.rb
+++ b/app/models/manual.rb
@@ -174,7 +174,7 @@ class Manual
   end
 
   def self.max_numbers_of_manuals
-    10000
+    10
   end
 
   def update_attributes(new_attributes)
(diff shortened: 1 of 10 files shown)


christophwong:
Tested on integration:
10 manuals will not cause a time-out error.

PR #775, opened 2016-05-25
Document#save! should raise an exception on failure

benilovj:
(this is a fix for an issue raised by @rgarner on [another PR](https://github.com/alphagov/specialist-publisher-rebuild/pull/766#discussion_r64013689))

Prior to this fix, the `save!` method didn't raise an exception
if there was a failure from one of the remote calls. This is a bit misleading
because it breaks the ActiveRecord convention around bang methods ie `method!`
usually raises exceptions on failure, whereas `method` returns false
on failure.

I've not added a specific message to the exception in `save!` because this code will be revisited shortly (for the offloading of remote calls onto a queue).

### DIFF
--- a/app/models/attachment_uploader.rb
+++ b/app/models/attachment_uploader.rb
@@ -7,7 +7,8 @@ class AttachmentUploader
     attachment.url = file_url(attachment.file)
     add_attachment(document, attachment) unless attachment.changed?
     document.save!
-  rescue GdsApi::BaseError => e
+    true
+  rescue Document::RecordNotSaved => e
     Airbrake.notify(e)
     false
   end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -206,10 +206,12 @@ class Document
       presented_document = DocumentPresenter.new(self)
       presented_links = DocumentLinksPresenter.new(self)
 
-      handle_remote_error do
+      success = handle_remote_error do
         Services.publishing_api.put_content(self.content_id, presented_document.to_json)
         Services.publishing_api.patch_links(self.content_id, presented_links.to_json)
       end
+
+      raise RecordNotSaved unless success
     else
       raise RecordNotSaved
     end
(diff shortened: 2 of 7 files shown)


app/models/attachment_uploader.rb
```diff
@@ -7,7 +7,8 @@ def upload(attachment, document)
     attachment.url = file_url(attachment.file)
     add_attachment(document, attachment) unless attachment.changed?
     document.save!
-  rescue GdsApi::BaseError => e
```
tuzz, line 10: 👍  for a more specific error class rescue.

app/models/document.rb
```diff
@@ -206,10 +206,12 @@ def save!
...
-      handle_remote_error do
+      success = handle_remote_error do
         Services.publishing_api.put_content(self.content_id, presented_document.to_json)
         Services.publishing_api.patch_links(self.content_id, presented_links.to_json)
       end
+
+      raise RecordNotSaved unless success
     else
       raise RecordNotSaved
     end
```
tuzz, line 217: Could we make `#valid?` a guard clause and then flatten out the method? i.e.

``` ruby
def save!
  raise RecordNotSaved unless self.valid?

  self.public_updated_at = Time.zone.now if self.update_type == 'major'
  # ...
end
```

I think it might be a bit cleaner. What do you think?

tuzz, line 217: Should we consider adding a non-error raising version of this method, too? e.g.

``` ruby
def save
  save!
  true
rescue RecordNotSaved => e
  # do something with e.message ???
  false
end
```

It seems like we're rescuing from `RecordNotSaved` in a few places. It might be cleaner to have a regular `#save` method so that we're not raising exceptions for flow control. What do you think? Maybe a separate refactor?

There are some reasons for why it might not be such a good idea to use exceptions for flow control  here: http://programmers.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

benilovj:
Superseded by #780.

PR #776, opened 2016-05-25
Generate specialist document test data using factory girl

benilovj:
This reduces copy-and-paste duplication which comes from using raw hashes to generate test data.

I've skipped the test data generation for CMA cases and Drug Safety Updates on this PR because changing those is a bit more involved, and will come on a separate PR.


tuzz:
This is great, a huge improvement. 👍 

I was wondering whether, as a subsequent refactor, we could be fully exhaustive about the fields that are unique to each document_type. For example, the `aaib_report` document type lists these fields as `FORMAT_SPECIFIC_FIELDS`:
- date_of_occurrence
- aircraft_category
- report_type
- location
- aircraft_type
- registration

However, the factory only sets:
- date_of_occurrence

It seems quite arbitrary as to which fields are present on these factories, which might be confusing to people wondering why.

Anyway, very happy for this to be merged. 🎉  🚢

benilovj:
@tuzz agreed, I was copying the metadata across from the existing payloads. I'll extend the factories to be more complete on a subsequent PR.

spec/fixtures/factories.rb
```diff
@@ -190,4 +190,18 @@
...
+
+  factory :maib_report, parent: :document do
+    base_path "/maib-reports/example-maib-report"
+    document_type "maib_report"
+
+    transient do
+      default_metadata {
+        {
+          "date_of_occurrence" => "2015-10-10",
+          "document_type" => "maib_report",
```
Rosa-Fox, line 202: Why does `document_type` need to be sent in the metadata as well as top level?

Rosa-Fox:
👍

christophwong:
LGTM 👍

spec/fixtures/factories.rb
benilovj, line 202: I think the document type in metadata was introduced first, and at some point later, document type was added to the top level as well. I reckon that we could simplify the schema to get rid of the `document_type` in the `metadata` block /cc @tuzz

PR #777, opened 2016-05-26
Display publishing alerts

Rosa-Fox:
[Trello](https://trello.com/c/LllK2XSl/133-publishing-major-minor-warning-small)
Publishing text alerts now match those that display in V1 specialist-publisher.
- When a document is publishable, text will appear above the publish button to inform the user that email alerts will be sent out to subscribers of that specific format
- When the document `update_type` is major the appropriate message will appear above the publish button
- When the document `update_type` is minor the appropriate message will appear above the publish button

How v2 publishing text alerts looked before this PR (aka there were none):

v2-before-changes

This is what the publishing warnings look like in specialist-publisher v1.

v1 - Before publish:

v1-first-published

v1 - minor update

v1-minor-update

v1 - major update

v1-major-edit

This is what the publish warnings looks like in v2 as a result of the changes in this PR:

v2 - before publish

v2-first-published

v2 - minor update

v2-minor-update

v2 - major update

v2-major-update

### DIFF
--- a/app/views/documents/_publish_panel.html.erb
+++ b/app/views/documents/_publish_panel.html.erb
@@ -8,6 +8,7 @@
         You don’t have permission to publish this document.
       </p>
     <% elsif @document.can_be_published? %>
+      <%= @publish_warning %>
       <%= form_tag(publish_document_path(current_format.slug, @document.content_id), method: :post) do %>
         <button name="submit" class="btn btn-danger" data-module="confirm" data-message="Publishing will email subscribers to <%= current_format.title.pluralize %> Continue?">Publish</button>
       <% end %>
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -48,7 +48,9 @@ class DocumentsController < ApplicationController
     end
   end
 
-  def show; end
+  def show
+    @publish_warning = PublishWarningPresenter.new.publish_warning(@document).html_safe
+  end
 
   def edit; end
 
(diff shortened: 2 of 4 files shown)


app/views/documents/_publish_panel.html.erb
```diff
@@ -8,6 +8,14 @@
         You don’t have permission to publish this document.
       </p>
     <% elsif @document.can_be_published? %>
+      <% if @document.update_type == 'minor' %>
+        <p>You are about to publish a <strong>minor edit</strong>.</p>
+      <% elsif @document.update_type == 'major' && @document.redrafted? %>
+        <strong>You are about to publish a major edit with a public change note.</strong>
+        <p>Publishing will email subscribers to <%= @document.class.title %>s.</p>
```
benilovj: I think it'd be better if this was `<%= @document.class.title.pluralize %>`.

app/views/documents/_publish_panel.html.erb
```diff
@@ -8,6 +8,14 @@
...
+      <% else %>
+        <p>Publishing will email subscribers to <%= @document.class.title %>s.</p>
```
benilovj: ditto here

benilovj:
Some minor comments, 👍  from me otherwise.

tuzz:
It seems like this might be something that would nicely fit into a presenter:

``` ruby
def publish_info
  if ...
  elsif ...
end
```

It might help to simplify the view a bit. What do you think?

Rosa-Fox:
Thanks @benilovj and @tuzz. I have moved the logic into a presenter, let me know if it looks ok.

PR #778, opened 2016-05-26
Simplify `document_presenter_spec`

benilovj:
Introduce a FactoryGirl factory for generating CMA case payloads.
Make the test data declarations more readable by bringing scenarios
and data declaration closer to each other.

### DIFF
--- a/spec/presenters/document_presenter_spec.rb
+++ b/spec/presenters/document_presenter_spec.rb
@@ -1,119 +1,11 @@
-require 'rails_helper'
+require 'spec_helper'
 
 RSpec.describe DocumentPresenter do
-  let(:cma_case) { Payloads.cma_case_content_item }
-  let(:content_item_with_rendered_body) {
-    cma_case.deep_merge!("details" => {
-      "body" => [
-        {
-          "content_type" => "text/govspeak",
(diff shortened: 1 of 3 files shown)


spec/presenters/document_presenter_spec.rb
```diff
@@ -122,9 +14,9 @@
   end
 
   describe "#to_json without attachments" do
-    let(:specialist_document) { CmaCase.find(content_item_without_attachments["content_id"]) }
-    let(:document_presenter) { DocumentPresenter.new(specialist_document) }
-    let(:presented_data) { document_presenter.to_json }
+    let(:payload) {
+      FactoryGirl.create(:cma_case).tap { |p| p["details"].delete("attachments") }
```
christophwong, line 18: Love this explicit way of showing what test-data matters in the test. 👍

christophwong:
Looks good to merge

Side-note: There are 3 test related to using "headers" (actually headings tags); it seems that they overlap what they are testing. Maybe we only need to keep one?

benilovj:
@christophwong I agree, I think the `#to_json with headers` may be superfluous, but I left it in to keep this a refactoring, as opposed to a change at the same time...

PR #779, opened 2016-05-26
Remove first published at

christophwong:
Remove logic in DrugSafetyUpdate#publish! to save `first_published_at` on publish, because this is duplicating a behaviour in _publishing-api_.
[Trello](https://trello.com/c/NECj21MU/124-handle-first-publication-date-logic-through-the-publishing-api-medium)

This PR is only part of the story, as it involves changes in SPv1 and a migration for publishing-api.
mob: @tuzz, @issyl0, @Rosa-Fox, and myself.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -3,7 +3,7 @@ class Document
   include ActiveModel::Validations
   include PublishingHelper
 
-  attr_accessor :content_id, :base_path, :title, :summary, :body, :format_specific_fields, :public_updated_at, :state, :bulk_published, :publication_state, :change_note, :document_type, :attachments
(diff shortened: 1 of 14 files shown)


benilovj:
I think that with the way the [`Document#publish!` method](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/app/models/document.rb#L220..L236) works, the `first_published_at` won't be sent to rummager for newly created documents. We probably need to reload the document after the Publishing API publish but before it's sent to Rummager in order to get the timestamp.

christophwong:
@benilovj We have addressed your comment with the new commit.

benilovj:
be59606 looks good but I think `DrugSafetyUpdate` still suffers from the same problem. This feels like a good time to collapse the `Document#publish!` and `DrugSafetyUpdate#publish!` methods with an email "toggle" of some sort (on by default, off for DSU).

benilovj:
This looks great. Merge when ready 👍

PR #783, opened 2016-05-31
Use Capybara ‘have_no’ matchers to mitigate race conditions…

tuzz:
When asserting that content does not appear on a
page it is important to use the ‘have_no’ matchers
rather than negating the positive case. This is
because Capybara asserts against the current state
of the page. If it is has not finished following a
link, it can result in a false positive.

This has a bit more information:
http://stackoverflow.com/questions/11980109/cucumber-capybara-check-that-a-page-does-not-have-content


christophwong:
👍

PR #784, opened 2016-05-31
Change pop-up text for publish button on a minor edit document

christophwong:
Previously the pop up text incorrectly says that email will be sent out
for publishing a minor update, it is now changed to match the text used in v1.

Before - publishing a DSU with a minor edit on V2:
dsu minor v2

After - publishing a DSU with a minor edit on V2:
after dsu

These examples use DSU, but note that there isn't currently any format-specific differences in the pop up text.


app/helpers/application_helper.rb
```diff
@@ -14,4 +14,12 @@ def state(document)
...
     content_tag(:span, state, class: classes).html_safe
   end
+
+  def pop_up_warning_for_publishing(document)
+    if document.update_type == 'minor'
+      "You are about to publish a minor edit. Continue?"
+    else
+      "Publishing will email subscribers to #{current_format.title.pluralize}. Continue?"
+    end
+  end
```
tuzz: I think this might be better placed in the newly created `PublishWarningPresenter` as it seems to fit with the intent of that presenter. What do you think?

spec/features/publishing_a_cma_case_spec.rb
```diff
@@ -174,6 +180,8 @@
 
       expect(page).to have_no_content("You are about to publish a minor edit")
       expect(page).to have_no_content("You are about to publish a major edit")
+
+      expect(page.find_button('Publish')["data-message"]).to eq("Publishing will email subscribers to CMA Cases. Continue?")
```
tuzz: What do you think about DRYing up these scenarios by having something like:

``` ruby
let(:publish_alert_message) { page.find_button('Publish')["data-message"] }
```

somewhere at the top of the file?

tuzz:
Looks good. A couple of suggestions.

Is that all that was required for this ticket or is there more to come? Seems relatively simple so far!

app/helpers/application_helper.rb
christophwong: We were actually looking to pull that apart and move these frontend logic into helpers rather than presenters. Worth a chat.

christophwong:
@tuzz we added the changes per your two comments.
The tests are DRY-ed up as commented.
As for publish_warning_presenter, we pulled that code into application helpers (see the second commit)

PR #786, opened 2016-05-31
More complete format-specific test data fields

benilovj:
This PR adds values for all format-specific fields for each specialist document format. The values are taken from (or inspired by) real documents. This ought to give us greater confidence that fields will be specified correctly (in the schema, for instance).


christophwong:
Checked that all the `format specific fields` are represented 👍

PR #787, opened 2016-06-01
Refactor add attachment functionality

Rosa-Fox:
Cleaning up the behaviour of attachments for Documents and Sections. We end up removing the `AttachmentUploader` class and simplifying various methods to make them easier to understand.
- 2nd PR as part of refactoring AttachmentUploader. The other is [here](https://github.com/alphagov/specialist-publisher-rebuild/pull/785).


app/models/document.rb
```diff
@@ -251,6 +251,11 @@ def send_email_on_publish?
     update_type == "major"
   end
 
+  def add_attachment(document, attachment)
```
davidbasalla: I think the document here refers to the class instance itself? So should the method be:

```
def add_attachment(attachment):
  attachments ||= []
  attachments.push(attachment)
end
```

benilovj: fixed..

tuzz:
Good work 👍 

Attachment is a lot cleaner and I think it's easier to see what's going on now.

PR #788, opened 2016-06-01
Disable double submit on save and publish for documents with js

christophwong:
Rails provides a "disable_with" data attribute on button tags that can
prevent double submission when js is enabled.

I was not able to find a way to mimic  double clicking in feature test, so I went for checking the data-attribute is set correctly. Please let me know of any suggestions.


christophwong:
This should handle https://github.com/alphagov/specialist-publisher-rebuild/issues/621

tuzz:
Will test on integration now.

PR #789, opened 2016-06-02
Stop writing/updating public_updated_at

christophwong:
The main behaviour change here is in the first commit. We stop updating and sending `public_updated_at` to publishing-api because this is handled by publishing-api at publish time.

The second commit is cleaning up unused code. Note that there are still some consumption of `public_updated_at` for `Rummager` in the `Search Presenter`, but that is fine because the document passed to `SearchPresenter` is always a published document, which will then have `public_updated_at` set.
Also, we were using `public_updated_at` to populate the `timestamp` within `change-history`, the logic there is flagged as currently incorrect. So instead of trying to dig into the issue to decouple the two properly, I "hardcoded" the timestamp in the correct data format to pass schema validations tests but hopefully this will make it easy for future changes to rip out.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -163,14 +163,6 @@ class Document
     document
   end
 
-  def public_updated_at
-    @public_updated_at ||= Time.zone.now
-  end
-
-  def public_updated_at=(timestamp)
-    @public_updated_at = Time.parse(timestamp.to_s) unless timestamp.nil?
-  end
-
   def self.all(page, per_page, q: nil)
     params = {
       document_type: self.document_type,
@@ -203,8 +195,6 @@ class Document
 
   def save
     return false unless self.valid?
-
-    self.public_updated_at = Time.zone.now if self.update_type == 'major'
 
     presented_document = DocumentPresenter.new(self)
     presented_links = DocumentLinksPresenter.new(self)
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -75,14 +75,11 @@ private
     metadata.reject { |_k, v| v.blank? }
   end
 
-  def public_updated_at
-    document.public_updated_at.to_datetime.rfc3339
-  end
-
   def change_history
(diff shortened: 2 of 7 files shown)


app/presenters/document_presenter.rb
```diff
@@ -76,14 +75,10 @@ def metadata
...
 
-  def public_updated_at
-    document.public_updated_at.to_datetime.rfc3339
-  end
-
   def change_history
     case document.update_type
     when "major"
-      document.change_history + [{ public_timestamp: public_updated_at, note: document.change_note || "First published." }]
+      document.change_history + [{ public_timestamp: STUB_TIME_STAMP, note: document.change_note || "First published." }]
```
tuzz: Shouldn't this be `Time.zone.now` if it's in application code?

app/models/document.rb
```diff
@@ -163,14 +163,6 @@ def self.from_publishing_api(payload)
...
   end
 
-  def public_updated_at
-    @public_updated_at ||= Time.zone.now
-  end
-
-  def public_updated_at=(timestamp)
-    @public_updated_at = Time.parse(timestamp.to_s) unless timestamp.nil?
-  end
-
```
tuzz, line 173: 👍

spec/support/publishing_api_helpers.rb
```diff
@@ -3,16 +3,16 @@ def write_payload(document)
     document.delete("updated_at")
     document.delete("publication_state")
     document.delete("first_published_at")
+    document.delete("public_updated_at")
```
tuzz, line 6: 👍

app/presenters/document_presenter.rb
tuzz: Oh right, sorry. Just read your commit message. I think it'd be useful to add a big ugly comment above `STUB_TIME_STAMP` to explain something similar to your commit.

tuzz:
Looks good. One minor suggestion.

PR #792, opened 2016-06-08
Handle rummager and email-alert-api requests asynchronously in specialist-publisher-rebuild

tuzz:
https://trello.com/c/ITpVoqdD/141-offload-search-and-email-alert-requests-to-a-message-queue-when-publishing-medium

This is one of four pull requests to make the handling of rummagerand email-alert-api requests asynchronous in specialist publisher rebuild. We are making this change so that the application is more resilient to network and wider-system problems. Here are all the pull requests:
- https://github.com/alphagov/specialist-publisher-rebuild/pull/792 (this one)
- https://github.com/alphagov/govuk-puppet/pull/4478
- https://github.com/alphagov/govuk-app-deployment/pull/16
- https://github.gds/gds/development/pull/323


Gemfile
```diff
@@ -10,6 +10,8 @@ gem 'mongoid', '5.0.1'
 gem 'uglifier', '>= 1.3.0'
 gem 'jquery-rails', '~> 3.1.4'
 gem 'select2-rails', '~> 4.0.0'
+gem 'sidekiq', '~> 3.5.1'
```
elliotcm: You want to be using https://github.com/alphagov/govuk_sidekiq here, which also removes the need for the initializer.

config/application.rb
```diff
@@ -6,6 +6,7 @@
 require "action_mailer/railtie"
 require "action_view/railtie"
 require "sprockets/railtie"
+require "sidekiq/api"
```
davidbasalla, line 9: I don't know much about sidekiq. Is this line required? Do we make use of any of these query methods in the code?

tuzz, line 9: We don't, but this provides access to `Sidekiq::Queue` from a `rails console`, which is useful for debugging purposes (either in development or integration) so I thought it made sense to add this.

tuzz:
@elliotcm I have addressed comments. Please could you take another look? Thanks.

PR #793, opened 2016-06-09
Embeddable attachment snippet

davidbasalla:
This displays attachment snippets under the Attachments heading (right hand side of screen), to be used for copying and pasting into the body as part of govspeak text. These snippets get expanded into full html links by GovukPresenter as part of saving a document (thereby being sent to the publishing api).

To start, it might be best to read the test in "editing_a_cma_case_spec.rb" around attachments and the changes made in this PR for a high level overview.

As part of the PR, the GovukPresenter was refactored to accommodate more logic for expanding the attachment snippets into html links.

Writing the feature tests proved trickier than anticipated. As a consequence, there's a commit that simplifies the default document factory by removing headers from it, as headers come with their own bit of logic which we probably shouldn't bring into every test. As part of this, we added extra tests to make sure that headers are put into the payload correctly.

Paired with @Rosa-Fox 

For https://trello.com/c/DKCyHGmg/151-ability-to-embed-attachments-into-documents
# Before

screen shot 2016-06-09 at 17 28 05
# After

screen shot 2016-06-09 at 17 28 29

screen shot 2016-06-09 at 17 28 00

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -33,4 +33,8 @@ class Attachment < Document
     Airbrake.notify(e)
     false
   end
+
+  def snippet
+    "[InlineAttachment:#{url.split('/').last}]"
+  end
 end
--- a/app/presenters/document_presenter.rb
+++ b/app/presenters/document_presenter.rb
@@ -35,7 +35,7 @@ private
 
   def details
     {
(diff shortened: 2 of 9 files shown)


tuzz:
This is really excellent work. Clear and concise! 👍  🎉

PR #794, opened 2016-06-10
Create attachment collection

davidbasalla:
An attempt to encapsulate common functionality for attachments.
This follows from trying to set up a `has_many` association between
Documents and Attachments, but associations come from ActiveRecord
so not really feasible use them here. Instead we opt for a Collections
model extracts related methods from the Document and Section classes.


tuzz:
Looks good. It seems to encapsulate the collection behaviour well.

PR #797, opened 2016-06-14
Preview Govspeak button for document new and edit views

Seraphiana:
This sends an AJAX request and renders the Govspeak body text on the page.
We decided to use a Govspeak controller because serving this ajax call is something that is not specific to documents. 
We've had to update jenkins to run using Xvfb, so that Capybara can run functional tests that rely on Javascript. More details [here](https://github.com/thoughtbot/capybara-webkit#ci).
[trello card](https://trello.com/c/bcRezNAX/145-preview-button-inline-preview-is-missing-medium)

### DIFF
--- a/app/controllers/govspeak_controller.rb
+++ b/app/controllers/govspeak_controller.rb
@@ -2,7 +2,7 @@ class GovspeakController < ApplicationController
   def preview
     skip_authorization
 
-    govspeak_document = Govspeak::Document.new(params["bodyText"])
-    render json: { renderedGovspeak: govspeak_document.to_html }
+    govspeak_preview = Govspeak::Document.new(params["bodyText"]).to_html
+    render html: govspeak_preview.html_safe
   end
 end
--- a/app/controllers/govspeak_controller.rb
+++ b/app/controllers/govspeak_controller.rb
@@ -1,0 +1,8 @@
+class GovspeakController < ApplicationController
+  def preview
+    skip_authorization
+
+    govspeak_document = Govspeak::Document.new(params["bodyText"])
+    render json: { renderedGovspeak: govspeak_document.to_html }
+  end
+end
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -52,7 +52,7 @@ if [[ ${GIT_BRANCH} != "origin/master" ]]; then
     app spec lib spec config
 fi
 
-GOVUK_CONTENT_SCHEMAS_PATH=tmp/govuk-content-schemas RAILS_ENV=test bundle exec rake
+GOVUK_CONTENT_SCHEMAS_PATH=tmp/govuk-content-schemas RAILS_ENV=test xvfb-run -a bundle exec rake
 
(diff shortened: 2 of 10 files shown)


app/controllers/govspeak_controller.rb
```diff
@@ -0,0 +1,8 @@
+class GovspeakController < ApplicationController
+  def preview
+    skip_authorization
+
+    govspeak_document = Govspeak::Document.new(params["bodyText"])
+    render json: { renderedGovspeak: govspeak_document.to_html }
+  end
+end
```
tuzz, line 8: I'm not super familiar with jQuery. Could this render HTML rather than JSON containing HTML and then have the `$.post` request's callback set `$('.govspeak').html(data);` ?

jenkins.sh
```diff
@@ -52,7 +52,7 @@ if [[ ${GIT_BRANCH} != "origin/master" ]]; then
     app spec lib spec config
 fi
 
-GOVUK_CONTENT_SCHEMAS_PATH=tmp/govuk-content-schemas RAILS_ENV=test bundle exec rake
+GOVUK_CONTENT_SCHEMAS_PATH=tmp/govuk-content-schemas RAILS_ENV=test xvfb-run -a bundle exec rake
```
tuzz, line 55: Do we need to do anything locally to run these tests or will it just work?

tuzz:
Looks good. A couple of minor questions. 👍

jenkins.sh
christophwong, line 55: It will just work. This is only a special flag for an error on jenkins when starting the webkit web driver.
I had faced a local error which was to do with asset routing, a fresh clone of the repo fixed it. It probably has something to do with some state change on my local repo after we changed the config to add an asset prefix.

tuzz:
👍

jenkins.sh
h-lame, line 55: To get the specs to run Iocally (in the dev vm) I also had to use this xvfb-run command too, otherwise it would fail the 3 `js: true` features with:

```
Capybara::Webkit::ConnectionError:
       /usr/lib/rbenv/versions/2.2.4/lib/ruby/gems/2.2.0/gems/capybara-webkit-1.7.1/bin/webkit_server failed to start.
```

Is that expected?

christophwong, line 55: From what I read it seems that we would need `xvfb` on any Ubuntu box, but when running on OSX it is not needed because there is the X11 driver.
However, now I'm not sure why I am able to run it in my VM, so we might want to compare our configurations?

h-lame, line 55: Does anyone else have the same problem on the dev vm?  Maybe it's just my setup? 

FWIW: every other app that does this uses Poltergeist - perhaps because of this very issue?

christophwong, line 55: Quite likely, since Poltergeist is a phantomJS driver, it will be able to bypass any problems to do with the need of a headless web driver.

PR #798, opened 2016-06-14
Rewrite and fix change_history logic

Chrislee187:
https://trello.com/c/oTNmyzgu/146-rewrite-and-fix-change-history-logic-in-spv2-medium

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -79,6 +79,12 @@ class Document
 
   def change_history
     @change_history ||= []
+
+    if change_note && update_type == 'major'
+      @change_history + [{ 'public_timestamp' => Time.current.iso8601, 'note' => change_note }]
+    else
+      @change_history
+    end
   end
 
   def update_type
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -59,7 +59,7 @@ class Document
     "live"
   end
 
-  %w{draft live redrafted}.each do |state|
+  %w{draft live redrafted superseded unpublished}.each do |state|
     define_method("#{state}?") do
       publication_state == state
     end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -216,6 +216,12 @@ class Document
   def publish!
     handle_remote_error do
       update_type = self.update_type || 'major'
+
+      if not_published?
+        self.change_note = "First published."
+        self.save
+      end
+
       Services.publishing_api.publish(content_id, update_type)
 
       published_document = self.class.find(self.content_id)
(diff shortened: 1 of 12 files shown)


app/models/document.rb
```diff
@@ -208,6 +214,12 @@ def save
   def publish!
     handle_remote_error do
       update_type = self.update_type || 'major'
+
+      if not_published?
+        self.change_note = "First published."
+        self.save
+      end
```
tuzz: I think we need to be careful with the `#not_published?` method. It returns true if the `publication_state` returned by Publishing API is not one of `live` or `redrafted`. The Publishing API can also return states of `republish` and `unpublish` I think.

I'm not sure what the right thing to do is, though. In the case of `republish` it's ambiguous as to whether the content item was previously published or not. It might be safer to look through the `change_history` for the "First published." item or maybe look at the `first_published_at` field which should always be set on a content item that has previously been published.

app/presenters/document_presenter.rb
```diff
@@ -75,13 +75,16 @@ def metadata
...
+  #   when "major"
+  #     if document.change_note
+  #       document.change_history + [{ public_timestamp: Time.current.iso8601, note: document.change_note }]
+  #     else
+  #       document.change_history
+  #     end
+  #   when "minor"
+  #     document.change_history
+  #   end
+  # end
```
tuzz: delete

config/application.rb
```diff
@@ -24,7 +22,7 @@ class Application < Rails::Application
     # Set Time.zone default to the specified zone and make Active Record auto-convert to this zone.
     # Run "rake -D time" for a list of tasks for finding time zone names. Default is UTC.
     # config.time_zone = 'Central Time (US & Canada)'
-
+    config.time_zone = 'London'
```
tuzz, line 25: 👍

tuzz:
Looks good. One comment about using the `#not_published?` method.

app/models/document.rb
h-lame: Some document formats (dfid research output) allow editing `first_published_at` directly so checking for that doesn't seem safe.  Checking the current state for any value doesn't seem safe either because the transitions aren't linear (e.g. we can go from draft to unpublished without going through published).  The "safest" option seems to be checking for "First published" in the change note (the edgecase here would be some editor adding that as their change note before it actually was published, but that'd be pretty pathological).

Hopefully we can do https://trello.com/c/kyRoUV0h/158-add-a-state-history-field-to-presented-items-in-publishing-api-medium soon and can re-write `ever_been_published?` to interrogate the full state history.  I'll add a note to https://trello.com/c/LU6PckOO/182-fix-document-history-for-redrafts-of-unpublished-documents to make it clear that this method would need to be changed.

How does that sound?

h-lame: Oh, this mysterious `ever_been_published?` method I mention is one I propose to write in this PR to do the search through change history for "First published", instead of using `not_published?` in this conditional.

tuzz: I think I'd independently arrived at the same conclusion, this approach seems most accurate and as you say, once we have the 'state_history' information we can make this cleaner.

PR #799, opened 2016-06-14
Make ‘Document.find’ return an instance of the right subclass

tuzz:
This makes `Document.find` return a `CmaCase` if the `document_type` is `cma_case`, for example.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -184,14 +184,22 @@ class Document
   def self.find(content_id)
     response = Services.publishing_api.get_content(content_id)
 
-    if response
-      self.from_publishing_api(response.to_hash)
+    raise RecordNotFound, "Document: #{content_id}" unless response
+
+    attributes = response.to_hash
+    document_type = attributes.fetch("document_type")
+    document_class = document_type.camelize.constantize
+
+    if [document_class, Document].include?(self)
+      document_class.from_publishing_api(response.to_hash)
     else
-      raise RecordNotFound
+      message = "#{self}.find('#{content_id}') returned the wrong type: '#{document_class}'"
+      raise TypeMismatchError, message
     end
   end
 
   class RecordNotFound < StandardError; end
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -184,14 +184,22 @@ def self.all(page, per_page, q: nil)
...
+    document_type = attributes.fetch("document_type")
+    document_class = document_type.camelize.constantize
+
+    if [document_class, Document].include?(self)
+      document_class.from_publishing_api(response.to_hash)
     else
-      raise RecordNotFound
+      message = "#{self}.find('#{content_id}') returned the wrong type: '#{document_class}'"
+      raise TypeMismatchError, message
     end
```
h-lame, line 198: I find this conditional a bit hard to parse, but I also don't have any better idea of how to express it so :shrugging_emoji_on_loop: ?

h-lame:
Unless anyone has a better idea about the conditional I mentioned above, I reckon this is good to merge - makes things way less surprising.

app/models/document.rb
tuzz, line 198: Would either of these be any clearer?

1) 

``` ruby
if document_class == self || Document == self
```

2) 

``` ruby
if document_class == self
  # ...
elsif Document == self
  # ...
```

h-lame, line 198:
``` ruby
if document_class == self || self.descendants.include? document_class
  #...
else
  #...
end
```

Maybe?  Not sure how reliable `descendants` would be - but then we do constantize before we do anything so it should be ok?

h-lame, line 198: TBH I reckon I'm dangerously close to having an opinion on the colour of this bikeshed - most likely it's fine as-is.

tuzz, line 198: I think `descendants` might be incorrect if Rails autoloading hasn't loaded all of the models yet. I'm going to merge as-is and we can change it later if needed.

PR #802, opened 2016-06-16
Scripts for 2ndline and republishing tasks

tuzz:
I've finished my manual testing and this is now open for review.

I did see a few peculiarites around republishing, but I think these are out-of-scope for this ticket as they are captured in this one: https://trello.com/c/IAknDPK2/9-data-comparison-of-raib-between-v1-and-v2

I will add "republishing should work" as an A/C of the data comparison / fixes story.

There's a pull request to update the opsmanual here: https://github.gds/gds/opsmanual/pull/739


app/models/document.rb
```diff
@@ -256,6 +256,14 @@ def withdraw
     end
   end
 
+  def discard
+    handle_remote_error { discard! }
+  end
+
+  def discard!
+    Services.publishing_api.discard_draft(content_id, previous_version: previous_version)
+  end
```
h-lame: This pair of methods has different semantics to the other publishing-api write methods.  `save`, `publish!`, and `unpublish` all use `handle_remote_error` to wrap the API calls and return `true` or `false` rather than raise.  It's not clear why `publish!` has a bang when the others don't.  

I'm not sure we're following the `ActiveRecord` pattern of `!` meaning a method that might raise and the unadorned method meaning one that won't.  It's a familiar pattern though so maybe we should?  On the other hand do we even need the raise variants, we're not currently using them and it's not clear we'd want to?

Either way, this new pair of methods stands out as odd compared to what we have and we should choose how to make discarding feel more similar to the other write methods.

tuzz: I don't think we need raise variants of those methods yet, so I'll remove `discard!`. We might want them eventually, but simpler is better for now. 👍 

I think I'll rename `publish!` to `publish` too.

README.md
```diff
@@ -40,9 +40,24 @@ $ ./startup.sh
...
+this, you need to grant user permissions in this application:
+
+```bash
+bundle exec rake permissions:grant['Chris Patuzzo'] # Or whichever user you're logged in as.
+```
+
+You also need to set the `app_name` for the Dummy API User in Publishing API:
+
+```ruby
+User.find_by(email: "dummyapiuser@domain.com").update!(app_name: "specialist-publisher")
```
christophwong: Can we consolidate this blocks of text with line 65 - 76? https://github.com/alphagov/specialist-publisher-rebuild/pull/802/files#diff-04c6e90faac2675aa89e2176d2eec7d8R74

spec/workers/republish_worker_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        subject.perform(content_id)
+
+        expect(WebMock).not_to have_requested(:post, /publish/)
+      end
+
+      it "does not speak to rummager" do
+        subject.perform(content_id)
+
+        expect(WebMock).not_to have_requested(:post, /search/)
+      end
```
christophwong, line 41: would it be worthwhile to add "does not speak to email alert api" as a test as well?

tuzz, line 41: added 👍

README.md
tuzz: Good catch. I've removed that content. I think this covers all of it and it's a bit shorter. What do you think?

spec/lib/ops_tasks_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+
+      it "it uses Time.zone.now if the timestamp is set to 'now'" do
+        Timecop.freeze(2016, 7, 8) do
+          subject.set_public_updated_at(content_id, "now")
+        end
+
+        assert_publishing_api_put_content(
+          content_id,
+          request_json_includes("public_updated_at" => "2016-07-08T00:00:00.000+01:00")
```
christophwong: Is the `+01:00` the time zone for BST? Will this test be flaky when we switch to GMT?

PR #803, opened 2016-06-16
Update finder signup content item presenter

Seraphiana:
- Deprecated 'format' and replaced it with 'schema_name' and 'document_type'
- Format 'public_updated_at' to rfc3339 string, as this is the default
  requirement of content-schema validation.

[Trello Card](https://trello.com/c/HnpJ2PQf/147-create-missing-schema-for-finder-email-signup-medium),

[Depends On](https://github.com/alphagov/govuk-content-schemas/pull/327)

### DIFF
--- a/spec/presenters/finders/finder_signup_content_item_presenter_spec.rb
+++ b/spec/presenters/finders/finder_signup_content_item_presenter_spec.rb
@@ -1,0 +1,19 @@
+require 'spec_helper'
+
+RSpec.describe FinderSignupContentItemPresenter do
+  describe "#to_json" do
+    Dir["lib/documents/schemas/*.json"].each do |file|
+      it "is valid against the #{file} content schemas" do
+        read_file = File.read(file)
+        payload = JSON.parse(read_file)
+        if payload.has_key?("signup_content_id")
+          finder_signup_content_presenter = FinderSignupContentItemPresenter.new(payload, File.mtime(file))
+          presented_data = finder_signup_content_presenter.to_json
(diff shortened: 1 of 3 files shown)


spec/presenters/finders/finder_signup_content_item_presenter_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:finder_signup_content_presenter) { FinderSignupContentItemPresenter.new(payload, File.mtime(file)) }
+  let(:presented_data) { finder_signup_content_presenter.to_json }
+
+  describe "#to_json" do
+    it "is valid against the content schemas" do
+      expect(presented_data[:schema_name]).to eq("finder_email_signup")
+      expect(presented_data).to be_valid_against_schema("finder_email_signup")
+    end
+  end
+end
```
tuzz: I was thinking it might be nice to iterate through all the schemas in `lib/documents/schemas` and validate them in this test. I think you could do `Dir["lib/documents/schemas/*.json"].each` or something similar.

I think this would help to future-proof us if we decide to make changes to the content schema and want to check that all of our schemas still validate. What do you think?

Rosa-Fox:
Thanks @tuzz. Following our conversation earlier, we have only looped through the finders that contain the `signup_content_id` as they are the only schemas used for publishing using the `FinderSignupContentItemPresenter`. 

We have also improved the wording in the logger of the `publishing_api:publish_finders` rake task.

PR #804, opened 2016-06-21
Sort by last_edited_at on index page

christophwong:
This is a new field added to publishing-api, since the sorting and
writing of this field is handled by publishing-api, we just need to send
the correct params in our request for multiple content items on our
index page

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -32,7 +32,6 @@ RSpec.describe Document do
           fields: [
             :base_path,
             :content_id,
-            :updated_at,
             :last_edited_at,
             :title,
             :publication_state,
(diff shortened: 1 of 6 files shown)


spec/models/document_spec.rb
```diff
@@ -33,12 +33,13 @@ def initialize(params = {})
             :base_path,
             :content_id,
             :updated_at,
```
tuzz: Do we use `updated_at` for anything or can this be removed from the list of fields we request?

If so, we can also remove it from document's fields and make things simpler.

tuzz:
This looks fine. 👍 

To test this properly, I think we should run the V1 republishing script and check that the timestamps aren't updated on the document index page. The republish script shouldn't affect the `last_edited_at`. We can do this together if you want.

christophwong:
@tuzz let's do a test after https://github.com/alphagov/publishing-api/pull/378 is merged

PR #805, opened 2016-06-21
Guard last edited at

christophwong:
As `last_edited_at` field is rather new from publishing-api, we will
need to do some data migration to populate that field for existing
documents. In the meantime, a nil value will cause an error at
`time_ago_in_words` so I'm adding a guard clause for a smoother
transition.

Branched off of #804 as we might not need this work-around if we know that the api will never return nil for `last_edited_at`

### DIFF
--- a/app/views/documents/_document.html.erb
+++ b/app/views/documents/_document.html.erb
@@ -1,7 +1,9 @@
 <li class="document">
   <%= link_to document.title, document_path(current_format.slug, document.content_id), class: 'document-title' %>
   <ul class="metadata">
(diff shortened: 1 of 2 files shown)


tuzz:
Closing as we'd rather have it error if that field is missing. We plan on republishing this content instead.

https://trello.com/c/ZXHlaqzq/167-republish-all-v1-content-to-populate-last-edited-at

tuzz:
I think it would be better to merge this actually.

I'd rather not republish all documents in Specialist Publisher yet as this puts unnecessary load on Publishing API and clutters up the event log. We'll need to do this eventually, but we can probably do this one format at a time rather than everything now.

PR #807, opened 2016-06-23
Show rich publication state on Document#show pages

christophwong:
Dependent on: https://github.com/alphagov/publishing-api/pull/384

Introducing StateHelper to handle presenting logic for states.
I left the original ApplicationHelper#state as Manual is using it and
will have to be strangled out in the future.


app/helpers/state_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+  def compose_state(state_history)
+    if state_history.count <= 1 || latest_state(state_history) == latest_non_draft_state(state_history)
+      latest_state(state_history)
+    else
+      "#{latest_non_draft_state(state_history)} with new draft"
+    end
+  end
+
+  def latest_state(state_history)
+    state_history.flatten.last
```
tuzz: As discussed, would be better to use the highest number as hash ordering is not guaranteed.

PR #809, opened 2016-06-27
Add a dfid_authors field that accepts multiple authors

rgarner:
Multiple free-form author set in `dfid_research_outputs`.

screen shot 2016-06-27 at 12 16 15

Authors are discrete pieces of text, a bit like free-form tags.
 Whereas values like `country` have multiple label/value pairs, we
 treat authors as if the label were always equal to the value. To
 achieve this, we add a `dfid_authors` facet which is of type `text`
 but has no `allowed_values`. To avoid a situation in which these
 stored authors show up incorrectly when showing a DFID research
 output, we add to `FinderSchema#humanized_facet_value`.
## Related PRs
- https://github.com/alphagov/rummager/pull/662 (**don't merge this PR until the Rummager PR's merged**)
- https://github.com/alphagov/dfid-transition/pull/33

### DIFF
--- a/app/assets/javascript/application.js
+++ b/app/assets/javascript/application.js
@@ -7,6 +7,19 @@ jQuery(function($) {
     placeholder: $(this).data('placeholder')
   });
 
+  ////
+  // Make a select2 that will create new values on return as you type them
+  $(".select2.free-form-list").select2({
+    tags: true,
+    createTag: function (params) {
+      return {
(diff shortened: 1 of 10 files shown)


tuzz:
👍  from me

rgarner:
Thanks @tuzz. Waiting on https://github.com/alphagov/rummager/issues/662

PR #812, opened 2016-06-30
Attachments can be added on first publish

Rosa-Fox:
[Trello](https://trello.com/c/8jJTC2vu/155-adding-an-attachment-should-preserve-the-update-type-medium)
- Paired with @seraphiana
- Previously, if an attachment was added to a draft document, the document would fail to publish
- Ensures that update type is always major for draft document
- If document is not a draft then the user can select the preferred update_type
- republish_worker no longer sets a `republish` update_type for draft documents
- If the update_type for drafts was set to `republish` then emails wouldn't be sent. Content wouldn't be sent on the message queue if the document was subsequently published by an editor
- [v1 fix for similar issue](alphagov/specialist-publisher@8f5f504)

Testing - We felt this [exisiting test](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/spec/features/editing_a_cma_case_spec.rb#L228) covered as much of the journey as we could achieve without full end to end testing

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -121,7 +121,11 @@ class Document
   end
 
   def update_type
-    @update_type || "major"
+    if self.draft?
+      'major'
+    else
+      @update_type || 'major'
+    end
   end
 
   def users
--- a/app/workers/republish_worker.rb
+++ b/app/workers/republish_worker.rb
@@ -5,19 +5,22 @@ class RepublishWorker
 
   def perform(content_id, _ = nil)
     document = Document.find(content_id)
-    document.update_type = "republish"
 
     unless safe_to_republish?(document)
       print_limitations_of_republishing(document)
       return
     end
 
-    publishing_api_put_content(document)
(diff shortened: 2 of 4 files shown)


app/models/document.rb
```diff
@@ -121,7 +121,11 @@ def change_history
   end
 
   def update_type
-    @update_type || "major"
+    if self.draft? && @update_type != "republish"
```
Seraphiana: think the second half should no longer be here: even if a draft is republished it still needs to be a major update so that email alerts are sent

tuzz:
As discussed, I think we should change the tests for the `@update_type != "republish"` thing. We should never have a case where there's a draft with an update_type of "republish", so the test that was failing needs to be changed to reflect that.

Rosa-Fox:
@tuzz @Seraphiana `@update_type != "republish"` has been removed and alterations made to the spec and worker. Do they seem ok?

PR #813, opened 2016-06-30
Hide the 'unpublish' button when appropriate and fix 'Actions' messaging for all states

tuzz:
https://trello.com/c/t3uBOAdX/161-withdraw-button-only-showing-when-state-of-document-makes-it-a-valid-option-small

This turned into quite a lot of yak shaving, including fixing a bug in V1:
https://github.com/alphagov/specialist-publisher/pull/728

I decided to refactor the code to bring all of the scattered presentation logic together in the ActionsPresenter.


danielroseman:
Makes sense from a publishing workflow point of view.

(I'm not wild about building up HTML strings in Ruby code, but there's probably not an easy alternative.)

tuzz:
> I'm not wild about building up HTML strings in Ruby code

Nor am I. It's a carry-over from existing code that was consolidated here. This already turned into lots of yak shaving so I decided not to tackle that now.

PR #814, opened 2016-07-01
Rummager withdraw

Chrislee187:
Make the 'unpublish' process also issue a delete_content to Rummager to remove the document from the finder indexes.

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -399,7 +399,7 @@ RSpec.describe Document do
       assert_publishing_api_unpublish(document.content_id)
     end
 
-    it "sends a payload a delete request to Rummager" do
+    it "sends a delete request to Rummager" do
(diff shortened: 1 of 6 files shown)


spec/models/document_spec.rb
```diff
@@ -398,6 +399,12 @@ def initialize(params = {})
       assert_publishing_api_unpublish(document.content_id)
     end
 
+    it "sends a payload to Rummager" do
```
tuzz: Please we say "sends a delete request to Rummager" as I don't think it's actually sending a request body payload, just a DELETE request.

tuzz:
Looks good. One minor suggestion. 👍  🎉

spec/models/document_spec.rb
Chrislee187: Your correct it's an empty DELETE request, I've changed the statement.

PR #815, opened 2016-07-01
Check dfid_research_output against schemas

rgarner:
Due to work in govuk_content_schemas, can enable same check as for
all other models.
## Related PRs

Checks here won't pass until https://github.com/alphagov/govuk-content-schemas/pull/339 is merged

### DIFF
--- a/spec/fixtures/factories.rb
+++ b/spec/fixtures/factories.rb
@@ -183,10 +183,11 @@ FactoryGirl.define do
     transient do
       default_metadata {
         {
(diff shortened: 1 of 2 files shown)


tuzz:
👍 

We have a bit of a workflow issue in that the Jenkins builds depend on each other. If this branch passes against your govuk-content-schemas branch locally then free free to merge this and we can merge the govuk-content-schemas afterwards.

rgarner:
It does; let's 👍

PR #817, opened 2016-07-05
Update type buttons

Seraphiana:
[Trello Card](https://trello.com/c/V7VRbJSq/175-bug-cannot-choose-an-update-type-after-unpublished-and-probably-republished-too-medium)

Fixed the issue whereby users couldn't set the update type when on a "republished with new draft" or "unpublished with new draft"

Also made a minor change so that the radio buttons wouldn't be preselected when editing a published or unpublished document, but would be when viewing a redrafted document

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -121,11 +121,7 @@ class Document
   end
 
   def update_type
-    if self.draft?
-      'major'
-    else
-      @update_type || 'major'
-    end
+    self.draft? ? 'major' : @update_type
   end
 
   def users
@@ -173,6 +169,10 @@ class Document
     end
   end
 
+  def self.extract_update_type_from_payload(payload)
+    payload['update_type'] if payload['publication_state'] == 'redrafted'
+  end
+
   def self.extract_change_note_from_payload(payload)
     # If the document is redrafted remove the last/most
     # recent change note from the change_history array
@@ -193,7 +193,7 @@ class Document
       state_history: payload['state_history'],
       public_updated_at: payload['public_updated_at'],
       first_published_at: payload['first_published_at'],
-      update_type: payload['update_type'],
+      update_type: extract_update_type_from_payload(payload),
       bulk_published: payload['details']['metadata']['bulk_published'],
       change_note: extract_change_note_from_payload(payload),
       change_history: payload['details']['change_history'].map(&:to_h),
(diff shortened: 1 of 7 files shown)


app/models/document.rb
```diff
@@ -173,6 +169,10 @@ def self.extract_body_from_payload(payload)
     end
   end
 
+  def self.extract_update_type_from_payload(payload)
+    payload['update_type'] if payload['publication_state'] == 'redrafted'
+  end
```
tuzz, line 174: Would it make sense to combine this logic with the logic in the `#update_type` method? It's currently split between these two.

We could do something like:

``` ruby
def update_type
  if draft?
    'major'
  elsif redrafted?
    @update_type
  end
end
```

(untested)

spec/support/publishing_api_helpers.rb
```diff
@@ -1,11 +1,12 @@
 module PublishingApiHelpers
   def write_payload(document)
-    document.delete("last_edited_at")
-    document.delete("publication_state")
-    document.delete("first_published_at")
-    document.delete("public_updated_at")
-    document.delete("state_history")
-    document
+    copy = FactoryGirl.create(document["document_type"], document)
```
tuzz, line 3: If the test makes changes to the document's attributes before it's passed to this method, they'll be lost because it's generating a new set of attributes from the factory.

I wonder if there's a way we can still create a deep copy of the hash without calling the factory. Something like `#deep_clone` ? I think Rails has methods to help with this.

tuzz:
Looks good. 👍  A couple of minor suggestions.

spec/support/publishing_api_helpers.rb
tuzz, line 3: As discussed, ignore the comment above. You're passing `document` in so it should preserve changes to the document's attributes made in the test.

app/models/document.rb
Seraphiana, line 174: so I'm not certain on this, but it -seems- that that doesn't work, because adding an attachment needs an update type. If you add an attachment to a non-redrafted, non-draft document then the update type gets set as nil and everything gets sad. 

Not really sure how to combine them; as far as I can tell one is a change that can only happen on the way up to the view, while the other has to happen both to the view and the db.

PR #818, opened 2016-07-06
Add dfid review status facet

rgarner:
Allows filtering of research outputs based on their review status, which can be either of `unreviewed` or `peer_reviewed`. Validates based on these two values.
- [x] This build won't pass until https://github.com/alphagov/govuk-content-schemas/pull/341 is merged
- [x] This facet won't work in finder-frontend until https://github.com/alphagov/rummager/pull/673 is merged

All good to go now 👍


rgarner:
After show and tell yesterday, DFID let us know they don't want this facet – at least, not in its current form. I'll close it, rebase, then re-open when we've removed the public-facing parts.

PR #821, opened 2016-07-07
Pre production user restrictions

Rosa-Fox:
[Trello](https://trello.com/c/bG8T12G6/163-pre-production-formats-are-visible-and-editable-on-production-environments-medium)

Pre-production formats can no longer be accessed by publishers.

Previously:
- Writers and departmental editors could only interact with documents of their own organisation
- `gds_editor` could view/edit/publish documents for all formats

This PR removes the ability for publishers to access pre-production formats on all environments other than development.

### DIFF
--- a/app/policies/document_policy.rb
+++ b/app/policies/document_policy.rb
@@ -1,6 +1,6 @@
 class DocumentPolicy < ApplicationPolicy
   def index?
-    gds_editor || departmental_editor || writer
+    gds_editor? || departmental_editor? || writer?
   end
 
   alias_method :new?, :index?
@@ -10,32 +10,32 @@ class DocumentPolicy < ApplicationPolicy
   alias_method :show?, :index?
 
   def publish?
-    gds_editor || departmental_editor
+    gds_editor? || departmental_editor?
   end
 
   alias_method :unpublish?, :publish?
 
-  def pre_production_formats
+  def environment_restricted_formats
     Rails.env.development? ? [] : PRE_PRODUCTION
   end
 
-  def removed_pre_production?
-    !pre_production_formats.include?(document_class.document_type)
+  def restricted_by_environment?
+    environment_restricted_formats.include?(document_class.document_type)
   end
 
   def user_organisation_owns_document_type?
-    document_class.organisations.include?(user.organisation_content_id) && removed_pre_production?
+    document_class.organisations.include?(user.organisation_content_id) && !restricted_by_environment?
   end
 
-  def departmental_editor
+  def departmental_editor?
     user_organisation_owns_document_type? && user.permissions.include?('editor')
   end
 
-  def writer
+  def writer?
     user_organisation_owns_document_type?
   end
 
-  def gds_editor
-    removed_pre_production? && user.gds_editor?
+  def gds_editor?
+    !restricted_by_environment? && user.gds_editor?
   end
 end
--- a/app/policies/document_policy.rb
+++ b/app/policies/document_policy.rb
@@ -16,7 +16,7 @@ class DocumentPolicy < ApplicationPolicy
   alias_method :unpublish?, :publish?
 
   def pre_production_formats
-    PRE_PRODUCTION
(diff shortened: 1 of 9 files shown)


app/policies/document_policy.rb
```diff
@@ -14,4 +14,28 @@ def publish?
...
 
   alias_method :unpublish?, :publish?
+
+  def pre_production_formats
+    Rails.env.development? ? [] : PRE_PRODUCTION
+  end
+
+  def removed_pre_production?
+    !pre_production_formats.include?(document_class.document_type)
+  end
```
tuzz, line 24: I don't understand this method name.

Is it asking whether the current document type should be hidden if this isn't a production environment?

tuzz, line 24: I've thought about this a bit, perhaps something like:

``` ruby
def environment_restricted_formats
  Rails.env.development? ? [] : PRE_PRODUCTION
end

def restricted_by_environment?
  environment_restricted_formats.include?(document_class.document_type)
end

def user_organisation_owns_document_type
  ... && !restricted_by_environment?
end

def gds_editor
   ... && !restricted_by_environment?
end
```

I find those names a bit more intuitive, but keen to hear other ideas. What do you think?

config/initializers/load_pre_production_formats.rb
```diff
@@ -0,0 +1,13 @@
+def pre_production
+  pre_production_formats = []
+  Dir["lib/documents/schemas/*.json"].each do |file|
+    read_file = File.read(file)
+    parsed_file = JSON.parse(read_file)
+    if read_file.include? "\"pre_production\": true"
```
tuzz: It might be cleaner to check `parsed_file["pre_production"]` rather than looking at the (unparsed) `read_file`

spec/features/access_control_spec.rb
```diff
@@ -93,6 +93,18 @@
...
+        visit "/tax-tribunal-decisions"
+        expect(page.current_path).to eq("/manuals")
+        expect(page).to have_content("You aren't permitted to access Tax Tribunal Decisions.")
+      end
+
+      scenario "visiting a non pre_production finder" do
+        visit "/cma-cases"
+        expect(page.status_code).to eq(200)
+        expect(page).to have_content("CMA Cases")
+      end
```
tuzz, line 107: 👍

spec/features/creating_a_dfid_research_output_spec.rb
```diff
@@ -7,6 +7,7 @@
   let(:public_updated_at) { research_output['public_updated_at'] }
 
   before do
+    allow_any_instance_of(DocumentPolicy).to receive(:departmental_editor).and_return(true)
```
tuzz: Would it make sense to stub `Rails.env.development?` instead? That would mean a bit more of the code runs and it'd also add a bit of confidence that the DFID finder appears in development.

spec/features/selecting_a_finder_spec.rb
```diff
@@ -8,24 +8,35 @@
...
     end
 
-    it 'has a finder for DFID research outputs' do
+    it 'does not have a finder for pre_production finder DFID research outputs' do
+      visit '/'
+
+      click_link('Finders')
+
+      expect(page).not_to have_text('DFID Research Outputs')
+    end
```
tuzz, line 32: 👍

app/policies/document_policy.rb
Rosa-Fox, line 24: Much better, have changed.

config/initializers/load_pre_production_formats.rb
Rosa-Fox: Good idea, need to check it has the key and if so that the key contains "pre-production: true"

Seraphiana: changed, added and commited!

spec/features/creating_a_dfid_research_output_spec.rb
Seraphiana: done!

PR #822, opened 2016-07-08
Make DFID research outputs visible for testing on integration.

Seraphiana:
- DFID will be visible in production to any user that is a writer, owns DFID outputs, or is a GDS account. This shouldn't impact user experience, since no-one owns DFID.

[trello card](https://trello.com/c/Q2zdWkSi/178-grant-dfid-access-to-the-dfid-finder-on-integration-small)

### DIFF
--- a/config/initializers/load_pre_production_formats.rb
+++ b/config/initializers/load_pre_production_formats.rb
@@ -3,7 +3,7 @@ def pre_production
   Dir["lib/documents/schemas/*.json"].each do |file|
     read_file = File.read(file)
     parsed_file = JSON.parse(read_file)
-    if parsed_file["pre_production"]
+    if parsed_file["pre_production"] && parsed_file["base_path"] != "/dfid-research-outputs"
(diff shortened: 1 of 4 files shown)


config/initializers/load_pre_production_formats.rb
```diff
@@ -3,7 +3,7 @@ def pre_production
   Dir["lib/documents/schemas/*.json"].each do |file|
     read_file = File.read(file)
     parsed_file = JSON.parse(read_file)
-    if parsed_file["pre_production"]
+    if parsed_file["pre_production"] && parsed_file["name"] != "Research for Development Outputs"
```
rgarner: The name is currently volatile due to iteration / liaison with content designers / branding discussions. Can we go by the `base_path` or `content_id` instead?

Seraphiana: changed to base path. Figured that's a lot more understandable than the id

PR #824, opened 2016-07-11
Add a task for performing data comparisons

tuzz:
https://trello.com/c/IAknDPK2/85-identify-and-fix-any-data-differences-between-v1-and-v2-medium

This lets us see if there are any differences between
data the v1 application has sent to the publishing api
vs. data the v2 application sends as part of the
application's save/publish/unpublish behaviour.

Example output:

```
<<<<<
Comparing 75c9c579-956d-45bd-9357-c6483fc0ecbc
Saving document
Publishing document
Differences:
~
details.attachments[0].updated_at
2016-05-27T14:01:15Z
2016-07-11T11:51:50+00:00
-
details.metadata.bulk_published
false
~
<<<<<
```

> Chris: Data comparison is done, here are my results:
> 
> 1) It re-highlighted the bug with attachment timestamps being updated when a document is saved. This is already captured here: https://trello.com/c/llClS12G/164-sp-rebuild-changes-the-attachment-timestamps-on-document-updates-even-if-the-attachments-themselves-haven-t-been-changed
> 
> 2) Some of the documents are technically invalid because they’re missing a “summary” field. These documents don’t save in either V1 or V2 and the editor would have to set this themselves. This is an existing problem in V1, so it’s fine for V2.
> 
> 3) There’s a `bulk_published` field that V1 uses to hide some fields from the frontend. Specifically, the `updated_at` field is hidden. We could technically drop this behaviour and simplify the app because we do this better now. I might write a ticket for this soon.
> 
> ​**TL;DR**​: The data comparison was successful and there are no major issues that will block us from going live

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,6 +47,7 @@ group :development, :test do
   gem 'timecop', '0.8.0'
   gem 'govuk-content-schema-test-helpers', '1.4.0'
   gem 'govuk-lint'
+  gem 'hashdiff'
 end
 
 group :test do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -136,6 +136,7 @@ GEM
       sidekiq (~> 4.1)
       sidekiq-logging-json (~> 0.0)
       sidekiq-statsd (~> 0.1)
+    hashdiff (0.3.0)
     hashie (3.4.3)
     hike (1.2.3)
     htmlentities (4.3.4)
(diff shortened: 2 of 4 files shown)


Rosa-Fox:
Hi @tuzz,

Looks good to us. 1 question... `last_published_at` and `public_updated_at` are changed when this script is run, will this be a problem and show up on the front end if we run on production?

tuzz:
@Rosa-Fox good question. Yes, this will have side-effects and update those timestamps.

As discussed, this isn't a problem because this rake task is only intended to be run locally after a republish of those documents. So although it does affect the data, it doesn't matter because those effects won't be on any environment other than our local development environments.

I hope that makes sense!

Rosa-Fox:
👍

PR #825, opened 2016-07-12
Update finder deploy process

Rosa-Fox:
- Alterations made as a result of running through the deploy process for `raib`

### DIFF
--- a/deploy-finder-process.md
+++ b/deploy-finder-process.md
@@ -11,23 +11,28 @@ This will enqueue the documents via Sidekiq.
 
 * Once inside backend integration go to: `cd /var/apps/specialist-publisher`
 
-* Check size of the Sidekiq queue using: `Queue.all.first.size`
+* Within `govuk_app_console specialist-publisher` check size of the Sidekiq queue using: `Sidekiq::Queue.all.first.size`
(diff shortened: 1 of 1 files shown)


deploy-finder-process.md
```diff
@@ -11,31 +11,36 @@ This will enqueue the documents via Sidekiq.
 
 * Once inside backend integration go to: `cd /var/apps/specialist-publisher`
 
-* Check size of the Sidekiq queue using: `Queue.all.first.size`
+* Within `govuk_app_console specialist-publisher` check size of the Sidekiq queue using: `Sidekiq::Queue.all.first.size`
```
Davidslv, line 14: You are just selecting the size of the first queue, but if the app has more than one queue that doesn't tell you the whole story.
https://github.com/mperham/sidekiq/wiki/API#stats

tuzz, line 14: Good spot. In our case, that's ok because we're only interested in the `bulk_republishing` queue, which is the first queue:

https://github.com/alphagov/specialist-publisher/blob/master/config/sidekiq.yml#L3

Seraphiana:
👍

PR #826, opened 2016-07-13
Hotfix: Fix RAIB permissions

tuzz:
There's an issue with gds-sso that requires further investigation.
For now, this uses the organisation_slug to lookup the
organisation_content_id if it is nil.


tuzz:
Merging as we need to get this out urgently.

PR #829, opened 2016-07-13
Use `document_type` and `schema_name`

tijmenb:
`document_type` and `schema_name` have replaced `format`.

Related: https://github.com/alphagov/govuk-content-schemas/pull/348


Seraphiana:
looks like format still shows up in a few places?
publishing_api_finder_publisher_spec 14, rummager_finder_publisher_spec 19, finder_rummager_presenter 15, manual_spec 44

I'm not really sure what's going on, but it seems a bit worrying that the tests are green when they all seem to still be expecting format.

tijmenb:
Thanks @Seraphiana. 

Removed `format` from that test, as it doesn't seem to serve any purpose.

tijmenb:
`rummager_finder_publisher_spec.rb`  and `finder_rummager_presenter.rb` both deal with rummager (search). That still expects a `format` so we'll keep that.

tijmenb:
Also "fixed" [manual_spec.rb](https://github.com/alphagov/specialist-publisher-rebuild/pull/829/commits/6e8d5ddf16b8c38f40e4abbe013efc40bc6c60fd).

What's happening here is that we stub the publishing-api with a handcrafted payload (`manual_content_item`), but this payload is not being validated against a schema. This is potentially a source of bugs because if the publishing-api data output changes, our tests will still pass. Unfortunately we don't have a publishing-api content schema, so we can't test against this yet.

This PR is ready for review again.

PR #834, opened 2016-07-18
Don't preselect the 'minor' radio button after adding an attachment

tuzz:
```
When attachments are added to a document, the document's
metadata is updated and the document is saved. Previously,
it always set the update_type to 'minor' when this
happened.

This is problematic because the next time the user edits
the document, the 'minor' update radio button is
pre-selected. The user may not have chosen an update_type
yet or they may have chosen a 'major' update_type. In both
these cases, the update_type is being overriden.

This change introduces a 'temporary_update_type' field that
is set to true when adding an attachment if the user has
not previously set an update_type. The next time the
document is initialized, the update_type will be cleared if
this field is true.

This approach fixes the problem for both of the problems
described above. Note that we should not unconditionally
set temporary_update_type to true because that would cause
the user's choice to be cleared out rather than preserved
the next time the document is loaded. We should only set it
if the update_type has not yet been set by the user.
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ log
 tmp
 coverage
 /coverage
+.DS_Store
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -46,9 +46,9 @@ class AttachmentsController < ApplicationController
 private
 
   def fetch_document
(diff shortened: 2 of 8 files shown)


Seraphiana:
👍

Rosa-Fox:
@tuzz have pulled down and tested and read through all the changes, looks fab!

PR #835, opened 2016-07-18
Changing the title updates the slug

Rosa-Fox:
[Trello](https://trello.com/c/4Wrw9Ea2/186-changing-the-title-of-raib-report-not-updating-the-slug-end-part-of-the-url-medium)
- Changing the title updates the slug only for documents that have never been published
- Alters `base_paths` in test data as they differed from the document title causing test failures

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -65,7 +65,11 @@ class Document
   end
 
   def base_path
-    @base_path ||= "#{finder_schema.base_path}/#{title.parameterize}"
+    if has_ever_been_published?
+      @base_path
+    else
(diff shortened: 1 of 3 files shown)


tuzz:
Looks good. It seems to me that we have the case covered where the title changes in draft. Would it be worth adding a feature test for the case when the title does not change after it has been published? That might be one test or potentially more for each of the states.

What do you think?

Seraphiana:
Looks good to me. I like the idea of adding a check for a;ready published documents not changing

Rosa-Fox:
@tuzz @Seraphiana Thanks! Have added a test which checks that the base path is not updated for published and unpublished documents.

PR #842, opened 2016-07-27
attachment titles are automatically set if not supplied

Seraphiana:
if an attachment is added to a document and the title is not set,
then the document will default to using the name of the file.

This had the side effect of prefilling the form when going to edit
an attachment.

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -10,7 +10,7 @@ class Attachment < Document
 
   def initialize(params = {})
     params = params.symbolize_keys
-    @title = params[:title]
+    @title = extract_title(params)
     @file = params[:file]
     @content_type = params[:content_type]
     @url = params[:url]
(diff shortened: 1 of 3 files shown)


app/models/attachment.rb
```diff
@@ -25,6 +25,17 @@ def update_attributes(new_params)
     end
   end
 
+  def extract_title params
+    if params[:title].blank?
+      if params[:url]
+        separated_url = params[:url].split('/')
+        separated_url[separated_url.length - 1].split('.')[0]
```
tuzz: Could this be `seperated_url.last.split('.').first` ?

tuzz:
👍  one minor suggestion

app/models/attachment.rb
Seraphiana: fixed

PR #844, opened 2016-07-28
Fixed flash message issue for new files

Seraphiana:
Fixed issue whereby flash message when a new attachment was added
would not show attachment name if it had been automatically generated

Also added some more tests to cover this

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -25,11 +25,14 @@ class Attachment < Document
     end
   end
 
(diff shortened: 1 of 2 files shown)


Seraphiana:
I'm not entirely sure the conditional needs to check the URL: it seems more reilient, since we should always be getting a URL from the asset manager, but I don't know if its possible to ever receive a file that doesn't have a title

PR #845, opened 2016-07-28
Untangle change history from redrafted

tuzz:
This change makes ChangeHistory a first-class
object that exposes some methods for adding and
updating items. This means the app doesn’t need
to know the details of how the change history is
represented.

Other than refactoring, the driver for this change
is to remove all reference of ‘redrafted’ from the
code base. Previously, change history was so
tangled up with ‘redrafted’ that it makes sense to
fix this first before continuing with that work.

### DIFF
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -192,10 +192,6 @@ RSpec.describe Document do
         expect(document.field2).to eq("open")
         expect(document.field3).to eq(%w(x y z))
       end
-
-      it "sets the change note to nil" do
-        expect(document.change_note).to be_nil
-      end
     end
 
     context "when the document is redrafted" do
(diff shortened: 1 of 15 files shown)


spec/presenters/actions_presenter_spec.rb
```diff
@@ -65,19 +65,13 @@
...
 
-    context "when the update_type is minor" do
-      let(:payload) { FactoryGirl.create(:cma_case, publication_state: "redrafted", update_type: "minor") }
-      specify { expect(subject.publish_text).to include("minor edit") }
-      specify { expect(subject.publish_text).not_to include("will email subscribers") }
-    end
-
     context "when the document is redrafted and the update_type is minor" do
-      let(:payload) { FactoryGirl.create(:cma_case, publication_state: "redrafted", update_type: "minor") }
+      let(:payload) { FactoryGirl.create(:cma_case, first_published_at: "2016-01-01", publication_state: "redrafted", update_type: "minor") }
```
Seraphiana: should this not be publication_state draft? I guess it doesn't need to change just yet, its just a little confusing in this weird still-sort-of-supporting-redrafted state

spec/models/document_spec.rb
```diff
@@ -689,30 +673,27 @@ def initialize(params = {})
...
+      subject.update_type = "minor"
+      subject.change_note = "some change note"
 
-      expect(document.change_history).to be_empty
+      expect(subject.change_note).to be_nil
     end
 
-    it 'should not add note when no update type' do
-      document.update_type = ''
+    it "does not the change note when the update_type is not set" do
```
Seraphiana, line 694: minor typo

PR #847, opened 2016-08-02
Add a subscription_list_title_prefix

rgarner:
All other finders have this, and we have a problem with our topic
only displaying as '#' on email signup (they still work).

Also, remove the `topics` key containing some/dummy-topic. It isn't
required.


Seraphiana:
👍

PR #862, opened 2016-08-11
Migrated Asylum Support Decisions to be supported by v2

Seraphiana:
Rebuild should now support Asylum Support Decisions documents


Seraphiana:
I'm not really happy with all the front end bits of this, but can't see a way to tidy them up without deviating a lot from V1

PR #868, opened 2016-08-17
Grant writers/editors access to pre-prod finders

tuzz:
Previously we only allowed gds_editors to access
pre-prod finders in the rebuild. This isn't right
as V1 has no restrictions on this. The only real
difference between prod/pre-prod is that the finder
hasn't been published and no user has yet published
a document for that finder.

### DIFF
--- a/app/policies/document_policy.rb
+++ b/app/policies/document_policy.rb
@@ -15,16 +15,8 @@ class DocumentPolicy < ApplicationPolicy
 
   alias_method :unpublish?, :publish?
 
-  def environment_restricted_formats
-    gds_editor? ? [] : PRE_PRODUCTION
-  end
-
-  def restricted_by_environment?
-    environment_restricted_formats.include?(document_class.document_type)
-  end
-
   def user_organisation_owns_document_type?
-    document_class.organisations.include?(user.organisation_content_id) && !restricted_by_environment?
+    document_class.organisations.include?(user.organisation_content_id)
   end
 
   def departmental_editor?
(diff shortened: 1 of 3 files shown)


spec/features/creating_an_employment_appeal_tribunal_decision_spec.rb
```diff
@@ -88,24 +88,27 @@
...
     before do
       allow(Rails.env).to receive(:development?).and_return(false)
-      log_in_as_editor(:gds_editor)
+      stub_any_publishing_api_call
+      publishing_api_has_item(research_output)
     end
 
     context "when logged in as an editor" do
-      before { log_in_as_editor(:editor) }
+      before { log_in_as_editor(:cma_editor) }
```
Seraphiana: seems like possibly this should be moved into a different test/renamed test, since its in a namespace for EAT stuff

app/policies/document_policy.rb
```diff
@@ -15,16 +15,8 @@ def publish?
...
-    gds_editor? ? [] : PRE_PRODUCTION
-  end
-
-  def restricted_by_environment?
-    environment_restricted_formats.include?(document_class.document_type)
-  end
-
   def user_organisation_owns_document_type?
-    document_class.organisations.include?(user.organisation_content_id) && !restricted_by_environment?
+    document_class.organisations.include?(user.organisation_content_id)
```
Seraphiana, line 19: 👍

PR #874, opened 2016-08-19
Add our learnings from migrating to phase 2

tuzz:
Quite a lot of changes. I consolidated some bits from the README into this directory.

cc @boffbowsh, @steventux 

I think a good next step would be for me to run through this with interested people.

**Update:** Rosa and I will be presenting this at the GOV.UK Show & Tell on 15th September!


docs/phase-2-migration/deployment-guide.md
```diff
@@ -1,4 +1,4 @@
-# Process for Deploying a Finder to V2
+## Deployment guide
 
 After following this process, a specified finder would be served from v2 `specialist-publisher` as opposed to v1.
```
boffbowsh, line 3: (Can't comment on stuff not changed in this PR)

Rather than running `Sidekiq::Queue.all.first.size` a lot, link to the documentation for sidekiq monitoring in the ops manual

docs/phase-2-migration/workflow-of-content.md
```diff
@@ -0,0 +1,65 @@
...
+
+## API requests
+
+This section lists the API requests made at each transition. In total,
+Specialist Publisher speaks to four external services.
+
+**new / edit**
+
+- Publishing API: PUT /v2/content
+- Publishing API: PATCH /v2/links
```
MatMoore, line 39: Was there a deliberate decision to PATCH links when creating/editing a draft?

I would have expected this happen on publish, although this is not really possible at the moment without storing the data somewhere.

Based on this, I'd expect that if a user creates a new draft of a previously published document, part of their changes (the tags) take effect before they click publish.

tuzz: What do you mean by "take effect" ?

Which tags are you referring to? I think I'm lacking context / understanding to answer your question – sorry.

tuzz:
Updated /cc @boffbowsh

docs/phase-2-migration/workflow-of-content.md
boffbowsh: This is probably relevant https://gov-uk.atlassian.net/wiki/display/GOVUK/RFC+31%3A+Requirements+for+tagging+architecture

MatMoore: @tuzz I just meant that `patch_links` is something that I'd expect to affect the site straight away (for example, by changing what shows up on a browse page).

It might be better to discuss this offline though - the documentation is great, I was just curious about why it works that way, in case we need to improve something.

tuzz: Ahh, I understand your point. I agree that this should probably happen at publish time. That being said, we always send the same links in Specialist Publisher Rebuild, so it doesn't actually matter in this case. I'll add a note to that documentation to that effect.

tuzz:
Updated with a note about the `PATCH /v2/links` request as per @MatMoore's comment.

Going to merge now.

PR #875, opened 2016-08-22
Truncate long urls

Rosa-Fox:
[Trello](https://trello.com/c/PVT5Fnpd/255-truncate-over-length-dfid-urls-medium)
- Shorten urls to 250 characters
- Every night a crawler backs up govuk. Before the change made in this PR, when the crawler tried to back up pages with long urls in specialist-publisher-rebuild it failed. This problems was fixed temporarily by [this change to Puppet](https://github.com/alphagov/govuk-puppet/pull/4791).

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -1,6 +1,7 @@
 class Document
   include ActiveModel::Model
   include ActiveModel::Validations
+  include ActionView::Helpers::TextHelper
   include PublishingHelper
 
   attr_accessor(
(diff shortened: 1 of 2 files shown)


tuzz:
👍  looks good. We also need to write a script to actually update those documents. We might be able to use the `RepublishWorker` for this, although I'm not 100% confident in its operation. If you'd like, we can break that into a separate ticket? What do you want to do?

Rosa-Fox:
@tuzz Separate PR for truncating the long urls that are already live sounds good. Will use the same Trello card for both tasks, but have updated it to explain.

PR #877, opened 2016-08-24
Validate attachment snippets

tuzz:
screen shot 2016-08-25 at 11 18 46
This pull request makes attachment snippet matching far more robust. It also fixes a problem where some attachments don't have a url, which was resulting in a 500 server error for a few cma case pages.

Eventually, I'd like to see if we can surface some of the information in `AttachmentReporter` to the user-interface so that users can see errors if the snippets are invalid. It also reports on attachments that aren't used in the body of the document.


Seraphiana:
👍

PR #878, opened 2016-08-25
Remove external link-icon from attachments

tuzz:
(reopened after tidying up the git history)


tuzz:
https://github.com/alphagov/specialist-publisher-rebuild/pull/881

PR #881, opened 2016-08-25
Remove external link-icon from attachments

tuzz:
Added www.gov.uk and assets.publication.service.gov.uk as "whitelist" entries in the Govspeak generation to avoid the 'external-link' icon appearing for attachments and internal references.


Seraphiana:
looks good to me 👍

PR #883, opened 2016-08-26
Fix organisation permissions so that they match V1

tuzz:
I also annotated the files so it's easy to see who
can access what. In total, there were two permission
problems:

Asylum Support Decisions:
- Missing ministry-of-justice, hm-courts-and-tribunals-service

UTAAC:
- Missing ministry-of-justice, hm-courts-and-tribunals-service


brenetic:
👍  Annotations very helpful!

PR #886, opened 2016-09-01
Medical Safety Alerts flagged as Urgent

Seraphiana:
Updated app to send an urgent flag with email requests for medical
safety alerts. Flag defaulted to nil in other formats.

[trello card](https://trello.com/c/QbYLGjL2/263-send-medical-device-alerts-urgently)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -347,6 +347,18 @@ class Document
     update_type == "major"
   end
 
+  # This is set to nil for all non-urgent emails.
+  # Override to true for urgent email handling for a specific format.
+  # Urgent emails are sent immediately to all users,
+  # regardless of how frequently the users are set to get email updates
+  #
+  # Sending false will force overriding of topic defaults, and should
+  # only be done where we explicitly want an email to be non urgent and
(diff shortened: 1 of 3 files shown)


Seraphiana:
@tijmenb @elliotcm

app/models/document.rb
```diff
@@ -347,6 +347,10 @@ def send_email_on_publish?
     update_type == "major"
   end
 
+  def urgent
```
tijmenb: Might be worth to put something like https://github.com/alphagov/specialist-publisher/blob/6646ebe34f03d8ea9537c622edf83c32060ba27d/app/exporters/formatters/abstract_document_publication_alert_formatter.rb#L99-L102 here too, to carry over the knowledge of this behaviour.

tijmenb:
👍 looks good

app/models/document.rb
Seraphiana: good point. Is that clearer?

tijmenb: 👍 excellent!

PR #888, opened 2016-09-02
Validate open close dates

Rosa-Fox:
[Trello](https://trello.com/c/qFKUSFH9/76-sp-will-let-you-set-a-closed-date-before-the-opened-date-small)
- Add validation to ensure `opened_date` is before `closed_date` to reduce risk of human error when creating CMA documents.
- Rake task to get `content_ids` of documents which have an `opened_date` later than their `closed_date`.
- Prints out `opened_date`, `closed_date` and `content_id` for the identified documents. The next step would be to find the problem documents and manually fix these date fields using the printed `content_ids` outputted by this rake task.

Run rake task:

```
bundle exec rake opened_before_closed_dates
```

Produces output like:

```
Opened date is before closed date. Opened: 2015-02-14, Closed: 2012-04-11, content_id: bd71f5c1-b319-4151-9c79-7186135c6bd2
Opened date is before closed date. Opened: 2002-08-27, Closed: 2001-04-18, content_id: d7ee9561-832f-4c6f-ba4c-3c2053520a4b
Opened date is before closed date. Opened: 2015-10-07, Closed: 2015-01-06, content_id: a22e15b2-51be-4d19-a8e2-801ebfc5722b
```

### DIFF
--- a/app/models/cma_case.rb
+++ b/app/models/cma_case.rb
@@ -1,5 +1,5 @@
 class CmaCase < Document
-  validates :opened_date, allow_blank: true, date: true
+  validates :opened_date, allow_blank: true, date: true, open_before_closed: true
   validates :market_sector, presence: true
   validates :case_type, presence: true
   validates :case_state, presence: true
--- a/app/validators/open_before_closed_validator.rb
+++ b/app/validators/open_before_closed_validator.rb
@@ -1,0 +1,15 @@
+class OpenBeforeClosedValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, _value)
+    record.errors.add(attribute, error_message) unless before_closed?(record)
+  end
+
+private
+
+  def before_closed?(record)
+    record.opened_date <= record.closed_date || record.closed_date.blank?
+  end
+
+  def error_message
+    options[:message] || "must be before closed date"
+  end
+end
--- a/lib/tasks/opened_closed_dates.rake
+++ b/lib/tasks/opened_closed_dates.rake
@@ -1,0 +1,18 @@
+desc "Find CMA Cases with opened dates that are before closed dates"
+task opened_before_closed_dates: :environment do
+  cma_cases = Services.publishing_api.get_content_items(
+    document_type: "cma_case",
+    per_page: 999999,
+  ).results
+
+  non_blank_opened_dates = cma_cases.select { |cma| !cma.details.metadata.opened_date.blank? }
+  non_blank_closed_dates = cma_cases.select { |cma| !cma.details.metadata.closed_date.blank? }
+
+  non_blank_opened_dates.each do |cma|
+    if non_blank_closed_dates.include? cma
+      unless cma.details.metadata.opened_date <= cma.details.metadata.closed_date
+        puts "Opened date is before closed date. Opened: #{cma.details.metadata.opened_date}, Closed: #{cma.details.metadata.closed_date}, content_id: #{cma.content_id}"
+      end
+    end
+  end
+end
--- a/spec/features/creating_a_cma_case_spec.rb
+++ b/spec/features/creating_a_cma_case_spec.rb
@@ -137,6 +137,78 @@ RSpec.feature "Creating a CMA case", type: :feature do
     expect(page).to have_content("Body cannot include invalid Govspeak")
   end
 
+  scenario "with closed date before opened date" do
+    visit "/cma-cases/new"
+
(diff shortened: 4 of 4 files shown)


app/validators/open_before_closed_validator.rb
```diff
@@ -0,0 +1,15 @@
+class OpenBeforeClosedValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, _value)
+    record.errors.add(attribute, error_message) unless before_closed?(record)
+  end
+
+private
+
+  def before_closed?(record)
+    record.opened_date <= record.closed_date || record.closed_date.blank?
+  end
```
tuzz, line 10: This method might raise an error if `record.opened_date` or `record.closed_date` are `nil` because the `<=` method isn't defined on `NilClass`. Would it be worth testing this case explicitly? We could potentially do those as unit tests for this validator if that's easier. What do you think?

lib/tasks/opened_closed_dates.rake
```diff
@@ -0,0 +1,18 @@
...
+  non_blank_closed_dates = cma_cases.select { |cma| !cma.details.metadata.closed_date.blank? }
+
+  non_blank_opened_dates.each do |cma|
+    if non_blank_closed_dates.include? cma
+      unless cma.details.metadata.opened_date <= cma.details.metadata.closed_date
+        puts "Opened date is before closed date. Opened: #{cma.details.metadata.opened_date}, Closed: #{cma.details.metadata.closed_date}, content_id: #{cma.content_id}"
+      end
+    end
+  end
+end
```
tuzz, line 18: 👍  is the plan to tell CMA which documents are invalid?

Are there any other formats that have an open/close date we should check?

spec/features/creating_a_cma_case_spec.rb
```diff
@@ -137,6 +137,27 @@ def cma_case_content_item_links
...
+
+    expect(page.status_code).to eq(422)
+
+    expect(page).to have_css('.elements-error-summary')
+    expect(page).to have_css('.elements-error-message')
+
+    expect(page).to have_content("Please fix the following errors")
+    expect(page).to have_content("Opened date must be before closed date")
+  end
+
```
tuzz, line 160: 👍

app/models/cma_case.rb
```diff
@@ -1,5 +1,5 @@
 class CmaCase < Document
-  validates :opened_date, allow_blank: true, date: true
+  validates :opened_date, allow_blank: true, date: true, open_before_closed: true
```
tuzz, line 2: 👍  good use of a custom validator ⭐

app/validators/open_before_closed_validator.rb
Rosa-Fox: Hi,

If `opened_date` or `closed_date` are empty then it still works, I guess maybe it sends as an empty string in the form and then persists as nil? So it compares against `""` which works ok. I added some more feature tests... let me know if I am missing something!

lib/tasks/opened_closed_dates.rake
Rosa-Fox: CMA is the only format that has `opened_date` and `closed_date`. As far as I can see the only other format with something similar is `Vehicle Recalls and Faults Alerts` which has `Build Start Date` and `Build End Date`. I would need to add a new validation method though due to the difference in field names... should I add it in?

tuzz: Nah, I think that's fine for the sake of scope creep. 👍

app/validators/open_before_closed_validator.rb
tuzz: Ahh, that makes sense. Now that we have tests, this is fine!

Seraphiana: that seems a bit dangerous on documents that have already been published/created with invalid dates. Do any documents with invalid data currently exist? I think having the check for nil to avoid an unnecessary 500 is probably worthwhile.

lib/tasks/opened_closed_dates.rake
Seraphiana: May as well, while we're here ::::)

tuzz: Could do if it's quick.

app/validators/open_before_closed_validator.rb
tuzz: Hmm, I hadn't thought about that. What do you think @Rosa-Fox ?

PR #892, opened 2016-09-06
Validate inline attachments

tuzz:
We now check that all inline attachments used in
the body of the document match with an attachment
in the list on the right-hand side.

I've made it so we don't validate the document when
saving/updating attachments. This was already a
problem with fields like 'summary' which was missing
on some documents.
## Screenshot

screen shot 2016-09-06 at 15 38 02


Seraphiana:
👍

PR #894, opened 2016-09-07
Fix CI: Clean up users before each test

tuzz:
On CI, there were 2300 users in the database.
There's no uniqueness constraint on the 'uid'
field which means the wrong user is retrieved
in feature tests if more than one request is
made because the user is cached in the
authenticated users header.

I'm not sure why this started failing now. Some
external factor must have changed


tuzz:
Self-merging because master is broken.

PR #895, opened 2016-09-07
Make OpenBeforeClosedValidator robust to nils

tuzz:
We received a Zendesk ticket. Upon investigation,
this validator was raising an error. Normally these
dates would be empty strings which would work with
the previous implementation but because this data
was imported from the old specialist publisher it
is sometimes nil.

https://govuk.zendesk.com/agent/tickets/1413926


tuzz:
Self-merging because this is an urgent production issue.

PR #897, opened 2016-09-07
Set attachment `updated_at` timestamp correctly

Rosa-Fox:
[Trello](https://trello.com/c/llClS12G/252-sp-rebuild-changes-the-attachment-timestamps-on-document-updates-even-if-the-attachments-themselves-haven-t-been-changed-medium)

Previously all attachment `updated_at` timestamps on the edited document were updated in these cases:
- Every time anything in the document was updated (for example the document title or body)
- Every time any attachment was updated or uploaded

This PR fixes this by:
- Setting a temporary `being_updated` flag when an attachment's title or file is being updated
- Only set or update the `updated_at` timestamp for attachments that are new or being updated

attachments_updated_at

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -43,7 +43,13 @@ class AttachmentsController < ApplicationController
 
 private
 
+  def flag_updated(document, attachment)
+    updated_attachment = document.attachments.find(attachment.content_id)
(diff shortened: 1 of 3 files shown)


Seraphiana:
had a look at some functional tests for this journey; sadly they don't really work without a proper asset manager rather than our stubs. 

so I think this is 👍

PR #898, opened 2016-09-08
Pass request tracing headers into Sidekiq jobs

tuzz:
Previously, the govuk_sidekiq gem failed tests
with an argument error which led us to make the
final argument to these methods an underscore to
make the implementation work against the tests
but also against the real sidekiq.

This issue has now been fixed in the govuk_sidekiq
gem and so I have swapped out the line in
spec_helper.rb as suggested. That means we can now
correctly pass request tracing headers through.

https://trello.com/c/LtwMm3Pn/205-publish-request-tracing-medium


Seraphiana:
looks good to me. I assumed it was more complicated than this, but looking through the background info I'm satisfied. 👍

PR #899, opened 2016-09-09
Reflect changes in PublishingAPI

kevindew:
This addresses the changes introduced in the Publishing API in https://github.com/alphagov/publishing-api/pull/503

The Publishing API change doesn't cause any immediate problems this just will add an extra layer of isolation between this and any apps that may use the same `document_type` later.


carlosmartinez:
approve12

PR #900, opened 2016-09-12
Fix schema validations

tuzz:
A mix of fixes for schema validations. I've published something on integration using this branch and no errors are raised. I also ran the finder/email_signups publishing task and no errors were raised.

I've added a schema validation test for the finders so that we'll know if there's a regression.


Seraphiana:
👍

PR #901, opened 2016-09-13
Stop sending "signup_link" to publishing-api

tijmenb:
This attribute used to be used to link a finder to its email signup page. This is no longer used because finders are linked to their signup page in the links hash.

Related: https://github.com/alphagov/govuk-content-schemas/pull/386


Seraphiana:
👍

PR #902, opened 2016-09-13
Separate finders into their own app

Rosa-Fox:
[Do not merge] because I am not 100% sure if we want to deploy this at the same time as we reconfigure Signon to allow users to log into Manuals and Finders separately?

**Context**

Specialist Publisher will be split into two apps. 'Specialist Publisher' and 'Manuals Publisher'.

At the time of this PR, the plan is to rename this application (Specialist Publisher Rebuild) to `Specialist Publisher`. This will only serve the Finder functionality.

V1 Specialist Publisher will serve the manuals functionality. V1 Specialist Publisher will be named 'Manuals Publisher'. In due course, Manuals Publisher will also be rebuilt.

**This PR**
- Removes manuals code 
- Redirects the `/` root url as it can no longer be `/manuals`

Screen grabs of the mock ups of how this will appear to a user can be found [here](https://github.com/Rosa-Fox/mockups).

[Trello: Create separate landing pages for manuals and finders](https://trello.com/c/Yj5tf91W/273-create-separate-landing-pages-for-manuals-and-finder-small)
[Trello: Delete manuals code from the rebuild](https://trello.com/c/nnCA1kR0/242-delete-manuals-code-from-rebuild-small)


app/controllers/passthrough_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+    if current_user.gds_editor?
+      redirect_to "/aaib-reports"
+    elsif current_organisation
+      redirect_to("/" + current_organisation.slug) && return
+    else
+      redirect_to error_path
+    end
+  end
+
+  def error
```
Seraphiana, line 16: not sure we want an empty error method here

app/controllers/passthrough_controller.rb
```diff
@@ -0,0 +1,18 @@
+class PassthroughController < ApplicationController
+  after_action :skip_authorization
+
+  def index
+    current_organisation = document_models.find { |model| model.organisations.include? current_user.organisation_content_id }
+
+    if current_user.gds_editor?
+      redirect_to "/aaib-reports"
+    elsif current_organisation
+      redirect_to("/" + current_organisation.slug) && return
```
Seraphiana: what changed to make && fine rather than and?

Seraphiana:
left a couple of comments, otherwise looks good 👍

app/controllers/passthrough_controller.rb
Rosa-Fox: The empty `error` action is needed for the railsy naming convention based routing to work. Let me know if another name for the route would be better than `/error`.

app/controllers/passthrough_controller.rb
Rosa-Fox: I had to use brackets after `redirect_to` so that using `&&` worked. Even though `and` would be correct, it seems like it is a [known problem](https://github.com/bbatsov/rubocop/issues/1487) with Rubocop (which govuk-lint is made from) so I don't think I could have changed anything in govuk-lint to make it work. Is a bit annoying but I think it is small enough to not be too much of an issue.

app/controllers/passthrough_controller.rb
Seraphiana: hmm. Maybe something a bit more specific. But if its rails magic then sticking with error might be best

app/controllers/passthrough_controller.rb
Seraphiana: ok, fair enough

boffbowsh:
It's the ghost of @tommyp past

tommyp:
People keep sending me links to PRs here and I have to feign interest.

Rosa-Fox:
It's good to know Specialist Publisher still holds a place in your heart @tommyp :D

tommyp:
More like a place in my inbox.

The junk folder.

app/views/passthrough/error.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>Permission Denied</h1>
+<p>Please check that your Signon organisation ID is correct</p>
```
tuzz: Perhaps we could change this so that it matches the other message, something like:

> You aren't permitted to access this application. If you feel you've reached this in error, contact your SPOC.

app/controllers/passthrough_controller.rb
tuzz: I reckon you can get rid of this `return` all together!

tuzz:
Looks fine to me! A couple of minor suggestions. 👍

Rosa-Fox:
Thanks @tuzz. I have made those changes.

tuzz:
Technically, I think this is good to merge. We just need to wait for the green light from product.

app/controllers/passthrough_controller.rb
```diff
@@ -0,0 +1,18 @@
+class PassthroughController < ApplicationController
```
steventux: I'm not convinced of the name `Passthrough` or the need for a separate controller.
I think we could just as easily add an action to documents_controller to achieve this.

app/controllers/passthrough_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  def index
+    current_organisation = document_models.find { |model| model.organisations.include? current_user.organisation_content_id }
+
+    if current_user.gds_editor?
+      redirect_to "/aaib-reports"
+    elsif current_organisation
+      redirect_to "/" + current_organisation.slug
+    else
+      redirect_to error_path
```
steventux: Rather than redirecting to an empty error action why not give an appropriate status (possibly a 403) and message?
eg. 
```ruby
status_code = 403
render status: status_code, text: "#{status_code} error"
```

steventux:
@Rosa-Fox I've tested this on my VM and it works perfectly. This is good to merge once we have proven the deploy path for specialist-publisher. :+1:

PR #903, opened 2016-09-13
Fix 'free-form-lists' which were broken by downgrading select2

tuzz:
We recently downgraded select2 to make it use the same styles
from the old Specialist Publisher that our users are familiar
with. In doing so, we broke 'free-form-lists' which are a
bespoke feature used on the dfid-research-outputs format.

These changes make 'free-form-lists' work with the older
select2 version. At present, there is a single 'tags'
array for all 'free-form-lists' that appear on the page.
This means that tags users type in between input fields
will appear as options for others.

This isn't great, but at present, we only have a single
'free-form-list' on the page and the code that stored
separate 'tags' arrays was becoming unreadable. It used
a function as the value of the 'tags' property instead
of a simple array.

### DIFF
--- a/app/assets/javascript/application.js
+++ b/app/assets/javascript/application.js
@@ -9,16 +9,24 @@ jQuery(function($) {
 
   ////
   // Make a select2 that will create new values on return as you type them
-  $(".select2.free-form-list").select2({
-    tags: true,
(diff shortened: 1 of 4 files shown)


Seraphiana:
looks fine. 👍

PR #905, opened 2016-09-15
Alert editors to unsaved changes

steventux:
https://trello.com/c/V8tu3RRf/259-show-a-popup-if-a-user-tries-to-navigate-away-from-this-page-without-saving-their-document-medium

Updates to new or existing documents could be lost without notice
so add a way to inform the user that they have changes to save
before they navigate away from the page or close the current window.

[The javascript class used to track changes to the current form is from Whitehall](https://github.com/alphagov/whitehall/blob/master/app/assets/javascripts/admin/form_change_protection.js).

screenshot from 2016-09-15 15 47 53

Worth noting that the custom message support has changed in some browsers recently, this doesn't affect behaviour but the message displayed is variable depending on the browser.

https://www.chromestatus.com/feature/5349061406228480
https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload#Notes

### DIFF
--- a/app/views/documents/edit.html.erb
+++ b/app/views/documents/edit.html.erb
@@ -5,12 +5,18 @@
     <li class='active'>Edit document</li>
 <% end %>
 
+<% content_for :document_ready do -%>
+  GOVUK.formChangeProtection.init($('.edit_document'), 'You have unsaved changes that will be lost if you leave this page.');
+<% end -%>
+
 <h2>Editing <%= @document.title %></h2>
 
 <div class="row">
   <div class="col-md-8">
     <%= form_for @document, url: document_path(document_type_slug: params[:document_type_slug],
-                                               content_id: @document.content_id), method: :put do |f| %>
+                                               content_id: @document.content_id),
+                                               method: :put,
+                                               html: { class: "edit_document" } do |f| %>
         <%= render partial: "shared/form_fields", locals: {f: f} %>
 
         <%= render partial: "shared/preview_govspeak", locals: {attachments: @document.attachments} %>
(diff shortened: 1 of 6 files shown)


app/views/documents/edit.html.erb
```diff
@@ -5,12 +5,18 @@
     <li class='active'>Edit document</li>
 <% end %>
 
+<% content_for :document_ready do -%>
+  GOVUK.formChangeProtection.init($('.edit_document'), 'You have unsaved changes that will be lost if you leave this page.');
+<% end -%>
```
tuzz, line 10: 👍

spec/features/editing_a_cma_case_spec.rb
```diff
@@ -452,7 +452,7 @@
     }
 
     scenario "showing the update type radio buttons" do
-      within(".new_cma_case") do
+      within(".edit_document") do
```
tuzz, line 455: 👍

spec/features/unsaved_changes_spec.rb
```diff
@@ -0,0 +1,162 @@
...
+        expect(page).to have_content("Updated Amended example document")
+      end
+
+      click_on "Edit document"
+      click_link "Add attachment"
+
+      expect(current_path).to eq("/cma-cases/#{content_id}/attachments/new")
+    end
+  end
+end
```
tuzz, line 162: This is great! Very thorough.

tuzz (approved):
Looks good. It might be worth doing a quick test with a really really big document, like this one:

/cma-cases/8e37c52f-3694-44f9-833b-263726923007

steventux:
@tuzz: I've updated the screenshot to show how this works with the example cma case you mentioned.

PR #907, opened 2016-09-26
Delete attachments

Seraphiana:
Functionality added to make use of the new delete asset endpoint, depends on https://github.com/alphagov/gds-api-adapters/pull/588

it currently looks like this:
screen shot 2016-09-27 at 11 48 20
screen shot 2016-09-27 at 11 48 29
screen shot 2016-09-27 at 11 48 43
screen shot 2016-09-27 at 11 48 53

note: from the work on this story, it seems that attachments policy permissions aren't being used anywhere in the application. I'm not entirely sure why they're in the app at all.

[Trello card](https://trello.com/c/zLrCEavw/263-delete-attachments-feature-medium)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,14 +1,3 @@
-PATH
-  remote: ../gds-api-adapters
-  specs:
-    gds-api-adapters (36.0.0)
-      link_header
-      lrucache (~> 0.1.1)
-      null_logger
-      plek (>= 1.9.0)
-      rack-cache
-      rest-client (~> 2.0)
-
 GEM
   remote: https://rubygems.org/
   specs:
@@ -105,6 +94,13 @@ GEM
     foreman (0.74.0)
       dotenv (~> 0.11.1)
       thor (~> 0.19.1)
+    gds-api-adapters (36.2.0)
+      link_header
+      lrucache (~> 0.1.1)
+      null_logger
+      plek (>= 1.9.0)
+      rack-cache
+      rest-client (~> 2.0)
     gds-sso (11.0.0)
       multi_json (~> 1.0)
       oauth2 (~> 1.0)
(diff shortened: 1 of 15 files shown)


Gemfile.lock
```diff
@@ -365,7 +365,7 @@ DEPENDENCIES
   database_cleaner (= 1.5.1)
   factory_girl
   foreman (= 0.74.0)
-  gds-api-adapters (= 32.1.0)
+  gds-api-adapters (= 36.0.1)
```
steventux: This will presumably need to be updated once https://github.com/alphagov/gds-api-adapters/pull/581 and https://github.com/alphagov/gds-api-adapters/pull/587 are merged and released in a newer version of the gem.

app/policies/document_policy.rb
```diff
@@ -8,6 +8,8 @@ def index?
   alias_method :edit?, :index?
   alias_method :update?, :index?
   alias_method :show?, :index?
+  #TODO fix this, attachments are using the wrong policy
```
steventux: Is this still true?

steventux (commented):
All good, we obviously need the relevant gds-api-adapters gem release before this can be merged.
Could you add a screenshot as it's an admin UI change?

Gemfile.lock
Seraphiana: yup, that's true. I think otherwise that's the only thing that needs work to make it work

app/policies/document_policy.rb
Seraphiana: sadly yes. The attachments policy is completely unused by the entire application, at the moment. I'm not very familiar with how our permissions works, and didn't want to be digging into it too much

steventux: OK I don't think this should block this work but it would be good to understand why the wrong policy is in use and add the reason to a FIXME comment so we can revisit this.

steventux:
This looks good, still pointing at the older gds-api-adapters gem version hence the CI failure

PR #908, opened 2016-09-28
Update name for Democratic Republic of the Congo

Davidslv:
**trello**: https://trello.com/c/rxjklHjL/191-country-name-change-democratic-republic-of-the-congo

> The country formerly known as 'Democratic Republic of Congo' is now officially called 'Democratic Republic of **the** Congo'.


steventux (commented):
:+1:

PR #909, opened 2016-09-28
Don't generate HTML in Specialist Publisher

carlosmartinez:
The bit of [that Trello card](https://trello.com/c/NIP6YIm5/803-govspeak-stop-rendering-html-in-specialist-publisher-stage-3-large) where we stop generating HTML in Specialist Publisher, and where we instead prepare the images and inline attachments so that the Publishing API understands them.


Gemfile.lock
```diff
@@ -395,4 +408,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.10.6
+   1.13.1
```
kevindew, line 411: This might cause us a problem. See: https://github.com/alphagov/publishing-api/pull/536#pullrequestreview-2506357 we have 1.11.2 in govuk_puppet.

app/presenters/specialist_publisher_body_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def image_replacement(identifier)
+    "![InlineAttachment:#{content_id_or_filename(identifier)}]"
+  end
+
+  def content_id_or_filename(identifier)
+    attachment = document.attachments.detect do |att|
+      att.content_id == identifier
+    end
+    attachment ? attachment.url.split("/").last : identifier
```
kevindew, line 60: I was a little concerned here that we could get tripped up by a filename that has characters in it like `]` but as far as I can tell they're [stripped out](https://specialist-publisher.integration.publishing.service.gov.uk/drug-safety-updates/37ae4a4e-3221-4965-ab11-ec0ab3bd9dc1/attachments/98422790-312a-4033-9164-8f73f1636f2c/edit) - do you think it's a risk here at all? I can't seem to find a sanitise function used in the [Attachment class](https://github.com/alphagov/specialist-publisher-rebuild/blob/master/app/models/attachment.rb) so I don't think this is a concern the specialist publisher team had.

spec/presenters/attachment_presenter_spec.rb
```diff
@@ -1,11 +1,12 @@
...
 
 RSpec.describe AttachmentPresenter do
   let(:content_id) { SecureRandom.uuid }
   let(:attachment) {
-    Attachment.new(url: 'path/to/file/in/asset/manger',
-      content_type: 'application/jpeg',
-      title: 'test specialist document attachment',
+    Attachment.new(url: "path/to/file/in/asset/manger",
+      content_type: "application/jpeg",
+      title: "test specialist document attachment",
```
kevindew, line 9: Oh yeah

spec/presenters/govspeak_body_presenter_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'spec_helper'
+
+RSpec.describe GovspeakBodyPresenter do
+  describe '#present' do
+    let(:document) { double(:document, body: body, attachments: attachments) }
```
kevindew, line 5: ⬆️ Quotes whoopsy

kevindew: Oh crap, that was me.

kevindew:
Looking good 👍 I'll fix my whoopsie

spec/presenters/govspeak_presenter_spec.rb
```diff
@@ -6,73 +6,66 @@
...
+        let(:content) { "[embed:attachments:image:foo.jpg]" }
+
+        it "replaces InlineAttachment syntax and references the filename" do
+          expect(presented).to eq expected
+        end
+      end
+    end
+
+    context "when the document has inline attachments" do
+      let(:body)        { "[InlineAttachment:foo.pdf]" }
```
steventux, line 46: Should we have a more realistic Govspeak body text?
The underlying regex is pretty straightforward but it would be good to test how it handles multiple inline attachments embedded in more complex Govspeak.

carlosmartinez: thanks @steventux - i've now added a test at line 71 that is hopefully a bit more realistic.

steventux: :+1: Perfect

steventux:
This looks good to me, if you are both happy with the changes feel free to merge when you are ready :+1:

PR #910, opened 2016-09-30
Reverting the value for Democratic Republic of the Congo

Davidslv:
During PR: https://github.com/alphagov/rummager/pull/700
It was agreed to keep the value so search continues to work.
This PR reverts 5895be401ad12789d4a2d9cf4e064254c67de059

Partially reverts https://github.com/alphagov/specialist-publisher-rebuild/pull/908

\cc @steventux


steventux (approved):
Thanks for spotting and reverting this. :+1:

PR #914, opened 2016-10-11
Change date field format

Rosa-Fox:
[Trello](https://trello.com/c/xbySvXdM/278-improve-date-field-to-be-more-user-friendly-medium)
- Improve date fields to be more user friendly

Before:

dates-creating-new-aaib-previous-way

After:

dates-creating-new-aaib-report-new-look

Based on the styling of [memorable dates](https://www.gov.uk/service-manual/user-centred-design/resources/patterns/dates.html#memorable-dates) but styled to fit with Specialist Publisher's look.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -55,8 +55,9 @@ class DocumentsController < ApplicationController
   def update
     new_params = filtered_params(params[current_format.document_type])
 
-    new_params.each do |k, v|
-      @document.public_send(:"#{k}=", v)
+    new_params.each do |k|
+      key = k[0].gsub(/\(\di\)$/, "")
+      @document.public_send(:"#{key}=", Document.param_value(new_params, key))
     end
 
     if @document.valid?
--- a/app/views/shared/_date_fields.html.erb
+++ b/app/views/shared/_date_fields.html.erb
@@ -1,0 +1,22 @@
+<%= render layout: "shared/form_group", locals: { f: f, field: field } do %>
+<% date = @document.public_send(field) %>
+  <div class="date-layout">
+    <p class="form-hint">For example, 2013 03 20</p>
+    <div class="row">
+      <%= f.label :name, 'Year', class: "col-md-2" %>
+      <%= f.label :name, 'Month', class: "col-md-2" %>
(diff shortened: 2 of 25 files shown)


app/views/shared/_date_fields.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <div class="date-layout">
+    <p class="form-hint">For example, 2013 03 20</p>
+    <div class="row">
+      <%= f.label :name, 'Year', class: "col-md-2" %>
+      <%= f.label :name, 'Month', class: "col-md-2" %>
+      <%= f.label :name, 'Day', class: "col-md-2" %>
+    </div>
+    <div class="form-group row">
+      <div class="col-md-2">
+        <%= f.text_field field, name: "[#{format}]#{field}(1i)", class: 'form-control', placeholder: '2013', value: date_value("year", @document.public_send(field)), type: "number" %>
```
steventux: Would it make sense to only evaluate `@document.public_send(field)` once? You could either set this in the erb template:

``` ruby
<% date = @document.public_send(field) %>
```

or change the `date_value` method signature:

``` ruby
def date_value(measure, document, field)
  date = document.public_send(field)
  ...
```

steventux (commented):
This looks good, I think the `date_value` method/invocation could be simplified a bit as mentioned but it's a good solution for all document types and different rails date input elements. :+1:

app/views/shared/_date_fields.html.erb
Rosa-Fox: Thanks @steventux :) I have amended.

app/controllers/documents_controller.rb
```diff
@@ -55,8 +55,9 @@ def edit; end
   def update
     new_params = filtered_params(params[current_format.document_type])
 
-    new_params.each do |k, v|
-      @document.public_send(:"#{k}=", v)
+    new_params.each do |k|
+      key = k[0].gsub(/\(\di\)$/, "")
+      @document.public_send(:"#{key}=", Document.param_value(new_params, key))
```
steventux, line 60: Not part of this PR but we should whitelist they keys here. (thanks @dougdroper)

PR #915, opened 2016-10-13
Validation for dates which do not exist

Rosa-Fox:
- Product manager noticed that if she added a date such as '31 Feb 2016' (which is obviously a wrong date), the error "something went wrong" displayed

This PR:
- limits the range of numbers that can be entered into the date form fields
- Gives a validation error if a date is not valid
- We have discussed that this could be further improved by not forcing the user to enter the format yyyy-mm-dd. We should allow single digits to be entered for day and month to improve user experience. This PR serves as a quick fix for now and will be further improved at a later date.

### DIFF
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -1,21 +1,7 @@
 class DateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    record.errors.add(attribute, error_message) unless parsable_date?(value)
-  end
-
-private
-
-  def parsable_date?(date_string)
-    date_string =~ iso8601_regex && Date.parse(date_string)
-  rescue ArgumentError
-    false
-  end
-
-  def iso8601_regex
-    /\A[0-9]{4}\-[0-9]{2}\-[0-9]{2}\z/
-  end
-
-  def error_message
-    options[:message] || "should be formatted YYYY-MM-DD"
+    Date.parse(value)
+  rescue ArgumentError, RangeError
+    record.errors.add(attribute, "is not a valid date")
   end
 end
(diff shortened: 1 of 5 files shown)


app/validators/date_validator.rb
```diff
@@ -15,6 +23,10 @@ def iso8601_regex
     /\A[0-9]{4}\-[0-9]{2}\-[0-9]{2}\z/
   end
 
+  def invalid_date_error_message
```
steventux: We don't really need methods to return this string and the `error_message` string if these are only used once when added to the errors collection.

app/validators/date_validator.rb
```diff
@@ -1,12 +1,20 @@
 class DateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    record.errors.add(attribute, error_message) unless parsable_date?(value)
+    if incorrect_format?(value)
```
steventux: These should be single if conditions so that all validation errors are populated at once.

app/validators/date_validator.rb
```diff
@@ -1,12 +1,20 @@
...
+      record.errors.add(attribute, error_message)
+    elsif !parsable_date?(value)
+      record.errors.add(attribute, invalid_date_error_message)
+    end
   end
 
 private
 
+  def incorrect_format?(date_string)
+    !(iso8601_regex === date_string)
```
steventux: I don't think we should force the user to zero pad days and months.

app/validators/date_validator.rb
```diff
@@ -1,12 +1,20 @@
...
+  end
+
   def parsable_date?(date_string)
-    date_string =~ iso8601_regex && Date.parse(date_string)
+    iso8601_regex === date_string && Date.parse(date_string)
```
steventux: This is repeating the implementation of `incorrect_format?`, not sure that is necessary.

steventux (changes requested):
I think the only validation we need is that the date parts constitute a valid date, forcing the user to zero pad the date is unnecessary, it seems like an implementation that suits the code more than the user.

steventux (approved):

PR #917, opened 2016-10-18
Warn user if document path is already taken

steventux:
https://trello.com/c/xiUYc6Y1/280-warn-the-user-and-disable-publish-if-a-document-s-path-is-already-taken-medium

There are a couple of scenarios where Specialist Publisher may attempt to save a draft with a path that is already taken:
1. A draft is present in the Publishing API with the same `base_path`
2. A published item is present in the Publishing API with the same `base_path`

In the case of a draft being present, the Publishing API responds with an error, in the case of a publishing item being present a warning is returned as part of a successful response.

This PR attempts to propagate these errors and warnings to the editor or writer by storing them on the document and displaying as part of the relevant interaction in the UI.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -113,7 +113,6 @@ private
       list_items = @document.errors.full_messages.map do |message|
         content_tag(:li, message.html_safe)
       end
-
       list_items.join.html_safe
     end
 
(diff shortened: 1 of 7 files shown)


Rosa-Fox:
Hi @steventux, looks good. I added a commit for the error message on save so let me know what you think.

Davidslv:
@Rosa-Fox I think you can setup that on the Document model, using `errors.add` http://guides.rubyonrails.org/active_record_validations.html#errors-add

Rosa-Fox:
Thanks @Davidslv. Have amended.

PR #919, opened 2016-10-19
Date formatting

Rosa-Fox:
- 0 pads single digit month and day date inputs to improve consistency 
- Strengthen date validation
- Improve number range that users can input in date fields

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -62,7 +62,16 @@ class Document
         params.fetch(:"#{key.to_s}(2i)"),
         params.fetch(:"#{key.to_s}(3i)")
       ]
-      format_date.delete_if(&:empty?).join("-")
+      format_date.delete_if(&:empty?)
+
+      format_date = format_date.map do |date|
+        if /^\d{1}$/ =~ date
+          sprintf('%02d', date)
+        else
+          date
+        end
+      end
+      format_date.join("-")
     else
       params.fetch(key, nil)
     end
--- a/spec/features/editing_a_cma_case_spec.rb
+++ b/spec/features/editing_a_cma_case_spec.rb
@@ -124,6 +124,17 @@ RSpec.feature "Editing a CMA case", type: :feature do
       assert_publishing_api_put_content(content_id, request_json_includes(changed_json))
     end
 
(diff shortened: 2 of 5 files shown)


app/models/document.rb
```diff
@@ -62,7 +62,16 @@ def self.param_value(params, key)
         params.fetch(:"#{key.to_s}(2i)"),
         params.fetch(:"#{key.to_s}(3i)")
       ]
-      format_date.delete_if(&:empty?).join("-")
+      format_date.delete_if(&:empty?)
```
steventux, line 65: I think it would be good to move some of this increasingly date-centric logic into `DateHelper` _somehow_ not sure that should block this PR though.

spec/features/editing_a_cma_case_spec.rb
```diff
@@ -124,6 +124,15 @@
       assert_publishing_api_put_content(content_id, request_json_includes(changed_json))
     end
 
+    scenario "date values display" do
+      visit "/cma-cases/#{content_id}"
+
+      expect(page).to have_content("2014")
```
steventux: Could we use more specific selectors here? Something like:

``` ruby
expect(page).to have_selector("input[name=[cma_case]opened_date(1i)]", with: "2014")
```

Also I think the labels for these inputs all have the wrong 'for' attribute when you inspect them

app/models/document.rb
```diff
@@ -62,7 +62,16 @@ def self.param_value(params, key)
...
+
+      format_date = format_date.map do |date|
+        if /^\d{1}$/ =~ date
+          sprintf('%02d', date)
```
steventux, line 69: I think this can be simplified to:

``` ruby
sprintf("%04d-%02d-%02d", *format_date)
```

perhaps inside a check for an empty `format_date` array?

app/models/document.rb
steventux: I'm working on refactoring the date centric stuff into DateHelper so don't worry about that in this PR

app/models/document.rb
steventux: I'll make this change when refactoring date logic.

PR #921, opened 2016-10-20
Avoid date parsing when re-presenting date input

steventux:
Date validation is done elsewhere in the `Document` model so
avoid parsing the form/record value when refilling date fields.
The date may be nonsensical in which case we want the `DateValidator`
to catch this.


steventux:
Closing in favour of https://github.com/alphagov/specialist-publisher/pull/922

PR #925, opened 2016-10-24
Update tribunal_decision_judges for UTAAC finder

lgarvey:
The list of judges has changed for the UTAAC finder

PR #924, opened 2016-10-24
Update tribunal_decision_categories for Employment Tribunal finder

lgarvey:
Amends to the ET category list

PR #926, opened 2016-10-24
Update email alert options for the Employment Tribunal Decisions finder

lgarvey:
A change to the email categories for the MOJ ET finder

### DIFF
--- a/lib/documents/schemas/employment_tribunal_decisions.json
+++ b/lib/documents/schemas/employment_tribunal_decisions.json
@@ -41,6 +41,12 @@
       "prechecked": false
     },
     {
+      "key": "breach-of-contract",
+      "radio_button_name": "Breach of Contract",
+      "topic_name": "Breach of Contract",
+      "prechecked": false
+    },
+    {
       "key": "central-arbitration-committee-cac",
       "radio_button_name": "Central Arbitration Committee (CAC)",
       "topic_name": "central arbitration committee cac",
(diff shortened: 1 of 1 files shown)


tijmenb:
This PR has some of the same changes in utaac_decisions.json as https://github.com/alphagov/specialist-publisher/pull/925, can you remove those?

tijmenb:
For some context, this PR should change the options for the email alerts for the finder.

I've published the finder on integration using the rake task:

https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/459/console

Which will make it visible in integration:

https://www-origin.integration.publishing.service.gov.uk/employment-tribunal-decisions

It currently shows this when you click the "Subscribe to email alerts" button:

screen shot 2016-10-24 at 19 38 30

I think the way that this is code reviewed is by checking that the `key` in each email alert choice corresponds to a valid `tribunal_decision_categories` (as configured by the [email_filter_by option](https://github.com/alphagov/specialist-publisher/blob/64fda6bb6b20af650924b6bb7a8ed3398b0a91f5/lib/documents/schemas/employment_tribunal_decisions.json#L23)).

I will check tomorrow if this is something we can do automatically. Otherwise the danger exists that users subscribe to email alerts that will never be sent out.

tijmenb:
This now fails correctly because one of the categories doesn't exist yet. Once https://github.com/alphagov/specialist-publisher/pull/924 is merged this will be green and can get merged.

tijmenb:
I've squashed this commit and force-pushed.

The tests don't pass because `employment-agencies-act-1973-obsolete-topic` is not a valid thing anymore (it was removed in https://github.com/alphagov/specialist-publisher/pull/924).

There are a bunch of other "obsolete" topics still, do we want to remove them?

lgarvey:
All the -obsolete-topic topics have been removed from the ET finder.  That
convention may have been used elsewhere in others finders, so if possible
only tests relating to the ET finder should be amended.

tijmenb:
This should be rebased after https://github.com/alphagov/specialist-publisher/pull/932 is merged.

tijmenb:
Rebased & all passing. Merging.

PR #927, opened 2016-10-24
Allow multiple values for tribunal_decision_category in Asylum Finder

lgarvey:
Amend the asylum finder category and sub category to allow multiple
types

### DIFF
--- a/app/models/asylum_support_decision.rb
+++ b/app/models/asylum_support_decision.rb
@@ -1,19 +1,21 @@
 class AsylumSupportDecision < Document
-  validates :tribunal_decision_category, presence: true
+  validates :tribunal_decision_categories, presence: true
   validates :tribunal_decision_decision_date, presence: true, date: true
   validates :tribunal_decision_judges, presence: true
   validates :tribunal_decision_landmark, presence: true
   validates :tribunal_decision_reference_number, presence: true
-  validates :tribunal_decision_sub_category, presence: true, asylum_support_decision_sub_category: true
+  validates :tribunal_decision_sub_categories, presence: true, asylum_support_decision_sub_category: true
 
   FORMAT_SPECIFIC_FIELDS = %i(
       hidden_indexable_content
       tribunal_decision_category
+      tribunal_decision_categories
       tribunal_decision_decision_date
       tribunal_decision_judges
       tribunal_decision_landmark
       tribunal_decision_reference_number
       tribunal_decision_sub_category
+      tribunal_decision_sub_categories
   ).freeze
 
   attr_accessor(*FORMAT_SPECIFIC_FIELDS)
--- a/app/views/metadata_fields/_asylum_support_decisions.html.erb
+++ b/app/views/metadata_fields/_asylum_support_decisions.html.erb
@@ -1,15 +1,27 @@
 <%= render layout: "shared/form_group", locals: {f: f, field: :tribunal_decision_category} do %>
   <%= f.select :tribunal_decision_category,
-     f.object.facet_options(:tribunal_decision_category),
+     f.object.facet_options(:tribunal_decision_categories),
      {},
-     {class: 'form-control'}
+     {
+         class: 'select2 form-control',
+         multiple: true,
+         data: {
+             placeholder: 'Select judges'
+         }
+     }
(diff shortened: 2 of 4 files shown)


tijmenb:
The way we do contract testing limits our ability to change the type of attributes in content items. 
Because schemas have to be backwards compatible, a test for a content schema change won't pass in CI. And without that change this PR won't pass in CI.

Also, I suspect that changing the type might cause problems in either elasticsearch or in the app itself.

I think the best thing to do here is to add `tribunal_decision_categories` and `tribunal_decision_sub_categories` attribute to the content schemas and rummager. This has the added benefit of having the correct pluralized name.

We can optionally copy over the current values in a rake task if we want to maintain the data (there are 83 `asylum_support_decision` documents).

I suspect that the same change has to be done for the Tax Tribunal decisions?

lgarvey:
That makes sense.

We haven't done any work with tax tribunals - so it's unlikely to need
amending.

app/models/asylum_support_decision.rb
```diff
@@ -1,19 +1,19 @@
...
   validates :tribunal_decision_judges, presence: true
   validates :tribunal_decision_landmark, presence: true
   validates :tribunal_decision_reference_number, presence: true
-  validates :tribunal_decision_sub_category, presence: true, asylum_support_decision_sub_category: true
+  validates :tribunal_decision_sub_categories, presence: true, asylum_support_decision_sub_category: true
 
   FORMAT_SPECIFIC_FIELDS = %i(
       hidden_indexable_content
-      tribunal_decision_category
+      tribunal_decision_categories
```
tijmenb, line 12: If you keep this `tribunal_decision_category` and `tribunal_decision_sub_category` here temporarily, we can use it for the migration.

tijmenb (changes requested):
This looks good. Can you squash the commits together?

tijmenb (approved):

lgarvey:
@tijmenb don't merge this yet. I see a missing sub category field in
the FORMAT_SPECIFIC_FIELDS
array. Will fix on Monday.

lgarvey:
@tijmenb I've added back in the missing tribunal_decision_sub_categories field, so this looks good to merge now.

app/views/metadata_fields/_asylum_support_decisions.html.erb
```diff
@@ -1,15 +1,27 @@
 <%= render layout: "shared/form_group", locals: {f: f, field: :tribunal_decision_category} do %>
   <%= f.select :tribunal_decision_category,
```
tijmenb, line 2: This needs to read `:tribunal_decision_categories`.

app/views/metadata_fields/_asylum_support_decisions.html.erb
```diff
@@ -1,15 +1,27 @@
...
-     f.object.facet_options(:tribunal_decision_category),
+     f.object.facet_options(:tribunal_decision_categories),
      {},
-     {class: 'form-control'}
+     {
+         class: 'select2 form-control',
+         multiple: true,
+         data: {
+             placeholder: 'Select judges'
```
tijmenb, line 9: This should probably be "Select categories"

app/views/metadata_fields/_asylum_support_decisions.html.erb
```diff
@@ -1,15 +1,27 @@
...
   <%= f.select :tribunal_decision_sub_category,
-     f.object.facet_options(:tribunal_decision_sub_category),
+     f.object.facet_options(:tribunal_decision_sub_categories),
      {},
-     {class: 'form-control'}
+     {
+         class: 'select2 form-control',
+         multiple: true,
+         data: {
+             placeholder: 'Select judges'
```
tijmenb, line 22: Same here.

tijmenb:
For this PR to work we also need to change

https://github.com/alphagov/specialist-publisher/blob/master/lib/documents/schemas/asylum_support_decisions.json#L49

and

https://github.com/alphagov/specialist-publisher/blob/master/lib/documents/schemas/asylum_support_decisions.json#L71

That way the dropdown will be populated.

PR #931, opened 2016-10-26
Restrict attachment filetypes

Rosa-Fox:
[Trello](https://trello.com/c/HiHWa2pE/281-there-are-no-restrictions-on-which-filetypes-can-be-uploaded-small)

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -12,15 +12,14 @@ class AttachmentsController < ApplicationController
 
   def create
     document = fetch_document
-
     if attachment_valid?
+
       attachment = document.attachments.build(attachment_params)
       attachment.content_type = attachment.file.content_type
 
       upload_attachment(document, attachment)
     else
-      flash[:danger] = "Adding an attachment failed. Please make sure you have uploaded an attachment."
-      redirect_to new_document_attachment_path(document_type_slug, document.content_id)
+      failed_to_attach(document)
     end
   end
 
@@ -36,8 +35,10 @@ class AttachmentsController < ApplicationController
 
     if attachment.file.nil?
       save_updated_title(document, attachment)
-    else
+    elsif white_listed?
       update_attachment(document, attachment)
+    else
+      failed_to_attach(document)
     end
   end
 
@@ -117,6 +118,16 @@ private
   end
 
   def attachment_valid?
-    !params["attachment"].nil? && !params["attachment"]["file"].nil?
+    attachment = params["attachment"]
+    attachment.present? && attachment["file"].present? && white_listed?
+  end
+
+  def white_listed?
+    Attachment.valid_filetype?(params["attachment"]["file"])
+  end
+
+  def failed_to_attach(document)
+    flash[:danger] = "Adding an attachment failed. Please make sure you have uploaded an attachment of a permitted file type."
+    redirect_to new_document_attachment_path(document_type_slug, document.content_id)
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/attachments_controller.rb
```diff
@@ -119,4 +121,14 @@ def attachment_params
   def attachment_valid?
     !params["attachment"].nil? && !params["attachment"]["file"].nil?
   end
+
+  def ensure_extension_white_listed
```
steventux: This seems to repeat some of the checking done in `attachment_valid?` and I wonder if it would be easier to have all checks done within that method and not use a before filter (sorry I know this was my original suggestion).
You could probably condense the logic down to something like:

``` ruby
def attachment_valid?
  attachment = params["attachment"]
  attachment.present? && attachment["file"].present? && Attachment.valid_filetype(attachment)
end
```

app/controllers/attachments_controller.rb
```diff
@@ -119,4 +121,14 @@ def attachment_params
...
+    document = fetch_document
+    file = params.fetch(:attachment, :file)
+    return if file["file"].blank?
+    return if Attachment.valid_filetype(file)
+
+    flash[:danger] = "Invalid file type: #{File.extname(file['file'].tempfile)}"
```
steventux: Given my previous comment you could just append this message to the existing invalid attachment error message:

``` ruby
flash[:danger] = "Adding an attachment failed. Please make sure you have uploaded an attachment of a permitted file type."
```

steventux (changes requested):
I think we can squish this into the same `attachment_valid?` method - let me know if you think this won't work though.

Rosa-Fox:
Thanks @steventux, I have made the changes. Let me know what you think.

spec/features/editing_a_cma_case_spec.rb
```diff
@@ -275,7 +275,7 @@
         {
           "content_id" => "77f2d40e-3853-451f-9ca3-a747e8402e34",
           "url" => "https://assets.digital.cabinet-office.gov.uk/media/513a0efbed915d425e000002/asylum-support-image.jpg",
-          "content_type" => "application/jpeg",
+          "content_type" => "application/jpg",
```
steventux, line 278: Are we explicitly preventing upload of files with the `.jpeg` extension?

app/controllers/attachments_controller.rb
```diff
@@ -117,6 +118,17 @@ def attachment_params
   end
 
   def attachment_valid?
-    !params["attachment"].nil? && !params["attachment"]["file"].nil?
+    if !params["attachment"].nil? && !params["attachment"]["file"].nil?
```
steventux: This can return `true`, `false` or `nil`, I think it would be more idiomatic to return simply `true` or `false`. You could also condense it to:

``` ruby
attachment = params["attachment"]
attachment.present? && attachment["file"].present? && Attachment.valid_filetype?(attachment["file"])
```

which makes the three conditions always return a boolean.

spec/features/editing_a_cma_case_spec.rb
Rosa-Fox: yes, though that was taken from the whitelist in Whitehall... should we let .jpeg though? I guess it makes sense to?

app/models/attachment.rb
```diff
@@ -21,6 +21,11 @@ def initialize(params = {})
     @being_updated = false
   end
 
+  def self.valid_filetype(file)
```
steventux: Should be called `valid_filetype?` as it returns a boolean

Rosa-Fox: 👍

Rosa-Fox:
@steventux thanks, have re-updated :)

steventux (approved):

PR #932, opened 2016-10-27
Don't allow email alerts for removed categories

tijmenb:
These categories were removed in https://github.com/alphagov/specialist-publisher/pull/924, but were still in the email options. This means people can sign up for emails that will never be sent.

https://github.com/alphagov/specialist-publisher/pull/924 shouldn't have been merged, but the test that checks these kinds of things was added after the PR was raised.

PR https://github.com/alphagov/specialist-publisher/pull/926 will add some other email options.

@lgarvey


rubenarakelyan:
👍

PR #934, opened 2016-10-31
Don't send change history to Publishing API

carlosmartinez:
Publishing API now manages change history, so Specialist Publisher only
needs to send a change note (now part of the top-level schema).


steventux (approved):
:+1: Less code yay

PR #937, opened 2016-11-02
Extract service objects from `Document` class

tijmenb:
This PR extracts all methods with "external" side effects (communicating with an API) into separate classes.

It's an attempt to make the class more easily digestible for the uninitiated. It reduces the number of lines in `Document` from 420 to about 300.

I've tried to keep code changes to a minimum, so all the classes have just one or two class methods. I actually prefer initialisers and instance methods ([because think piece](http://blog.codeclimate.com/blog/2012/11/14/why-ruby-class-methods-resist-refactoring/)) but this was the smallest change that would still make a difference. Further refactoring is left as an exercise to the reader.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -170,61 +170,8 @@ class Document
     end
   end
 
-  def self.extract_body_from_payload(payload)
-    body_attribute = payload.fetch('details').fetch('body')
-
-    case body_attribute
-    when Array
-      govspeak_body = body_attribute.detect do |body_hash|
-        body_hash['content_type'] == 'text/govspeak'
-      end
-      govspeak_body['content']
-    when String
-      body_attribute
-    end
-  end
-
-  def self.set_update_type(document, payload)
-    if document.temporary_update_type?
-      document.update_type = nil
-      document.temporary_update_type = false
-    elsif document.published? || document.unpublished?
-      document.update_type = nil
-    elsif document.first_draft?
-      document.update_type = 'major'
-    else
-      document.update_type = payload["update_type"]
-    end
-  end
-
   def self.from_publishing_api(payload)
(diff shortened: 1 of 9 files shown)


app/models/document.rb
```diff
@@ -230,10 +230,7 @@ def publish
 
   def unpublish
     handle_remote_error do
-      Services.publishing_api.unpublish(content_id, type: 'gone')
-
-      AttachmentDeleteWorker.perform_async(content_id)
-      RummagerDeleteWorker.perform_async(base_path)
+      DocumentUnpublisher.unpublish(content_id, base_path)
```
carvil, line 203: One thing we tried to do in other apps is something like this:

```
UnpublishDocument.call(content_id, base_path)
```

WDYT?

carvil (changes requested):
👍 good change, only one comment left below.

app/models/document.rb
h-lame: Despite the seeming redundancy with the name of the receiver, I'm in favour of methods with names that describe what they do.  It's not clear what a method called `call` actually does without reading the method, whereas a method called `unpublish` gives us a hint.

tijmenb: That's my thinking too. Giving the method a descriptive name forces more thought about the behaviour. It also communicates intent in places where the receiver isn't known (`expect(described_class.call).to eql(...)`).

carvil: The idea was that the class name explains what the service is responsible for (`UnpublishDocument`), and `call` makes it easier to call these services without having to check what the method names are (`FindAllDocuments.call`, `PublishDocument.call`, etc).

But I'm happy with both solutions!

carvil (approved):

PR #939, opened 2016-11-02
Fix jenkins after specialist-publisher rename

tijmenb:
This will make the tests run properly.


carvil:
👍 

Do we need to update the jenkins job that triggers all schema tests?

tijmenb:
No, that'll keep working.

PR #941, opened 2016-11-02
Use eager_load_paths to include lib directory.

danielroseman:
Fixes issue with constants not found in production only.


kevindew (approved):
Looks good 👍

PR #942, opened 2016-11-02
Move hashdiff dependency to top level.

danielroseman:
It's required by the DataComparison module so needs to be present in
all environments.


carlosmartinez (approved):
approve36

PR #945, opened 2016-11-04
Add log directory

kevindew:
This prevents errors like "No such file or directory @ rb_sysopen - ./log/sidekiq.json.log" when you start specialist publisher.


cbaines:
:+1:

PR #946, opened 2016-11-04
Faster AttachmentReporter report

kevindew:
This refactors the AttachmentReporter to no longer have a nested loop
and instead use ruby array methods to compare arrays. It is
substantially quicker than the previous code and thus should resolve the
timeout errors that have been occurring.

We also make more clear attachments that have matching snippets.


steventux:
:+1: This looks good to me. Did testing this branch on integration work as expected?

kevindew:
Yup, it was able to save without timeout. Still kinda slow at about 2s but a good 13s quicker than before.

PR #947, opened 2016-11-04
Use YAML instead of JSON for the finder configuration

tijmenb:
I think YAML is better here because it allows whitespace to improve readability and it allows comments.

It also makes clear that the file here is not the finder content item,  but rather a kind of configuration file. Making this clearer will allow us to move more behaviour into this file, rather than the class.

Files converted with:

```ruby
require 'yaml'
require 'json'

Dir.glob("./lib/documents/schemas/*.json").each do |filename|
  data = YAML.dump(JSON.parse(File.read(filename)))
  File.write(filename.sub('.json', '.yml'), data)
  File.unlink(filename)
end
```


tijmenb:
Will reopen once I have time and the tests pass.

PR #948, opened 2016-11-07
Faster Govspeak preview

kevindew:
This improves the performance of generating a Govspeak preview and makes the rendering of the preview more consistent with Govspeak rendering.

For a large document with many attachments we're still getting around ~3s of time to render the Govspeak into HTML, to improve this we'd have to look inside the Govspeak gem and see if there are any boosts we can give it there, or it could be that kramdown itself is the bottleneck.


dougdroper:
Looks good to me, it is just failing on CI.

kevindew:
@dougdroper green again now after fighting a broken CI server and a broken Jenkins configuration

PR #950, opened 2016-11-09
Fix asylum finder category config

lgarvey:
The asylum finder schema was still refering to the older *_category
fields rather than the new multi select *_categories variants.

Follow up to https://github.com/alphagov/specialist-publisher/pull/927.

### DIFF
--- a/app/views/metadata_fields/_asylum_support_decisions.html.erb
+++ b/app/views/metadata_fields/_asylum_support_decisions.html.erb
@@ -6,7 +6,7 @@
          class: 'select2 form-control',
          multiple: true,
          data: {
-             placeholder: 'Select judges'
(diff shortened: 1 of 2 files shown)


lgarvey:
@tijmenb this should rectify the problem.

tijmenb:
The tests fail because the email alert configuration (`email_filter_by`) is still relying on `tribunal_decision_category`.

lgarvey:
OK just amended this. I'll squash all these commits together when we're
happy they're complete.

tijmenb:
👍 I think the commits are okay as-is, they tell what's being updated. I've updated the PR description to https://github.com/alphagov/specialist-publisher/pull/927 for context.

🚀

PR #951, opened 2016-11-14
Use Rails.root to remove relative resolution of path

kevindew:
Without using Rails.root this will based off of the working directory of
the application. The reason this hasn't presented any problems before is
because this is nearly always the value of the Rails root.

However if we utilise the rails server feature of running this as a
daemon the value of the working directory is "/" which causes this path
to be looked for in the wrong place.

### DIFF
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -111,7 +111,7 @@ private
   end
 
   def view_renderer
-    ActionView::Base.new("app/views")
+    ActionView::Base.new(Rails.root.join("app/views"))
   end
 
   def header


cbaines:
I didn't know about the `--daemon` flag... do you know what it does, as not even looking at the source is clear?

Its reasonably normal to set the working directory of a daemon, so while I think this is a good change, I'm not sure it specifically relates to running it as a daemon.

cbaines:
Ah, just tried it out, and it seemed to fork to run in the background.

:+1: on this change, but you may want to run the process from the rails root anyway.

kevindew:
Yeah I came across this when running the Rails server in the background as part of the Puppet spike. If you're in say `/var/govuk/specialist-publisher` and run `rails s` Ruby sees your cwd as `/var/govuk/specialist-publisher` whereas if you run `rails s -d` it see's your cwd as `/`.

When the app is run through unicorn the cwd is correct - hence why this hasn't been a problem. 

I'm working on the basis though that if a standard rails command causes this problem, then it could catch someone out.

PR #957, opened 2016-11-30
Paginate specialist document finders

tijmenb:
The finders published by this application have a lot of documents in them, but aren't paginated. The number 50 is a bit arbitrary, but seems alright for specialist users. We'll probably do more user research into finder behaviour later next year, which might influence the number of results.

From Tara:

> As they're more specialist, maybe 50 would be better than 20, to help the type of users who want to see everything about a thing. But we've got no analytics evidence unless we either track the positions or work them out from the URLs clicked.

https://trello.com/c/VKqW8wEV

### DIFF
--- a/app/presenters/finders/finder_content_item_presenter.rb
+++ b/app/presenters/finders/finder_content_item_presenter.rb
@@ -46,6 +46,7 @@ private
       summary: file.fetch("summary", nil),
       facets: file.fetch("facets", nil),
       default_order: file.fetch("default_order", nil),
+      default_documents_per_page: 50,
     }.reject { |_, value| value.nil? }
   end
 


fofr:
:100: How much quicker does this make it?

tijmenb:
I'm not sure if this is the scientific way of measuring frontend performance, but it looks like this on my machine:

## Before

screen shot 2016-11-30 at 16 07 52

## After

screen shot 2016-11-30 at 16 08 48

rubenarakelyan:
🎉

NickColley:
@tijmenb nice! If you wanted to test more real world connections in the future you could try using https://www.webpagetest.org/ and throttle your test to 2g-3g.

You can also do something similar (but not as accurate) by throttling connections in Chrome Dev Tools. https://developers.google.com/web/tools/chrome-devtools/network-performance/network-conditions

PR #958, opened 2016-11-30
[Don't merge] Search by elasticsearch_type

tijmenb:
We are trying to remove usage of `document_type` in rummager, because
it's easily confused with the `document_type` in the content-store.

https://github.com/alphagov/rummager/pull/722 allows searching with `elasticsearch_type`. This PR can be merged after that is deployed.

https://trello.com/c/C8A9bnuO/344-work-on-making-the-document-type-concept-clearer


tijmenb:
This is hard to change because `document_type` is entangled with the email system. Needs some further thought.

PR #959, opened 2016-12-01
Make Utaac Decisions and Tax Tribunal Finders live

sihugh:
This removes the pre-production flag so these will be visible in production.

Also adding a rake task that will persist until we've got MOJ live.  The documents
referred to in the CSV files have been pre-checked and approved for publication.

We're publishing the items listed in the CSV files so that we can make the
UTAAC and Tax Tribunal finders live later on.

We'll remove the CSVs from the repo when we're done.  We have subsequent
finders to go live so we'll retain the rake task/worker for a bit longer.


brenetic:
👍

PR #961, opened 2016-12-06
[Don't merge] Test finder schemas match specialist document schema

Davidslv:
Move the FinderSchemaConverter class from govuk-content-schemas.
We used to have a script there that generated the `metadata` part
of the `specialist_document` content schema, based on the finder
schemas published by specialist publisher:
https://github.com/alphagov/govuk-content-schemas/blob/66b11b4004b49064b6aae16715158740977dc706/bin/convert_finder_schemas

This was a bit out of date, and didn't work with all the schemas we've
created. We've changed this to a set of tests in specialist publisher,
so that consistency is always checked for new schemas, unless they are
explicitly added to the `IGNORED_FINDERS` array.

Trello: https://trello.com/c/lFZnNhs4/349-create-a-service-assessment-finder
Paired with @klssmith @MatMoore

### DIFF
--- a/lib/finder_schema_converter.rb
+++ b/lib/finder_schema_converter.rb
@@ -1,0 +1,186 @@
+# FinderSchemaConverter looks at the finder content schemas, and generates
+# The corresponding metadata fields we would expect to see in the content
+# schema for `specialist_document`.
+#
+# The specialist document schema is used for all document types that can be
+# found by the finders, and it should have fields for everything that can be
+# facetted on.
+class FinderSchemaConverter
+  MULTI_SELECTS_BY_DOCUMENT_TYPE = {
+    "aaib_report" => %w[aircraft_category].freeze,
+    "cma_case" => %w[market_sector].freeze,
+    "countryside_stewardship_grant" => %w[
+      land_use
+      tiers_or_standalone_items
+      funding_amount
+    ].freeze,
+    "drug_safety_update" => %w[therapeutic_area].freeze,
+    "european_structural_investment_fund" => %w[
+      fund_type
+      location
+      funding_source
+    ].freeze,
+    "international_development_fund" => %w[
+      location
+      development_sector
+      eligible_entities
+      value_of_funding
+    ].freeze,
+    "maib_report" => %w[vessel_type].freeze,
+    "medical_safety_alert" => %w[medical_specialism].freeze,
+    "raib_report" => %w[railway_type].freeze,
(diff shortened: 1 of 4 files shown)


spec/lib/schema_consistency_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'govuk_schemas'
+require "finder_schema_converter"
+
+RSpec.describe "the finder schemas" do
+  def specialist_document_schema_definitions
+    GovukSchemas::Schema.find(publisher_schema: "specialist_document")["definitions"]
+  end
+
+  context "all facets on finder schema should appear as metadata in the specialist document schema" do
+    FinderSchemaConverter.except_ignored_finders.each do |finder_schema|
```
tijmenb, line 11: The selection, looping and reading of the finders to test could be moved into here to make `FinderSchemaConverter` do less. You could instantiate the `FinderSchema` in here.

lib/finder_schema_converter.rb
```diff
@@ -0,0 +1,186 @@
+# FinderSchemaConverter looks at the finder content schemas, and generates
+# The corresponding metadata fields we would expect to see in the content
+# schema for `specialist_document`.
+#
+# The specialist document schema is used for all document types that can be
+# found by the finders, and it should have fields for everything that can be
+# facetted on.
+class FinderSchemaConverter
+  MULTI_SELECTS_BY_DOCUMENT_TYPE = {
```
tijmenb, line 9: Would it be possible to get this data from the schemas?

tijmenb (changes requested):
Cool, very useful! `FinderSchemaConverter` has a lot of responsibilities though, perhaps it could be refactored a little.

lib/finder_schema_converter.rb
Davidslv: There might be, but I'm not sure how would we go about it.

spec/lib/schema_consistency_spec.rb
MatMoore: Yeah that sounds reasonable, since the test is the only thing making use of it, and it would make it a bit more obvious what's being excluded.

Davidslv:
I think we should close this for the moment being?

tijmenb:
@Davidslv yeah, seems sensible. Let's revisit once the finder has been shipped.

Davidslv:
Thanks

PR #965, opened 2016-12-12
Use model names from schemas

tijmenb:
This automatically generate the model names from the schemas. 

Makes it easier to add a new specialist document type.


Davidslv (approved):
🍰

PR #970, opened 2016-12-14
Add assessment_date for Service Standard Report

Davidslv:
This depends on https://github.com/alphagov/govuk-content-schemas/pull/462 to be merged first in order for the test suite to be green.

We are adding `assessment_date` as part of the metadata sent to
Publishing API. This means that editors will be able to set the date of a given assessment.

This is a follow up work on moving all the service assessments and self
certification documents away from GDS Data blog
(https://gdsdata.blog.gov.uk/all-service-assessments-and-self-certification/)
and move them to Specialist Publisher.

Trello:
https://trello.com/c/ADPnDXsE/385-add-assessment-date-to-service-standard-specialist-document

Paired with: @klssmith

### DIFF
--- a/app/models/service_standard_report.rb
+++ b/app/models/service_standard_report.rb
@@ -1,4 +1,12 @@
 class ServiceStandardReport < Document
+  FORMAT_SPECIFIC_FIELDS = %i(assessment_date).freeze
+
+  attr_accessor(*FORMAT_SPECIFIC_FIELDS)
+
+  def initialize(params = {})
+    super(params, FORMAT_SPECIFIC_FIELDS)
+  end
+
   def self.title
     "Service Standard Report"
   end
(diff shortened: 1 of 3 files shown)


app/models/service_standard_report.rb
```diff
@@ -1,4 +1,14 @@
 class ServiceStandardReport < Document
+  validates :assessment_date, presence: true
```
tijmenb: I thought we talked about this not being required?

tijmenb (changes requested):
Looks good, apart from the requiredness. If we make the attribute required at the moment we won't be able to import data -- it's something the editors have to set manually.

Davidslv:
@tijmenb I've updated this PR to not validate `assessment_date`.
Thanks

tijmenb (approved):

tijmenb:
Cool. Good to merge after content schemas and rummager is updated.

PR #977, opened 2016-12-21
Get the release branch back in sync with master

suzannehamilton:
The `release` and `deployed-to-*` branches aren't quite in sync with `master`. It looks like there were some commits to the `191` branch which were also pushed directly to `release`, e.g. 94f7a77f.

The old integration deployment process could cope with this - it must have been merging `master` into `release` on every deployment, though I can't spot where that happened. The new Jenkins 2 release process doesn't do this merge, so the Jenkins 2 `master` build is failing because `master` and `release` are on parallel branches.

Could we just merge `release` back into `master` to fix this?

Merging this into master will re-add a few lines to the README which are out of date (they were removed in bec2e122). I'll raise another PR to delete those again once this is merged.


suzannehamilton:
Here's the Christmas tree built by all the extra merges between release and master:

screen shot 2016-12-21 at 11 13 40

PR #981, opened 2017-01-05
Provide rake task to generate report for Content Operating Model team

h-lame:
The report is to show the organisation(s), url, finder, status and
first published date for all the documents in the system.

We expect that this is a one-off, but it might be useful for the code
that generates it to persist as this kind of report could be needed
again, or could be used as a jumping off point for future reports.

The trello card describing the report is: https://trello.com/c/e3nDUdE2/88-report-for-content-operating-model

### DIFF
--- a/lib/tasks/report.rake
+++ b/lib/tasks/report.rake
@@ -120,21 +120,48 @@ namespace :report do
       end
     end
 
+    def get_finder_document_hash(finder_schema)
+      begin
+        finder_document = Services.publishing_api.get_content(finder_schema.content_id).to_hash
+      rescue GdsApi::HTTPErrorResponse
+        finder_document = {}
+      end
+      {
+        "base_path" => finder_schema.base_path,
+        "first_published_at" => finder_document["first_published_at"] || "Never published to GOV.UK",
+        "publication_state" => finder_document["publication_state"] || "not-published",
+      }
+    end
+
     require 'csv'
 
-    output_filename = Rails.root.join("content-operating-report-#{Time.zone.today.strftime('%Y-%m-%d')}.csv")
+    document_report_filename = Rails.root.join("content-operating-report-for-documents-#{Time.zone.today.strftime('%Y-%m-%d')}.csv")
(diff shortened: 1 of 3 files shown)


h-lame:
Marked as WIP because we've some outstanding questions about what exactly is needed in the report.

h-lame:
Removed WIP tag as we've checked with the requestor that these reports provide them what they need.

lib/tasks/report.rake
```diff
@@ -120,21 +120,48 @@ namespace :report do
...
+    def get_finder_document_hash(finder_schema)
+      begin
+        finder_document = Services.publishing_api.get_content(finder_schema.content_id).to_hash
+      rescue GdsApi::HTTPErrorResponse
+        finder_document = {}
+      end
+      {
+        "base_path" => finder_schema.base_path,
+        "first_published_at" => finder_document["first_published_at"] || "Never published to GOV.UK",
+        "publication_state" => finder_document["publication_state"],
```
neilvanbeinum: Might it be useful to have a fallback value here too?

neilvanbeinum:
Overall looks good 👍

lib/tasks/report.rake
h-lame: Oh yeah, good point.  Will do.

lib/tasks/report.rake
```diff
@@ -120,21 +120,48 @@ namespace :report do
...
+        all_document_classes.each do |document_class|
+          document_count = 0
+          organisations_for_csv = organisations_for_document_class(document_class).join(', ')
+          each_document(document_class) do |document_hash|
+            document_count += 1
+            document_csv << [
+              public_url_for(document_hash),
+              organisations_for_csv,
+              document_class.title,
+              document_hash["publication_state"],
```
neilvanbeinum, line 154: We should probably have a fallback value here too

neilvanbeinum (changes requested):

lib/tasks/report.rake
h-lame: I don't think it's possible for a document in the publishing api to not have a state. Unlike first published at which can be blank for documents that haven't been published.

neilvanbeinum (approved):

PR #985, opened 2017-01-16
Test for email alert summaries over 255 characters

rubenarakelyan:
When a new email alert subscription is created, a short name is generated describing the subscription. If enough subscription items are chosen by the user, the short name becomes longer than 255 characters, which is the GovDelivery limit for this field.

This commit adds a test to catch cases where adding a new item takes the short name over 255 characters, so that it can be fixed.


Davidslv (approved):
🐝

PR #986, opened 2017-01-17
Add more text options for email alerts

rubenarakelyan:
This commit adds fields to most finders to define a separate email alert list title text which will be used when the normally generated string is too long (> 255 characters). It will be appended with a string of the form “x items”, where “x” is the number of selected topics and “items” is either the singular or plural string defined as `email_filter_name`.

Also reverts https://github.com/alphagov/specialist-publisher/pull/984 since this is a more comprehensive solution.

Deployment checklist:

- [x] Run the rake task `publishing_api:publish_finders` to re-publish all the finders

### DIFF
--- a/lib/documents/schemas/asylum_support_decisions.json
+++ b/lib/documents/schemas/asylum_support_decisions.json
@@ -18,9 +18,14 @@
   "signup_copy": "You'll get an email each time a decision is updated or a new decision is published.",
   "subscription_list_title_prefix": {
     "singular": "First-tier Tribunal (Asylum Support) decisions with the following category: ",
-    "plural": "First-tier Tribunal (Asylum Support) decisions with the following categories: "
+    "plural": "First-tier Tribunal (Asylum Support) decisions with the following categories: ",
+    "many": "First-tier Tribunal (Asylum Support) decisions: "
   },
   "email_filter_by": "tribunal_decision_categories",
+  "email_filter_name": {
+    "singular": "category",
+    "plural": "categories"
+  },
   "email_signup_choice": [
     {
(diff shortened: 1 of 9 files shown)


jennyd:
I'm not sure if this approach will work with GovDelivery, because it sounds like we would generate identical names for multiple topics. I've just tried to manually create a topic with the same name as an existing one (in our Integration account) and it errors, saying the name needs to be unique 😢  The [API docs](http://developer.govdelivery.com/api/comm_cloud_v1/Default.htm#API/Comm%20Cloud%20V1/API_CommCloudV1_Topics_CreateTopic.htm%3FTocPath%3DCommunications%2520Cloud%2520API%2520v1%7CResources%7CTopics%7C_____1) don't mention that constraint though.

We use the generated title as [both `name` and `short-name`](https://github.com/alphagov/email-alert-api/blob/599ea9b5b73bf57008284410448d79ea3a5b756a/app/services/gov_delivery/request_builder.rb#L7-L8) - it looks like only `name` needs to be unique, whereas the [length error was on `short-name`](https://errbit.publishing.service.gov.uk/apps/54452e0a0da115d037001662/problems/587a7d3d6578633cb4ad2a00), so there might be a way around those constraints. It would be simpler for our future plans to keep those two fields identical if possible, though. I can help you test on integration this afternoon if you like.

rubenarakelyan:
Discussed with @jennyd: following this PR I will raise another PR for `finder-frontend` that splits the logic for `title` and `short_name` to be separate so that we can continue to set `title` as we do now and use new logic for `short_name` that limits it to 255 characters. This PR is ready for review.

Davidslv (approved):
🌻 🍯

PR #988, opened 2017-01-24
Downgrade airbrake to 4.x

h-lame:
Our errbit installation doesn't like the airbrake 5.x gem and so deploys
don't get notified properly.  Downgrading to 4.x will fix this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,7 @@ source 'https://rubygems.org'
 
 gem 'rails', '5.0.0.1'
 
-gem 'airbrake', '~> 5.5'
-gem 'airbrake-ruby', '1.5'
+gem 'airbrake', '~> 4.0'
 gem 'logstasher', '0.6.2'
 gem 'unicorn', '~> 4.9.0'
 gem 'sass-rails', '~> 5.0.4'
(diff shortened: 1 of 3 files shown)


davidbasalla:
Does the config need to get updated/rolled back too? It changed in the [commit that bumped the gem](https://github.com/alphagov/specialist-publisher/commit/9b760f3cb899f7f2516f330dfa0d5a0e7a47799f). Just tried deploying the branch and got errors related to config

h-lame:
That was my plan today; to try and deploy it and see if it works.  I'll check out that commit, thanks @davidbasalla .

davidbasalla:
I was only being nosey because we'll need to do the same for LocalLinksManager :)

h-lame:
Hmm, deployed [this version](ef82f6ba1ee1de9234f3a844fb03343109588950) to integration and got [a 504 gateway timeout](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/9655/console) - is that better than a 404 not found ?

h-lame:
Oh! But it has recorded the deploy in errbit (twice...).  Gonna call this a success.

PR #990, opened 2017-01-26
Shorten CMA email subscription list name

rubenarakelyan:
The current name results in strings that are longer then 255 characters, which breaks a GovDelivery limit.

Deployment checklist:

- [x] Run the rake task `publishing_api:publish_finders` to re-publish all the finders


jennyd:
Thanks! Please could you also revert the various other merged PRs for the other approach we decided to abandon?

PR #991, opened 2017-01-26
Test length of subscription list names

rubenarakelyan:
This commit amend the test for the length of subscription list names to take into account the change of approach in how the names are generated. It now checks to make sure either the prefix (where only the prefix is used) or the plural prefix with all topics appended is a maximum of 255 characters long. Any finders that are not live are excluded from this test to allow for pre-production testing.

### DIFF
--- a/spec/finders/email_alert_configuration_spec.rb
+++ b/spec/finders/email_alert_configuration_spec.rb
@@ -25,21 +25,26 @@ RSpec.describe "Email alert configuration" do
         end
       end
 
-      # This is a GovDelivery maximum limit
-      # finder-frontend detects generated strings that are too long and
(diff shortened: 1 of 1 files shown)


tijmenb:
This fails for CMA cases, what can we do about that?

https://ci.integration.publishing.service.gov.uk/job/specialist-publisher/job/test-length-of-subscription-list-names/2/console

rubenarakelyan:
Yup that's expected. Once https://github.com/alphagov/specialist-publisher/pull/990 is merged then this will go green.

tijmenb:
Reviewing in the wrong order :) That's merged now, can you rebase?

rubenarakelyan:
Ready

PR #994, opened 2017-01-31
Make hidden_indexable_content indexable

lgarvey:
All MOJ finders use a hidden_indexable_field which should be indexed by
Rummger but currently isn't.  This PR adds code that includes the field
in a document's indexable content, when the field is present.

NOTE: This functionality used to work so likely it is a regression
introduced when specialist publisher was rebuilt.

### DIFF
--- a/app/presenters/search_presenter.rb
+++ b/app/presenters/search_presenter.rb
@@ -22,7 +22,8 @@ class SearchPresenter
   end
 
   def indexable_content
-    Govspeak::Document.new(document.body).to_text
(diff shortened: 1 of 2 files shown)


lgarvey:
@tijmenb As per our discussion I've created an initial PR to look at getting hidden content indexed.

sihugh:
The build's failing because of the linter.  If you can run `bundle exec govuk-lint-ruby -a`, and push up the changes, then it should get a bit further @lgarvey.  (You only need to worry about fixing the warnings in the files in your PR)

lgarvey:
Thanks Simon!

PR #996, opened 2017-02-02
Add content store document type attribute for search

Davidslv:
Add `content_store_document_type` to be sent to rummager.

Also, currently we have in place a task to republish documents to Publishing API and
Rummager, but in some cases we only want to republish to rummager in
order to update the document.

This commit adds an extra task into the republish.rake file to republish
all documents to rummager.

And also adds `content_id`, `rendering_app` and `publishing_app`

This might be best to be reviewed commit by commit (will rebase afterwards)

Trello:
https://trello.com/c/iMotNdUv/461-add-more-document-type-to-search

### DIFF
--- a/lib/rummager_republisher.rb
+++ b/lib/rummager_republisher.rb
@@ -1,0 +1,58 @@
+class RummagerRepublisher
+  class << self
+    def republish_all
+      document_types.each do |document_type|
+        pagination = publishing_api_pagination_for(document_type)
+
+        if pagination[:total_results].zero?
+          puts "skipping - could not find #{document_type}"
+          next
+        end
+
+        puts "Schedulling rummager republishing for document_type: #{document_type}"
+
+        current_page = pagination[:current_page]
+        per_page = pagination[:request_per_page]
+
+        while current_page <= pagination[:total_pages]
+          puts "Page: #{current_page} of #{pagination[:total_pages]}"
+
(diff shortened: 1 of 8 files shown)


Davidslv:
This needs more testing but I wanted to know your opinion on the implementation first.

lib/rummager_republisher.rb
```diff
@@ -0,0 +1,48 @@
+class RummagerRepublisher
+  def republish_all
+    document_types.each do |document_type|
+      pagination = publishing_api_pagination_for(document_type)
+
+      puts "skipping - could not find #{document_type}" if pagination[:total_results].zero?
+      next if pagination[:total_results].zero?
+
+      while pagination[:current_page] <= pagination[:total_pages]
```
tijmenb: I don't get what's going on here. 

`pagination[:current_page]` is not being mutated, and the inside of this block is  always the same?

tijmenb (changes requested):
As discussed on Slack, there's a sizeable chunk of copy-pasted code that isn't necessary for what we're trying to achieve.

Davidslv:
@tijmenb I believe I've addressed your comments. Let me know your thoughts.

Davidslv:
We think this work might have caused problems after running on Integration and we have decided to run the original republish.rake task (https://trello.com/c/zw0qWTz9/474-investigate-why-integration-is-missing-100k-search-results)

PR #997, opened 2017-02-06
Make rummager finder task consistent with Publishing API task

rubenarakelyan:
This commit makes the publishing-api finder loader more generic so that it can also be used by the rummager finder publishing task. It also makes the rummager finder publishing task more consistent with its publishing-api equivalent, and changes it to send to rummager immediately instead of placing changes in a queue.

This is what the publishing-api task does, and since these tasks are called manually, it makes sense for their results to be immediately available.

### DIFF
--- a/lib/rummager_finder_publisher.rb
+++ b/lib/rummager_finder_publisher.rb
@@ -31,6 +31,9 @@ private
 
   def export_finder(schema)
     presenter = FinderRummagerPresenter.new(schema[:file], schema[:timestamp])
-    RummagerWorker.perform_async(presenter.type, presenter.id, presenter.to_json)
+
+    logger.info("Publishing '#{schema[:file]['name']}' finder")
+
+    Services.rummager.add_document(presenter.type, presenter.id, presenter.to_json)
   end
 end
--- a/lib/finder_loader.rb
+++ b/lib/finder_loader.rb
@@ -1,6 +1,6 @@
 require "multi_json"
 
-class PublishingApiFinderLoader
+class FinderLoader
   def finders
     files.map do |file|
       {
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -1,10 +1,10 @@
 namespace :publishing_api do
   desc "Publish all Finders to the Publishing API"
   task publish_finders: :environment do
+    require "finder_loader"
     require "publishing_api_finder_publisher"
-    require "publishing_api_finder_loader"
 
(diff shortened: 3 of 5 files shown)


lib/rummager_finder_publisher.rb
```diff
@@ -31,6 +31,9 @@ def pre_production?(schema)
 
   def export_finder(schema)
     presenter = FinderRummagerPresenter.new(schema[:file], schema[:timestamp])
-    RummagerWorker.perform_async(presenter.type, presenter.id, presenter.to_json)
+
+    logger.info("Publishing '#{schema[:file]['name']}' finder")
+
+    Services.rummager.add_document(presenter.type, presenter.id, presenter.to_json)
```
Davidslv, line 37: That doesn't look to be the same, `rummager.add_document` is not an asynchronous operation https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/rummager.rb#L34 (unless I'm missing something)

Davidslv (changes requested):
I'm not sure about replacing RummagerWorker with `Service.rummager.add_document`

MatMoore (approved):
LGTM. Publishing them synchronously should be fine, there's only a small number of them.

lib/rummager_finder_publisher.rb
MatMoore: I think it's an intentional change. This is called manually from a rake task so doesn't need to be asynchronous.

rubenarakelyan: Yup you're right, this is a synchronous call rather than using the asynchronous worker. However, the reason I'm doing this is that:

1. This is only called manually as a rake task, most likely during a deployment, so it won't affect any publishing workflow
2. There are only a small number of finders and this call is pretty quick
3. It makes this publisher consistent with the publishing-api version, which also uses asynchronous calls
4. It means that changes are pretty much immediate during deployment

Davidslv: I wouldn't say that because you trigger this manually it shouldn't be an asynchronous operation. If there was lots of records then it would be beneficial to have Sidekiq performing the job.

Davidslv (approved):
Approving to keep consistency, maybe one day we will have to make it asynchronous again :P

lib/rummager_finder_publisher.rb
MatMoore: Yep definitely - most of the tasks in whitehall work that way because there's a massive amount of content to send, but this one is very small.

PR #999, opened 2017-02-07
Remove email topics for new MOJ finders

sihugh:
We've become aware of a limitation with signing up for multiple topics where
the total length of the selected [topic names + the `subscription_list_title_prefix`][1]
must [be less than 256 characters][2].  The number of topics and the length of
the names for these document types mean that users are likely to breach this
total if they select a few topics for email notification. To prevent against
this in the short term, we're removing the ability to select sub-topics in the
finders that will be going live in the near future.

We have work in the pipeline to re-engineer how our email notifications system
works - with this limitation being one of the things that will be tackled.  In
the short term, it's better to have notifications working rather than not.

[Trello card][3]

[1]: https://github.com/alphagov/finder-frontend/blob/92c9b30a51ab76b3fb51c067079a01ca090f9872/lib/email_alert_signup_api.rb#L37

[2]: https://errbit.publishing.service.gov.uk/apps/54452e0a0da115d037001662/problems/5889882c65786330adcd3100

[3]: https://trello.com/c/FautPYqf/629-email-subscriptions-issue-for-moj-finders


rubenarakelyan:
Yup, this is a pain but at least we can get some non-filterable alerts rather than none at all!

PR #1000, opened 2017-02-07
Enable ET and AST finders in production

sihugh:
We want to enable the Employment Tribunal Decisions and Asylum Support
Decisions finders in production.

This should not be merged until 8th Feb 2017 as this is go live date.

[Trello card](https://trello.com/c/yIqEcIQi/637-publish-et-decisions-and-asylum-support-decisions-finders)


sihugh:
This should be OK once https://github.com/alphagov/specialist-publisher/pull/999 is merged

davidbasalla:
🚀

PR #1005, opened 2017-02-17
Move presenters to correct directory

tijmenb:
The "rummager:index" rake task is currently failing because Rails can't find `FinderContentItemPresenter`. This is because the class is defined in `app/presenters/finders`, but is not in a `Finders` namespace.

By moving them to `app/presenters` autoloading just works. Also means that we can remove now-redundant requires.

@rubenarakelyan


rubenarakelyan:
Ah ha, this is why the Sidekiq rummager job wasn't working!

rubenarakelyan (approved):

PR #1007, opened 2017-02-22
Add PDF reporting task

neilvanbeinum:
[Trello Card](https://trello.com/c/JIT8o1Tn/90-pdf-report-for-natalie-baron-com)

This task produces a CSV of counts of published PDFs over time, grouped
by the owning organisation. This is for the Content Operating model team.

Also see (https://github.com/alphagov/manuals-publisher/pull/814) and (https://github.com/alphagov/whitehall/pull/3036).

There was a timeout problem on the integration environment when accessing a certain document category from publishing-api (on which this script relies). This was addressed in PR https://github.com/alphagov/publishing-api/pull/818.

### DIFF
--- a/lib/tasks/report.rake
+++ b/lib/tasks/report.rake
@@ -72,45 +72,9 @@ namespace :report do
       URI.join(Plek.new.website_root, document['base_path']).to_s
     end
 
-    class Paginator
-      def initialize(document_class)
-        @document_class = document_class
-      end
-
-      def document_type
-        @document_type ||= @document_class.document_type
-      end
-
-      def params(page)
-        {
-          publishing_app: "specialist-publisher",
-          document_type: document_type,
-          fields: [
-            :base_path,
-            :content_id,
-            :publication_state,
-            :first_published_at,
-          ],
-          page: page,
-          per_page: 100,
-          order: "-last_edited_at",
-        }
-      end
-
-      def each(&block)
-        page = 1
-        loop do
-          response = Services.publishing_api.get_content_items(params(page))
-          break if response['results'].empty?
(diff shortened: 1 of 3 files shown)


lib/tasks/report.rake
```diff
@@ -60,6 +60,117 @@ namespace :report do
...
+
+    def all_document_ids_and_version_hashes_for_document_class(document_class)
+      params = {
+        publishing_app: "specialist-publisher",
+        document_type: document_class.document_type,
+        fields: [
+          :content_id,
+          :state_history,
+        ],
+        per_page: 100000,
```
h-lame: I don't know if it'll help the timeout issues you were seeing, but I reckon we could reuse the Paginator in the `content_operating_model` report below to avoid doing this (and reduce load on publishing-api trying to return all 30,000 DFID documents in one go).

h-lame:
Other than the paginator, this looks good so far.

lib/tasks/report.rake
neilvanbeinum: Good shout, thanks! I've extracted it to a class https://github.com/alphagov/specialist-publisher/pull/1007/commits/20df4318d88361e0faf5c92fb2030b5d5d6d7a56

neilvanbeinum:
Thanks for the chat @h-lame. I've extracted a shared paginator class for use in reporting here https://github.com/alphagov/specialist-publisher/pull/1007/commits/59fd76191466c95f81932802818c06b8c9fdfca5#diff-b699b07caf1636ff1aaa3f702fc286de

h-lame (approved):

PR #1006, opened 2017-02-21
Add business finance support scheme

rubenarakelyan:
This commit adds support to specialist-publisher for business finance support schemes, which will be imported from the current business-support-finder, and the related finder.

Trello: https://trello.com/c/6q9hCkPK/492-add-business-finance-support-schemes-as-a-document-type-in-specialist-publisher

Deployment checklist:

- [x] Deploy specialist-publisher
- [x] Run the rake tasks `publishing_api:publish_finders` and `rummager:publish_finders` to publish the new business finance support schemes finder (pre-production)

screencapture-specialist-publisher-dev-gov-uk-business-finance-support-schemes-new-1487849074057

screencapture-dev-gov-uk-business-finance-support-1487846327515

PR #1008, opened 2017-02-22
Add rake task to import business finance support schemes

rubenarakelyan:
This commit adds a rake task to import all existing business finance support schemes into specialist-publisher as draft documents, using business-support-api. It also includes a task to delete all such imported documents for testing purposes.

This rake task will be deleted as part of a clean-up once it has been run and verified.

Merge after https://github.com/alphagov/specialist-publisher/pull/1006.

Trello: https://trello.com/c/ZZOQjhMV/487-import-business-support-schemes-data-from-business-support-api-into-specialist-publisher-as-draft

Deployment checklist:

- [x] Run the rake task `business_support_schemes:import_all` to import all business finance support schemes into specialist-publisher

### DIFF
--- a/lib/tasks/business_support_schemes.rake
+++ b/lib/tasks/business_support_schemes.rake
@@ -1,0 +1,185 @@
+# This is a one-time task that can be deleted after it has been run
+require 'gds_api/business_support_api'
+
+include ActionView::Helpers::SanitizeHelper
+
+# Provide our own save method so that we can override the base_path to be the
+# same as the imported one (this prevents some base_path clashes)
+def save_document(document, base_path)
+  return false if !document.valid?
+
+  document.update_type = 'major'
+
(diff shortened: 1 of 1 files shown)


lib/tasks/business_support_schemes.rake
```diff
@@ -0,0 +1,187 @@
...
+require 'gds_api/business_support_api'
+
+include ActionView::Helpers::SanitizeHelper
+
+# Provide our own save method so that we can override the base_path to be the
+# same as the imported one (this prevents some base_path clashes)
+def save_document(document, base_path)
+  return false if !document.valid?
+
+  begin
```
Davidslv: I think you can take this out, your beginning will be the start of the method

rubenarakelyan: Fixed

PR #1012, opened 2017-02-27
Add finder links to specialist documents

thomasleese:
Add edition links to the finder of a specialist document when saving.

Although this works, it shouldn't be merged until govuk-content-schemas#539 is merged. The reason why it currently passes schema validation checks is that [the schema for edition links is very lax](https://github.com/alphagov/govuk-content-schemas/blob/master/dist/formats/specialist_document/publisher_v2/schema.json#L125).

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -23,7 +23,8 @@ class Document
     :previous_version,
     :temporary_update_type,
     :update_type,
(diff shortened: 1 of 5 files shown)


carlosmartinez (approved):
giphy 1

PR #1013, opened 2017-02-27
Add “select all”/“clear all” buttons to dropdown menus

rubenarakelyan:
This commit adds “select all”/“clear all” buttons to dropdown menus. These buttons allow a user to quickly select or clear all options in a dropdown menu with one action, which is useful if there are many options that are often all selected.

The buttons require JavaScript and are also added using JavaScript, so they won’t appear if the browser has JavaScript disabled.

This change was requested by content designers working on the migration of the business finance support schemes, but can be useful elsewhere due to its generic nature.

Trello: https://trello.com/c/3Dnak3do/515-add-select-all-clear-all-buttons-for-select-fields-in-specialist-publisher

out

### DIFF
--- a/app/assets/javascript/application.js
+++ b/app/assets/javascript/application.js
@@ -1,5 +1,6 @@
 //= require jquery
 //= require select2
+//= require dropdown_select_all
 //= require length_counter
 //= require form_change_protection
 
@@ -7,6 +8,11 @@ jQuery(function($) {
   $(".select2").select2({
     placeholder: $(this).data('placeholder')
   });
+
+  ////
+  // Add "select all"/"clear all" buttons to each select2 dropdown menu
+  var dropdownSelectAll = new GOVUKAdmin.Modules.DropdownSelectAll();
+  dropdownSelectAll.start($("select.select2"));
 
   ////
   // Make a select2 that will create new values on return as you type them
--- a/app/assets/javascript/dropdown_select_all.js
+++ b/app/assets/javascript/dropdown_select_all.js
@@ -1,0 +1,32 @@
+(function(Modules) {
+  "use strict";
+  Modules.DropdownSelectAll = function() {
+    var that = this;
+    that.start = function(elements) {
+      elements.each(function() {
+        var selectId = $(this).attr("id");
+        $(this).parent()
+          .append('<a href="#" class="dropdown-select-all" data-select-id="#' + selectId  + '">Select all</a>')
+          .append('<a href="#" class="dropdown-clear-all" data-select-id="#' + selectId + '">Clear all</a>');
+      });
+
+      $(".dropdown-select-all").click(function() {
+        var selectId = $(this).data("select-id");
+        $(selectId + " option:not(:selected)")
+          .prop("selected", true)
+          .parent()
+          .trigger("change");
+        return false;
+      });
+
+      $(".dropdown-clear-all").click(function() {
+        var selectId = $(this).data("select-id");
+        $(selectId + " option:selected")
+          .prop("selected", false)
+          .parent()
+          .trigger("change");
+        return false;
+      });
+    };
+  };
+})(window.GOVUKAdmin.Modules);
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -18,6 +18,12 @@ form {
     width: 100%;
   }
 
+  .dropdown-select-all,
+  .dropdown-clear-all {
+    margin-top: 10px;
+    margin-right: 10px;
+  }
+
   .short-textarea {
     height: 50px;
   }


app/assets/javascript/application.js
```diff
@@ -9,6 +9,33 @@ jQuery(function($) {
   });
 
   ////
+  // Add "select all"/"clear all" buttons to each select2 dropdown menu
+  $("select.select2").each(function() {
+    var select_id = $(this).attr("id");
+    $(this).parent()
+      .append('<a href="#" class="btn btn-default select2-select-all" data-select-id="#' + select_id  + '">Select all</a>')
```
tijmenb: Not sure, but would links instead of buttons be better?

app/assets/javascript/application.js
```diff
@@ -9,6 +9,33 @@ jQuery(function($) {
   });
 
   ////
+  // Add "select all"/"clear all" buttons to each select2 dropdown menu
+  $("select.select2").each(function() {
```
tijmenb: Would be good to pull this out into a `GOVUK` module. Perhaps we can even pass put it into `govuk_admin_template`.

app/assets/javascript/application.js
```diff
@@ -9,6 +9,33 @@ jQuery(function($) {
...
+    var select_id = $(this).attr("id");
+    $(this).parent()
+      .append('<a href="#" class="btn btn-default select2-select-all" data-select-id="#' + select_id  + '">Select all</a>')
+      .append('<a href="#" class="btn btn-default select2-clear-all" data-select-id="#' + select_id + '">Clear all</a>');
+  })
+
+  $(".select2-select-all").click(function() {
```
tijmenb: Can we use a different class, or is this actually coupled to `select2`?

app/assets/javascript/application.js
rubenarakelyan: OK I've pulled it out into a separate module.

app/assets/javascript/application.js
rubenarakelyan: Yeah, I thought buttons would stand out more but I guess they could also be confused for buttons that will make you lose your place/unsaved edits. I have turned them into links which also has the good side-effect of saving space.

app/assets/javascript/application.js
rubenarakelyan: Not coupled, so I've changed it to be more generic.

app/assets/javascript/application.js
MatMoore: Agree it would be good to get it into admin templates. I think we prefer to wait until it's needed in more than one apps to move things into it though.

app/assets/javascript/application.js
MatMoore: Does the screenshot above include this change? Probably worth checking the design with Joe and seeing what he thinks.

app/assets/javascript/dropdown_select_all.js
```diff
@@ -0,0 +1,40 @@
+(function($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  function DropdownSelectAll(options) {
+    this.$dropdowns = $(options.$el);
+
+    if (this.$dropdowns.length > 0) {
+      this.$dropdowns.each(function() {
+        var select_id = $(this).attr("id");
```
MatMoore: should the identifiers be camelCase since this is JS?

MatMoore: if so, not sure why the linter doesn't catch it...

app/assets/javascript/dropdown_select_all.js
```diff
@@ -0,0 +1,40 @@
+(function($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
```
MatMoore: This is maybe a question for frontend devs, but is this module pattern consistent with the rest of the app?

The pattern I've seen in other publishing apps is this one: https://github.com/alphagov/govuk_admin_template/blob/master/JAVASCRIPT.md

app/assets/javascript/dropdown_select_all.js
```diff
@@ -0,0 +1,40 @@
...
+
+  function DropdownSelectAll(options) {
+    this.$dropdowns = $(options.$el);
+
+    if (this.$dropdowns.length > 0) {
```
MatMoore: I think this `if` is redundant

app/assets/javascript/dropdown_select_all.js
```diff
@@ -0,0 +1,40 @@
+(function($) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  function DropdownSelectAll(options) {
+    this.$dropdowns = $(options.$el);
```
MatMoore: Minor nitpick/style thing so feel free to ignore, but all these `this`s hurt my brain. Might help to assign the outer `this` to a variable so it's harder to confuse it with the ones inside the callbacks.

app/assets/javascript/application.js
```diff
@@ -9,6 +10,12 @@ jQuery(function($) {
   });
 
   ////
+  // Add "select all"/"clear all" buttons to each select2 dropdown menu
+  GOVUK.DropdownSelectAll({
+    $el: $("select.select2")
```
MatMoore: Is it safe to apply this to every select2 in the app? Could add another class for this instead.

app/assets/javascript/dropdown_select_all.js
rubenarakelyan: I've changed it to camel case

app/assets/javascript/dropdown_select_all.js
rubenarakelyan: Yup, removed

app/assets/javascript/dropdown_select_all.js
rubenarakelyan: I've assigned the outside `this` to `that` to match what happens in `govuk_admin_template`.

app/assets/javascript/application.js
rubenarakelyan: Having a chat with @tijmenb we decided to start by applying it to every dropdown by default, and only adding in a toggle if there are specific ones that we don't want to show this on.

app/assets/javascript/application.js
rubenarakelyan: It does now (I didn't upload the new screenshot until this morning). I'll check with Alex since Joe is out of the office.

app/assets/javascript/application.js
rubenarakelyan: Sounds good

app/assets/javascript/dropdown_select_all.js
rubenarakelyan: Ah I wasn't aware of this. I've changed the module to conform to this standard, which should also make it easier if we want to move this to the admin template later.

MatMoore (approved):
Looks good to me.

Regarding buttons vs links - maybe we could right align them so they're more out of the way? I expect most people probably won't need to use them (?)

PR #1020, opened 2017-03-23
Change label for "will continue on" metadata field

rubenarakelyan:
This commit changes the label for the "will continue on" metadata field to "will continue" to prevent users from thinking that "on" will be prepended to the text they enter in the field.

Requested by Constance from the Content Team.


Davidslv (approved):
👍

PR #1024, opened 2017-03-31
Use the new buildProject options syntax in the Jenkinsfile

suzannehamilton:
Replace the deprecated un-named parameter syntax with the new Map of options.


suzannehamilton:
This uses the new `buildProject` syntax in Jenkinslib: alphagov/govuk-puppet#5676

dwhenry (approved):

PR #1035, opened 2017-05-22
Make date field in DocumentSearch optional

dwhenry:
Due to a problem with the republish task documents are being returned
without a public_updated_at value. Instead of creating am error during
rummager republishing, these will now be republished to ES with the
field being nil.


MatMoore (approved):

MatMoore:
Looks good - we will look into why this is allowed to be nil at all separately.

PR #1038, opened 2017-05-26
Add a cautionary note to the republish tasks

kevindew:
These can be quite dangerous in a production environment, even now we
have fixed bugs in the Publishing API surrounding them.

### DIFF
--- a/lib/tasks/republish.rake
+++ b/lib/tasks/republish.rake
@@ -4,6 +4,14 @@ require 'csv'
 require 'services'
 
 namespace :republish do
+  # Caution: These tasks should be avoided in production environnments (with
+  # the exception of republish to rummager) as they change the date information
+  # of draft documents and extra superfluous history to published documents.
+  #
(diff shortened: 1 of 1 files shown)


thomasleese (approved):
Message looks good. I wonder if we can have some sort of extra argument to make sure they really want to run the task, something like this?

```bash
$ rake republish:all --yes
```

kevindew:
@thomasleese that makes sense but I'm not sure there is the need for it. I'm thinking currently anyone who wants to run this has to go to the source code to check the syntax so hopefully they see it. 

I'm going to go out on a limb and say we add that if somehow this gets run again and the message ignored. What do you think?

thomasleese:
True, I suspect it probably will be ignored in that case.

My thinking was that people might use `rake -T` to get a list of tasks, and then the comment wouldn't show up, but I think this is probably fine.

kevindew:
Yeah I'm expecting that since the command is run through Jenkins people won't tend to use the `rake -T` approach. We'll find out though I guess heh.

PR #1045, opened 2017-06-26
Add rake task to find pages with missing search content

suzannehamilton:
Identify pages in finders which have missing search content. Specialist documents have a hidden field named `hidden_indexable_content` which is used just for search indexing, but this data is missing from Rummager for many documents. This rake task finds those documents so that they can be republished.

This rake task should be deleted once the missing data has been added.

There is an existing rake task which republishes content by document type which should let us fix all the unindexed content.

https://trello.com/c/ikx387pT/140-republish-specialist-documents-in-administrative-appeals-tribunal-decisions

### DIFF
--- a/lib/tasks/indexable_content.rake
+++ b/lib/tasks/indexable_content.rake
@@ -1,0 +1,96 @@
+require 'gds_api/content_store'
+
+# TODO: Delete this rake task once the missing field has been populated
+
+# Identify pages in finders which have missing search content.
+#
+# Specialist documents have a hidden field which is used just for search
+# indexing, but this data is missing from Rummager for many documents. This rake
+# task finds those documents so that they can be republished.
+task find_unindexed_content: :environment do
+  finders = [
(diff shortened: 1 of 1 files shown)


suzannehamilton:
When I ran this in dev it found 854 documents, and took a few minutes to run.

lib/indexable_content_checker.rb
```diff
@@ -0,0 +1,71 @@
+require 'gds_api/content_store'
```
dwhenry: As discussed, can we move this class inside the rake task so that it is clearer they will both be deleted once we have the data. :)

suzannehamilton: Done. Also renamed the `y` variable to `yielder`, which is the terminology that `Enumerator` uses.

dwhenry (approved):
Looks good, let make sure we have a task to delete this once run. :)

suzannehamilton:
I've added it to the Trello card checklist. 👍

PR #1047, opened 2017-07-11
Update MHRA’s email address

tuzz:
https://trello.com/c/FdvTwmGr/901-change-email-address-in-mrha-drug-alert-2016-email-footer-template

These emails are sent via a call to the `email-alert-api` from a Sidekiq worker in this app. Payloads are pre-computed when jobs are enqueued so existing messages will be sent with the old address. After the app is restarted after deploy, emails will be sent with the new address.


davidbasalla (approved):
Looks good!

PR #1048, opened 2017-07-25
Send update_type with put content

emmabeynon:
We are deprecating sending the update_type in publish requests made to the
Publishing API, and now expect clients to send update_type in PUT requests
instead. We also upgrade gds-api-adapters to bring this app up to date with
the new expected behaviour

https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk


kevindew (approved):
Looks good 👍

PR #1053, opened 2017-09-07
Business finance support scheme content report

h-lame:
For: https://trello.com/c/pk7rYWJ7/210-in-built-report-for-specialist-publisher

When business finance support schemes were moved from publisher to specialist-publisher we lost the ability to export all the docs to CSV which is something the folk who manage these docs like to use.  This PR re-creates that functionality in specialist publisher.  The commit that removed the report from publisher is https://github.com/alphagov/publisher/pull/607/commits/e9ef795c4afab16f7d53e84289d18c6b3884efe1 and we used that to work out what should be in the reports.  It's not an exact copy because migrating from publisher to specialist-publisher also changed the shape of the documents by dropping many fields in preference for using the body.

We follow the pattern for reports that whitehall uses, which is that the user clicks "export to CSV" in the UI on a document index and we then email them that list.  This allows them to get a subset by doing a search first.

The email process relies on https://github.com/alphagov/govuk-app-deployment/pull/174 which adds the SES config to the app.  We should only merge this after that has been merged.


h-lame:
Marked as DNM because we need to make the deployment change first.

h-lame:
Removed DNM as https://github.com/alphagov/govuk-app-deployment/pull/174 is merged.  I've deployed this branch for testing on integration and it sends out email, so the config is good.

DennisTsiang (approved):
Looks good to me

PR #1058, opened 2017-10-17
Add task to publishing special routes for Finder Frontend site search

brenetic:
We are adding site search to Finder Frontend. As these are published using Specialist Publisher to manage the special route as it already talks to the Publishing API.

https://trello.com/c/lxKbO5Xw


brenetic:
Taking a different approach

PR #1060, opened 2017-10-27
Present all necessary fields to email-alert-api

rubenarakelyan:
These new fields are required by the API for GOV.UK Notify integration.

Trello: https://trello.com/c/02JCCXB6/185-tap-and-specialist-publisher-should-match-service-regarding-what-they-send-to-the-api

### DIFF
--- a/app/presenters/email_alert_presenter.rb
+++ b/app/presenters/email_alert_presenter.rb
@@ -7,12 +7,25 @@ class EmailAlertPresenter
 
   def to_json
     {
+      title: title,
+      description: summary,
+      change_note: change_note,
       subject: subject,
       body: body,
       tags: tags,
       urgent: urgent,
       document_type: document.document_type,
+      email_document_supertype: "other",
+      government_document_supertype: "other",
+      content_id: content_id,
+      public_updated_at: public_updated_at,
+      publishing_app: "specialist-publisher",
+      base_path: base_path
     }.merge(extra_options)
+  end
+
+  def content_id
+    document.content_id
   end
 
 private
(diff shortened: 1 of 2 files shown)


spec/presenters/email_alert_presenter_spec.rb
```diff
@@ -31,8 +33,16 @@
...
+        expect(presented_data[:public_updated_at]).to include(cma_case_payload["public_updated_at"])
+        expect(presented_data[:base_path]).to include(cma_case_payload["base_path"])
         expect(presented_data[:footer]).to include("SUBSCRIBER_PREFERENCES_URL")
         expect(presented_data[:header]).to include("govuk-email-header")
+
+        expect(presented_data.keys.map(&:to_s)).to match_array(%w(
+          title description change_note subject body tags document_type
+          email_document_supertype government_document_supertype content_id
+          public_updated_at publishing_app base_path urgent header footer
+        ))
```
kevindew: might be a little cleaner to drop the map and use `%i` instead of `%w`

kevindew (approved):
Small comment - but not a blocker

spec/presenters/email_alert_presenter_spec.rb
rubenarakelyan: Good point, fixed!

app/presenters/email_alert_presenter.rb
```diff
@@ -89,14 +102,34 @@ def document_noun
...
+  def subject
+    title
+  end
+
   def urgent
     document.urgent
   end
 
+  def public_updated_at
+    document.public_updated_at
```
thomasleese, line 126: Apparently this can still come through as `null`... https://sentry.io/govuk/app-email-alert-api/issues/391875467/

PR #1061, opened 2017-11-08
Send the published document to email-alert-api

rubenarakelyan:
This commit changes the document publisher to send the published document (rather than the in-memory instance before it is published) to email-alert-api and other classes. This means that fields that are populated by publishing-api (such as `public_updated_at`) are available to use.

Trello: https://trello.com/c/vEVazSDL/302-fix-missing-email-alert-service-fields-in-specialist-publisher


tuzz (approved):

thomasleese (approved):
LGTM!

PR #1063, opened 2017-11-09
[DO NOT MERGE] Remove direct indexing of search data

dwhenry:
Search indexing is now done via the Publishing API.

https://trello.com/c/vqEInKpK/421-remove-rummager-api-from-specialist-publisher

This is pending final migration of specialist publisher formats to be sourced from publishing API

I have left: https://github.com/alphagov/specialist-publisher/blob/remove-direct-rummager-indexing/lib/rummager_finder_publisher.rb as we don't appear to have included this on the migration plan at this stage


brenetic (approved):
🚀

PR #1067, opened 2017-11-17
Add rake task to unpublish unsupported schemes

emmabeynon:
For [Trello card](https://trello.com/c/kEvC2Bpc/158-remove-ni-and-welsh-schemes-from-business-support-finder)

Northern Ireland and Wales entries within business support finder are not
supported by the business support finder team, so we need to remove them.  This
adds a temporary rake task to unpublish these schemes in bulk and return a 410 'Gone' page.  We also temporarily add a CSV list of the schemes that need to be removed.  

This has been run on integration and all 112 schemes were successfully unpublished.
Note, this PR will be reverted after the task has been run in production.


NickColley:
I'm not super savvy with the specifics of the publishers but the code is very readable and makes good sense to me 👍

leenagupte (approved):
I agree with @nickcolley's comments. The code is easy to follow.

👍  LGTM

PR #1104, opened 2017-12-12
Add conditionals so we can run E2E tests in rails production mode

adrianclay:
These allow us to connect to single instances of Mongo rather than clusters, compile the assets for testing purposes and seed the database while still in production mode without interfering with the existing configurations.

These also allow for easily switching to run the application in development.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,5 +1,5 @@
 def seed
-  unless Rails.env.development?
+  unless Rails.env.development? || ENV["RUN_SEEDS_IN_PRODUCTION"]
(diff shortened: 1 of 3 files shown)


db/seeds.rb
```diff
@@ -1,5 +1,5 @@
 def seed
-  unless Rails.env.development?
+  unless Rails.env.development? || ENV["RUN_SEEDS_IN_PRODUCTION"] == "true"
```
kevindew: might be better to just check this env var exists than that it equals a string?

kevindew (approved):

PR #1117, opened 2018-01-03
Fix deprecation warnings for gds-sso and factory_bot

boffbowsh:
Remove defunct call to `require_signin_permission!`, gds-sso always checks for signin permission now.
Replace `factory_girl` with `factory_bot`. Only factory_bot will receive updates from now on. See
https://robots.thoughtbot.com/factory_bot

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,9 +6,8 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
   include Pundit
 
-  before_action :require_signin_permission!
   before_action :set_authenticated_user_header
-  after_action :verify_authorized
+  before_action :authenticate_user!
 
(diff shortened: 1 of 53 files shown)


app/controllers/application_controller.rb
```diff
@@ -6,9 +6,8 @@ class ApplicationController < ActionController::Base
   include GDS::SSO::ControllerMethods
   include Pundit
 
-  before_action :require_signin_permission!
   before_action :set_authenticated_user_header
-  after_action :verify_authorized
```
cbaines, line 11: This seems unrelated to GDS::SSO, but instead to pundit. It sounds from the pundit docs that it's useful to have?

cbaines (approved):

boffbowsh:
Needs more work.

PR #1120, opened 2018-01-04
Downgrade select2-rails to fix broken form

suzannehamilton:
The recent upgrade of select2-rails from version 3 to 4 broke the Authors field: authors were appearing as a single string rather than in separate boxes with a delete button. JS errors also stopped Body preview from working.

We can't use select2's backwards-compatibility mode (called select2-full) because it's not completely compatible with the Authors field: it doesn't allow users to enter names which don't match the autocomplete values.

Instead, this just downgrades select2-rails back to the working version until we can follow the full upgrade path: https://select2.org/upgrading/migrating-from-35

This fixes https://govuk.zendesk.com/agent/tickets/2545435

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ gem 'logstasher', '1.2.2'
 gem 'mongoid', '~> 6.0.0'
 gem 'pundit'
 gem 'sass-rails', '~> 5.0.4'
-gem 'select2-rails', '~> 4.0.3'
+gem 'select2-rails', '~> 3.5.10'
 gem 'uglifier', '>= 1.3.0'
 gem 'unicorn', '~> 5.4.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -332,7 +332,7 @@ GEM
     scss_lint (0.56.0)
       rake (>= 0.9, < 13)
       sass (~> 3.5.3)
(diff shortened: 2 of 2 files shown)


thomasleese (approved):
We caught something similar in Mainstream Publisher - https://github.com/alphagov/publisher/issues/721

Davidslv (approved):

Gemfile
```diff
@@ -12,7 +12,7 @@ gem 'logstasher', '1.2.2'
 gem 'mongoid', '~> 6.0.0'
 gem 'pundit'
 gem 'sass-rails', '~> 5.0.4'
-gem 'select2-rails', '~> 4.0.3'
+gem 'select2-rails', '~> 3.5.10'
```
Davidslv, line 15: We should probably comment why we keeping this version

suzannehamilton: Good point. I've added a comment.

h-lame: Is there a magic comment we can add to the gemfile for telling dependabot not to bother us for this gem?

suzannehamilton: When you close a dependabot PR, it gives you instructions on how to tell it to ignore a minor or major version. I didn't do with this gem in case someone wanted to try the upgrade process, which might be overly optimistic. 🙂

PR #1146, opened 2018-01-30
Add rake task to publish a single finder

elenatanasoiu:
For: https://trello.com/c/bogvHiGv/53-add-a-rake-task-for-publishing-a-single-finder-to-specialist-publisher

To publish a single finder:

```
bundle exec rake publishing_api:publish_finder["residential_property_tribunal_decisions"]
```

The parameter is the name of the json schema in ```lib/documents/schemas/```

### DIFF
--- a/lib/finder_loader.rb
+++ b/lib/finder_loader.rb
@@ -2,11 +2,24 @@ require "multi_json"
 
 class FinderLoader
   def finders
-    files.map do |file|
+    files.map do |json_schema|
       {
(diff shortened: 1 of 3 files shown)


h-lame (changes requested):
We could add a spec for the new `FinderLoader#finder` as there are already ones for `finders`.  We can use it to explain / make explicit that even though you ask for a single finder, you're gonna get an array back.

h-lame (approved):

PR #1167, opened 2018-02-21
Revert using foreman in dev environment

kevindew:
Wider context: https://github.com/alphagov/publishing-e2e-tests/pull/202

This removes foreman from the Gemfile and from the startup.sh usage and
instead installs foreman separately via `gem install` in the Dockerfile
so that it can be used in the docker environment.

The reason for this is that running everything via unicorn in dev can
cause some wtfs (such as breaking better_errors) and installing foreman
via Gemfile does contradict some contenious advice given by the foreman
gem maintainer.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source 'https://rubygems.org'
 gem 'rails', '5.0.2'
 
 gem 'bootstrap-kaminari-views', '0.0.5'
-gem 'foreman'
 gem 'govuk_sidekiq', '~> 2.0'
 gem 'hashdiff'
 gem 'jquery-rails', '~> 4.3.1'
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -31,6 +30,7 @@ group :development, :test do
   gem 'capybara-webkit'
   gem 'database_cleaner'
   gem 'factory_bot'
+  gem 'foreman'
```
adrianclay: The commit talks about removing `foreman`, but in this case you're still keeping it.  Why is it still required here?

kevindew: Ah probably copy and pasting identical messages fail. I kept it in dev so it could still be run in https://github.com/alphagov/govuk-puppet/blob/master/development-vm/Procfile but I've since updated that so can now remove it. Thanks

thomasleese (approved):

PR #1209, opened 2018-04-19
Add new judge appointments

kevindew:
From Zendesk ticket: https://govuk.zendesk.com/agent/tickets/2746246


schmie (approved):
There is precedence for using two characters for first names (JG Mitchell) - Should it be Church, TH, then?
Otherwise all looks good.

kevindew:
@schmie good spot - I hadn't noticed that. I'll amend.

PR #1230, opened 2018-06-15
Add organisations fields

steventux:
https://trello.com/c/Hf90xZZq/219-drop-down-for-organisations-in-eu-exit-statutory-instrument-specialist-publisher-document-type

Adds the abilities to select Primary publishing organisation and organisations for Statutory Instruments.

screenshot from 2018-06-15 11-53-04

### DIFF
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -1,6 +1,6 @@
 class Organisation
   def self.all
-    @organisations ||= fetch_all_organisations(organisation_params)
+    @organisations = fetch_all_organisations(organisation_params)
   end
 
   attr_reader :content_id, :title
--- a/app/models/organisation.rb
+++ b/app/models/organisation.rb
@@ -1,0 +1,42 @@
+class Organisation
+  def self.all
+    @organisations ||= fetch_all_organisations(organisation_params)
+  end
+
+  attr_reader :content_id, :title
+
+  def initialize(attrs)
+    @content_id = attrs["content_id"]
+    @title = attrs["title"]
+  end
+
+  class << self
+    def document_type
+      "organisation"
+    end
+
+    def fetch_all_organisations(params)
+      orgs = []
+      current_page = 1
+      loop do
+        response = Services.publishing_api.get_content_items(params.merge(page: current_page))
+        orgs << response["results"].map { |attrs| Organisation.new(attrs) }
+        if response["pages"] > current_page
+          current_page += 1
+        else
+          break
+        end
+      end
+      orgs.flatten
+    end
+
+    def organisation_params
+      {
+        document_type: document_type,
+        fields: %i[content_id title],
+        order: "base_path",
+        per_page: "600", # Around ~1070 orgs.
+      }
+    end
+  end
+end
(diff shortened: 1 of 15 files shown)


app/models/organisation.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    def fetch_all_organisations(params)
+      orgs = []
+      current_page = 1
+      loop do
+        response = Services.publishing_api.get_content_items(params.merge(page: current_page))
+        orgs << response["results"].map { |attrs| Organisation.new(attrs) }
+        if response["pages"] > current_page
+          current_page += 1
+        else
```
elliotcm, line 26: Will this enter an infinite loop if there's some issue with returning the orgs?

app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -17,3 +17,29 @@
...
+<% end %>
+<%= render "shared/form_group", f: f, field: :organisations do %>
+  <%= f.select :organisations,
+      organisations_options,
+      {},
+      {
+        class: "select2 form-control",
+        multiple: true,
+        data: {
+          placeholder: "Select organisations"
```
elliotcm, line 41: I think the terms we wanted in the editor itself were "Publishing organisation" and "Other associated organisations".

app/models/statutory_instrument.rb
```diff
@@ -12,16 +13,26 @@ class StatutoryInstrument < Document
...
 
   def self.title
     "Statutory instrument"
   end
 
-  def primary_publishing_organisation
-    "fef4ac7c-024a-4943-9f19-e85a8369a1f3"
+  def links
+    super.merge(
+      organisations: organisations,
```
elliotcm: Since `organisations` should also contain the primary publishing org we should probably union them here so we can provide the right list of orgs regardless of whether the user put the primary org in the additional field or not.

elliotcm (changes requested):

app/models/organisation.rb
steventux: It's a good question, I'd borrowed heavily from https://github.com/alphagov/specialist-publisher/blob/master/app/services/all_documents_finder.rb#L35 
I could put in some 'fuse' condition to stop after a finite number of iterations.

elliotcm (approved):
The explicit 10 loops makes me twitch a little because it's a time bomb for, what, 10 years from now, but the chances of us hitting 6k orgs seems slim so..  approved.

steventux:
@elliotcm I've added a commit to include the primary publishing org with the organisations in the links hash.

app/models/statutory_instrument.rb
```diff
@@ -27,7 +27,7 @@ def self.title
 
   def links
     super.merge(
-      organisations: organisations,
+      organisations: organisations << primary_publishing_organisation,
```
elliotcm: This will result in duplicates if they put the org in both fields, probably want `organisations | [primary_publishing_organisation]`

elliotcm (changes requested):

steventux:
@elliotcm good spot - updated.

elliotcm (approved):

PR #1241, opened 2018-06-21
Tag Statutory Instruments to Brexit

1pretz1:
Every statutory_instrument document will now be tagged to the
Topic Taxonomy's Brexit taxon.

### DIFF
--- a/app/presenters/document_links_presenter.rb
+++ b/app/presenters/document_links_presenter.rb
@@ -1,16 +1,22 @@
 class DocumentLinksPresenter
+  BREXIT_CONTENT_ID = "d6c2de5d-ef90-45d1-82d4-5f2438369eea".freeze
+
   def initialize(document)
     @document = document
   end
 
   def to_json
-    {
+    json = {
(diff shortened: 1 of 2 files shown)


leenagupte (changes requested):
@1pretz1 you could make this change a lot simpler by limiting your changes to the `to_json` method in the `DocumentLinksPresenter` class.

If you add a conditional like:

```
if document.is_a? StatutoryInstrument
  links[:taxons] = [BREXIT_CONTENT_ID]
end
```

to the end of the `to_json` method, you won't need a separate `statutory_links` method and you won't need the extra conditional in `DocumentSaver`.

Or, could you add `taxons` to the links in the `StatutoryInstruments` model? i.e. here: https://github.com/alphagov/specialist-publisher/blob/master/app/models/statutory_instrument.rb#L30

1pretz1:
@leenagupte Agreed, had to put the hash in a variable however :)

leenagupte (approved):

PR #1242, opened 2018-06-22
Fix Schema validation test failure

1pretz1:
This test is throwing a Validation error, due to a `statutory_instrument`
document being created with a `subject` of 'oil-and-gas' which isn't a
permitted field in the govuk_content_schemas. Changing the `subject` to
`business` passes the validation as this is a permitted value.

Error;

```
The property '#/details/metadata/subject/0' value "oil-and-gas" did not
match one of the following values: business, crime-justice-and-law,
defence, education-training-and-skills, entering-and-staying-in-the-uk,
environment, going-and-being-abroad, government, health-and-social-care,
housing-local-and-community, international, life-circumstances, money,
parenting-childcare-and-childrens-services,
regional-and-local-government, society-and-culture, transport, welfare,
work
```

### DIFF
--- a/spec/fixtures/factories.rb
+++ b/spec/fixtures/factories.rb
@@ -433,7 +433,7 @@ FactoryBot.define do
           "laid_date" => "2018-01-01",
           "sift_end_date" => "2018-01-05",
           "sifting_status" => "open",
-          "subject" => ["oil-and-gas"],
+          "subject" => ["business"],
         }
       }
 


spec/fixtures/factories.rb
```diff
@@ -433,7 +433,7 @@
           "laid_date" => "2018-01-01",
           "sift_end_date" => "2018-01-05",
           "sifting_status" => "open",
-          "subject" => ["oil-and-gas"],
+          "subject" => ["business"],
```
leenagupte, line 436: @1pretz1 I don't understand why this test is failing. According to the schema definition, `oil-and-gas` is a valid value for subject: 

https://github.com/alphagov/govuk-content-schemas/blob/correct-links-schemas/formats/shared/definitions/_specialist_document.jsonnet#L1297

leenagupte: `business` isn't a valid value, so the test should still fail! https://github.com/alphagov/govuk-content-schemas/blob/correct-links-schemas/dist/formats/specialist_document/publisher_v2/schema.json#L1710

leenagupte (approved):
@1pretz1 It turns out I was looking that the wrong version of the code in the previous comments. Sorry about that!

🎉 💯 LGTM!

PR #1249, opened 2018-06-26
Automatically tag specialist content to the topic taxonomy.

koetsier:
- When specialist content is saved, it is automatically tagged to the topic taxonomy.
- Added a rake task to automatically tag all specialist content to the topic taxonomy.

`rake auto_tagging[tag]
` 
Tags all content if the argument 'tag' is supplied.

Trello: https://trello.com/c/iOt4YLW5/176-setup-specialist-publisher-to-automatically-tag-more-content-to-the-topic-taxonomy

### DIFF
--- a/app/models/countryside_stewardship_grant.rb
+++ b/app/models/countryside_stewardship_grant.rb
@@ -12,6 +12,10 @@ class CountrysideStewardshipGrant < Document
     super(params, FORMAT_SPECIFIC_FIELDS)
   end
 
+  def taxons
+    [COUNTRYSIDE_TAXON_ID]
+  end
+
   def self.title
     "Countryside Stewardship Grant"
   end
--- a/lib/documents_tagger.rb
+++ b/lib/documents_tagger.rb
@@ -1,0 +1,42 @@
+class DocumentsTagger
+  def self.tag_all(do_tag = true)
+    new(do_tag).tag_all
+  end
+
+  def initialize(do_tag = true)
+    @do_tag = do_tag
+  end
+
+  def tag_all
+    all_documents.map do |document|
+      mapped_taxon_ids = mapped_taxons(document['document_type'])
+      taxons_tagged = mapped_taxon_ids.empty? ? [] : tag_to_taxons(document['content_id'], mapped_taxon_ids)
+      { base_path: document['base_path'], content_id: document['content_id'], taxons: taxons_tagged }
+    end
+  end
+
+private
+
+  def tag_to_taxons(content_id, taxon_ids)
+    tagged = Tagger.add_tags(content_id, @do_tag) do |existing_taxon_ids|
+      existing_taxon_ids + taxon_ids
+    end
+    tagged ? taxon_ids : []
+  end
+
+  def all_document_types
+    FinderSchema.schema_names.map(&:singularize)
+  end
+
+  def mapped_taxons(document_type)
+    document_type.camelize.constantize.new.taxons
+  end
+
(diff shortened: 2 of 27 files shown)


app/models/countryside_stewardship_grant.rb
```diff
@@ -6,12 +6,18 @@ class CountrysideStewardshipGrant < Document
     funding_amount
   ).freeze
 
+  COUNTRYSIDE_TAXON_ID = '9129d716-365b-44ba-9856-383423fe1e41'.freeze
```
leenagupte: @koetsier This one should be moved to `Document` with its friends.

lib/tagger.rb
```diff
@@ -0,0 +1,22 @@
+class Tagger
+  def self.add_tags(content_id, taxon_ids, do_tag = true)
+    link_content = Services.publishing_api.get_links(content_id)
+    version = link_content['version']
+    existing_taxon_ids = link_content.dig('links', 'taxons') || []
+    return [] if  (taxon_ids - existing_taxon_ids).empty?
```
leenagupte: I would extract this line out to a function that explains what it's doing.

lib/tagger.rb
```diff
@@ -0,0 +1,22 @@
+class Tagger
+  def self.add_tags(content_id, taxon_ids, do_tag = true)
+    link_content = Services.publishing_api.get_links(content_id)
+    version = link_content['version']
+    existing_taxon_ids = link_content.dig('links', 'taxons') || []
```
leenagupte: You could extract these lines out to a method, e.g. `get_existing_tagged_taxons` or something to make this a bit more readable.

lib/tagger.rb
```diff
@@ -0,0 +1,22 @@
...
+    return [] if  (taxon_ids - existing_taxon_ids).empty?
+    new_taxon_ids = (existing_taxon_ids + taxon_ids).uniq
+    if do_tag
```
leenagupte: Perhaps extract this too?

lib/documents_tagger.rb
```diff
@@ -0,0 +1,35 @@
...
+    new.tag_all(do_tag)
+  end
+
+  def tag_all(do_tag)
+    all_documents.map do |document|
+      taxon_ids = taxons(document)
+      taxons_tagged = if taxon_ids.empty?
+                        []
+                      else
+                        Tagger.add_tags(document['content_id'], taxon_ids, do_tag) unless taxon_ids.empty?
```
leenagupte: I don't think you need this `unless` because you are already catching empty taxons in the `if`.

cbaines:
I haven't had time to review this, but one thought I had was what should happen if a few bits of content are also tagged to taxons other than the ones they would be automatically tagged to?

I'm not sure if this would ever happen, or if it does happen, but it would be good to think about.

app/models/countryside_stewardship_grant.rb
koetsier: fixed

lib/tagger.rb
koetsier: done

lib/tagger.rb
koetsier: done

lib/tagger.rb
koetsier: done - refactored this method

peteglondon (approved):
All looks good - small typo in 'add rake task to perform autmatic tagging'

PR #1259, opened 2018-07-03
Use  get_paged_editions instead of get_content_items_enum

koetsier:
get_content_items_enum times out because of SQL issues wrt
OFFSET. get_paged_editions does not time out because it does
not use OFFSET

### DIFF
--- a/lib/documents_tagger.rb
+++ b/lib/documents_tagger.rb
@@ -32,10 +32,22 @@ private
     document_type.camelize.constantize.new.taxons
   end
 
+  def editions_enum(params = {})
+    enum = Services.publishing_api.get_paged_editions(params)
+    Enumerator.new do |yielder|
+      enum.each do |response|
+        response
+          .to_h
+          .fetch('results', [])
+          .each { |result| yielder << result }
+      end
+    end
+  end
+
   def all_documents
-    Services.publishing_api.get_content_items_enum(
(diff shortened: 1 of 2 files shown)


lib/documents_tagger.rb
```diff
@@ -32,10 +32,22 @@ def mapped_taxons(document_type)
     document_type.camelize.constantize.new.taxons
   end
 
+  def get_paged_editions_enum(params = {})
+    enum = Services.publishing_api.get_paged_editions(params)
+    Enumerator.new do |yielder|
+      enum.each do |response|
```
cbaines: `enum` isn't very clear to me. I guess it is a page? If so, `pages` might read better as a variable name?

lib/documents_tagger.rb
```diff
@@ -32,10 +32,22 @@ def mapped_taxons(document_type)
...
+        results = response.to_h.fetch('results', [])
+        results.each do |result|
+          yielder << result
+        end
```
cbaines: I'd perhaps consider getting rid of the `results` variable. Chaining the methods directly together may be easier to read, e.g.

```ruby
        response
          .to_h
          .fetch('results', [])
          .each { |result| yielder << result }
```

cbaines (approved):
I've added a couple of comments, but this looks good to me :+1:

PR #1317, opened 2018-10-12
Add rake task to edit a document's base_bath

AlanGabbianelli:
Add rake task to edit a document's base_bath

Trello card:
https://trello.com/c/kTKWF48j/530-document-how-to-change-a-url-in-specialist-publisher

How to change a URL in Specialist Publisher wasn't a documented process.
Making it into a rake task seems a nicer solution than simply
documenting how to do it and then having to do it manually.

Now, in order to edit a document's base_path, it's possible to just use
the rake task passing the document's `content_id` and the new
`base_path` as arguments.

### DIFF
--- a/lib/tasks/base_path.rake
+++ b/lib/tasks/base_path.rake
@@ -1,0 +1,24 @@
+namespace :base_path do
+  desc "Edits the base_path of a document, given its content_id and the new base_path"
+  task :edit, [:content_id, :base_path] => [:environment] do |_t, args|
+    content_id = args[:content_id]
+    base_path = args[:base_path]
+
+    begin
+      document = Document.find(content_id)
+    rescue DocumentFinder::RecordNotFound => e
+      puts "Error finding the document: #{e.inspect}"
+    end
+
+    if document
+      document.base_path = base_path
+      document.update_type = 'minor'
+      if document.save
+        puts "The #{document.class} with title \"#{document.title}\" has been successfully edited"
+        puts "Its new base_path is now \"#{document.base_path}\""
+      else
+        puts "Couldn't edit the #{document.class} with title \"#{document.title}\""
+      end
+    end
+  end
+end
--- a/lib/tasks/base_path.rake
+++ b/lib/tasks/base_path.rake
@@ -1,24 +1,13 @@
 namespace :base_path do
   desc "Edits the base_path of a document, given its content_id and the new base_path"
-  task :edit, [:content_id, :base_path] => [:environment] do |_t, args|
-    content_id = args[:content_id]
-    base_path = args[:base_path]
+  task :edit, %i[content_id new_base_path] => %i[environment] do |_t, args|
+    document = Document.find(args[:content_id])
+    old_base_path = document.base_path
 
-    begin
-      document = Document.find(content_id)
-    rescue DocumentFinder::RecordNotFound => e
-      puts "Error finding the document: #{e.inspect}"
-    end
+    document.base_path = args[:new_base_path]
+    document.update_type = 'minor'
+    document.save!
 
-    if document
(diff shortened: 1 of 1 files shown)


lib/tasks/base_path.rake
```diff
@@ -0,0 +1,24 @@
+namespace :base_path do
+  desc "Edits the base_path of a document, given its content_id and the new base_path"
+  task :edit, [:content_id, :base_path] => [:environment] do |_t, args|
+    content_id = args[:content_id]
+    base_path = args[:base_path]
+
+    begin
+      document = Document.find(content_id)
+    rescue DocumentFinder::RecordNotFound => e
+      puts "Error finding the document: #{e.inspect}"
```
barrucadu: It looks like there are actually a few different exceptions which could be raised here:

```
irb(main):001:0> document = Document.find('')
...
GdsApi::TimedOutException (GdsApi::TimedOutException)

irb(main):003:0> document = Document.find('foo')
...
DocumentFinder::RecordNotFound (Document: foo)

irb(main):004:0> document = Document.find('foo bar baz')
...
GdsApi::InvalidUrl (GdsApi::InvalidUrl)
```

I'm not sure that catching the exception and displaying a message is any more useful than just letting the exception give a stack trace and terminate the task.

lib/tasks/base_path.rake
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    if document
```
barrucadu: If the exception were allowed to propagate, above, then this check wouldn't be needed.

lib/tasks/base_path.rake
```diff
@@ -0,0 +1,24 @@
...
+      document.base_path = base_path
+      document.update_type = 'minor'
+      if document.save
+        puts "The #{document.class} with title \"#{document.title}\" has been successfully edited"
+        puts "Its new base_path is now \"#{document.base_path}\""
+      else
+        puts "Couldn't edit the #{document.class} with title \"#{document.title}\""
```
barrucadu: If saving fails, there will probably be an associated validation error.  Perhaps use `document.save!`, which throws an exception if there's something wrong.

barrucadu:
I've added a few comments, but I think they can be summarised into one guiding principle: if something is intended for developers, it can fail in a noisy way full of implementation details (eg, giving a stack trace and the process terminating) when it goes wrong.

lib/tasks/base_path.rake
AlanGabbianelli: Yes, I wasn't sure about this either, to be honest.
I followed the pattern that I saw already implemented in the code, but I'm usually a fan of letting things fail and dealing with the consequences.
I'll change it.

lib/tasks/base_path.rake
```diff
@@ -0,0 +1,24 @@
...
+    begin
+      document = Document.find(content_id)
+    rescue DocumentFinder::RecordNotFound => e
+      puts "Error finding the document: #{e.inspect}"
+    end
+
+    if document
+      document.base_path = base_path
+      document.update_type = 'minor'
+      if document.save
```
barrucadu: Before making the change it can be nice to print out what it will be.  Just something simple like

```
puts "#{document.base_path} -> #{base_path}"
```

AlanGabbianelli:
Yes, I replied to the first comment, and I agree, let it fail and deal with it. 😄

lib/tasks/base_path.rake
bevanloon: don't think there's a `save!` method here - specialist publisher ~~isn't a~~ doesn't use it's own local database ~~backed application~~ so this isn't a traditional ActiveRecord model.

rubenarakelyan (approved):

PR #1319, opened 2018-10-18
Add universal credit subcategories to utaac decisions

steventux:
https://trello.com/c/csn6hfq3/555-add-sub-categories-to-universal-credit-category-on-administrative-appeals-tribunal-decisions-finder

### DIFF
--- a/lib/documents/schemas/utaac_decisions.json
+++ b/lib/documents/schemas/utaac_decisions.json
@@ -2035,6 +2035,42 @@
           "value": "tribunal-procedure-and-practice-tribunal-practice"
         },
         {
+          "label": "Universal Credit - Sanctions",
+          "value": "universal-credit-sanctions"
+        },
+        {
+          "label":  "Universal Credit - Standard Allowance",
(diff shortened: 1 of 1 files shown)


lib/documents/schemas/utaac_decisions.json
```diff
@@ -2034,6 +2034,42 @@
           "label": "Tribunal procedure and practice - tribunal practice",
           "value": "tribunal-procedure-and-practice-tribunal-practice"
         },
+        {
+          "label": "Universal Credit - Sanctions",
+          "value": "universal_credit-sanctions"
```
thomasleese: In the other `value`s, we seem to use dashes to separate words even if the whole label has got a dash in it (and not use underscores and dashes). Although I think this is clearer, is it a problem that it's not consistent with the others?

steventux: Ah good point, I'll go with consistency :+1:

deborahchua (approved):

thomasleese (approved):

PR #1363, opened 2019-01-16
We should have prepositions on facets

sihugh:
The new facet tag presenter needs prepositions to make sense. We're specifying a default in finder frontend, but it's better if we can be specific.


bilbof (approved):
lgtm

PR #1382, opened 2019-02-21
Bugfix for minor Drug Safety Updates being published as major updates

bilbof:
https://trello.com/c/bzOxc6iS/463-bug-specialist-publisher-minor-updates-considered-major

This fixes a bug where a document would be sent to publishing_api
as a major update even if the publisher user had set the
update_type to minor.

The bug would occur because first_published_at could be sent as
nil when the user clicks publish (probably due to PR #1362).

Because first_published_at was nil, we would consider the document
to be a first draft, and set the update_type to major before
sending the document to the publishing_api.

This was reported as a bug via 2nd line at:
https://govuk.zendesk.com/agent/tickets/3603919

The fix copies the DfidResearchOutput approach,
which overrides first_draft? to use the state
history rather than the first_published_at date.

DfidResearchOutput and DrugSafetyUpdate are the only
classes that extend Document and also provide
a way to set a custom first_published_at,
therefore there isnt a need for further changes.

### DIFF
--- a/app/models/dfid_research_output.rb
+++ b/app/models/dfid_research_output.rb
@@ -36,14 +36,6 @@ class DfidResearchOutput < Document
     'DFID Research Output'
   end
 
-  def first_draft?
-    draft? && state_history_one_or_shorter?
-  end
-
-  def state_history_one_or_shorter?
-    state_history.nil? ? true : state_history.size < 2
-  end
-
   def dfid_author_tags
     (dfid_authors || []).join("::")
   end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -135,7 +135,11 @@ class Document
   end
 
   def first_draft?
-    draft? && first_published_at.blank?
+    draft? && state_history_one_or_shorter?
+  end
+
+  def state_history_one_or_shorter?
+    state_history.nil? || state_history.size < 2
   end
 
   def change_note_required?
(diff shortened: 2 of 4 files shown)


kevindew (commented):
This looks good - well dug into 👍 

I wonder if we want to just change the method in Document to be this and then not have to override it anywhere?

bilbof:
Yeah, it's not necessary to fix this bug, but others will unknowingly resurrect this bug if they do the same as https://github.com/alphagov/specialist-publisher/pull/1362 for other document types.

I didn't want to make that courageous decision, though. 😄 The [commit](https://github.com/alphagov/specialist-publisher/pull/840) adding the same change to the Dfid class also removed this logic from the document class as "preparatory changes for removing redrafted" so I expect it will break something else if we switch it back.

I made the same change to the Document class and ran the tests, and received this error:

```
Document successful #publish when document is in redrafted state behaves like publishing changes to a document that has previously been published does not add a "First published" change note before asking the api to publish
```

kevindew:
@bilbof interesting, the state part of Publishing API has always been a bit contentious. I don't think it's a good situation though to have two different ways so I think this change applied to all is the most consistent way forward until Publishing API changes.

Looking where that test might have failed - all the examples seem to use a trait of published https://github.com/alphagov/specialist-publisher/blob/e2782bcfb518ac497e978e48320e556675c4940a/spec/fixtures/factories.rb#L135-L141 which has just one item in the state history - which probably explains why it failed (though it all seems a bit confusing to me that it would try publish a published thing). I think that could be resolved by it using the redrafted trait for https://github.com/alphagov/specialist-publisher/blob/e2782bcfb518ac497e978e48320e556675c4940a/spec/models/document_spec.rb#L356

bilbof:
@kevindew sure thing. I've made the change you recommended. I lack a lot of context on the publishing workflow so let me know if anything looks odd.

kevindew (approved):
Looks good 👍

PR #1389, opened 2019-03-12
Add Sweden and remove Northern Ireland exception copy from the Exports Form Finder

AgaDufrat:
Adds ‘Sweden’ as a destination country, which was missing and removes the “Do not use this tool if you are exporting from Northern Ireland” callout text, since NI now follows the same process.

Screen Shot 2019-03-12 at 10 30 12

in Specialist publisher
Screen Shot 2019-05-10 at 13 27 14

[Trello card](https://trello.com/c/dZIPRphR/826-add-sweden-and-remove-northern-ireland-exception-copy-from-the-exports-form-finder)


thomasleese (approved):
Looks good to me!

PR #1407, opened 2019-04-09
Fix some broken bits of the NANDOs finder

sihugh:
There were some bits of the new specialist document set up for the UK market conformity bodies that didn't quite work, or were changed after seeing what they actually look like in the wild.

They are:
- We'd overridden the slug which meant that the logic to derive which metadata partial to use broke
- The labels on the form fields weren't quite in the right place which meant we fell back to using the default
- We had a `body name` field defined, but we can use the existing `title` field instead
- A small tweak to put UK at the top of the select drop down
- Added some tests which just make sure that basic stuff works OK

I've disabled some validation (temporarily) as we're about to create and partially import a bunch of things.  We'll re-enable this once the initial tranche is complete.

https://trello.com/c/ayXlI8nK/594-automatically-generate-the-public-bodies-for-ops-to-tag-to-the-nandos-finder

### DIFF
--- a/app/models/uk_market_conformity_assessment_body.rb
+++ b/app/models/uk_market_conformity_assessment_body.rb
@@ -10,7 +10,6 @@ class UkMarketConformityAssessmentBody < Document
   validates :uk_market_conformity_assessment_body_legislative_area, presence: true
 
   FORMAT_SPECIFIC_FIELDS = %i(
-    uk_market_conformity_assessment_body_name
     updated_at
     uk_market_conformity_assessment_body_number
     uk_market_conformity_assessment_body_type
--- a/app/models/uk_market_conformity_assessment_body.rb
+++ b/app/models/uk_market_conformity_assessment_body.rb
@@ -36,10 +36,6 @@ class UkMarketConformityAssessmentBody < Document
     "UK Market Conformity Assessment Body"
   end
 
-  def self.slug
-    "uk-market-conformity-assessment-body"
-  end
-
   def primary_publishing_organisation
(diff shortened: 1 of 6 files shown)


app/models/uk_market_conformity_assessment_body.rb
```diff
@@ -1,16 +1,15 @@
...
-  validates :uk_market_conformity_assessment_body_legislative_area, presence: true
+  # validates :updated_at, presence: true, date: true
+  # validates :uk_market_conformity_assessment_body_number, presence: true
+  # validates :uk_market_conformity_assessment_body_type, presence: true
+  # validates :uk_market_conformity_assessment_body_registered_office_location, presence: true
+  # validates :uk_market_conformity_assessment_body_testing_locations, presence: true
+  # validates :uk_market_conformity_assessment_body_website, presence: true
+  # validates :uk_market_conformity_assessment_body_email, presence: true
+  # validates :uk_market_conformity_assessment_body_phone, presence: true
+  # validates :uk_market_conformity_assessment_body_legislative_area, presence: true
```
bilbof, line 10: Could these be deleted? Or do we want to use them in future?

bilbof (approved):
LGTM

sophiebrown01 (approved):
Looks good 👍

app/models/uk_market_conformity_assessment_body.rb
sihugh: They'll be reactivated as soon as we've done the initial import 👍

sihugh:
Thanks both!

PR #1471, opened 2019-07-25
Use :hybrid mode cookie serialisation

barrucadu:
This accepts :json and :marshal cookies, and produces :json cookies.
Brakeman wants us to use :json and not :marshal or :hybrid, but
switching straight to :json from :marshal breaks existing sessions.

We'll have to override brakeman for this PR, as it will still fail.
But a little time in the future we can change to :json cookies.

Partially reverts #1470


thomasleese:
We can use [`brakeman -I` to ignore the error](https://docs.publishing.service.gov.uk/manual/brakeman.html#dealing-with-false-positives).

barrucadu:
I'll add the exclusion

thomasleese (approved):

PR #1479, opened 2019-08-15
[WIP] Add OPT to IDF, and rename Burma to Myanmar

edwardkerry:
Adds the Occupied Palestinian Territories to the International Development Funds, following this [Zendesk](https://govuk.zendesk.com/agent/tickets/3771248)

The OPT already exists as a world location
https://www.gov.uk/api/world-locations/the-occupied-palestinian-territories

Also renames Burma to Myanmar

### DIFF
--- a/lib/documents/schemas/international_development_funds.json
+++ b/lib/documents/schemas/international_development_funds.json
@@ -143,10 +143,6 @@
           "label": "Burkina Faso"
         },
         {
-          "value": "burma",
-          "label": "Burma"
-        },
-        {
           "value": "burundi",
           "label": "Burundi"
         },
@@ -519,6 +515,10 @@
           "label": "Mozambique"
         },
         {
+          "value": "myanmar",
+          "label": "Myanmar"
+        },
+        {
           "value": "namibia",
           "label": "Namibia"
         },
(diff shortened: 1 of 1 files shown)


lib/documents/schemas/international_development_funds.json
```diff
@@ -142,10 +142,6 @@
           "value": "burkina-faso",
           "label": "Burkina Faso"
         },
-        {
-          "value": "burma",
```
ChrisBAshton, line 146: I don't know enough about this repo to know if removing `burma` will cause regression issues (e.g. if there are things that rely on the `burma` value existing).

edwardkerry: I'm not sure either on the downstream implications of this file - but in the case of Burma it has recently been renamed to Myanmar across the rest of gov.uk by Platform Health so I think this change at least should be OK.

edwardkerry: https://trello.com/c/8Y2fbezn/879-country-name-change-burma-myanmar

thomasleese: Good catch!

We might need to check every published document of type `international_development_fund` and change any values of `burma` to `myanmar` before doing this, otherwise I suspect the finder will change to filter against Myanmar, but they'll be no documents tagged to it.

edwardkerry:
This has been picked up in subsequent PRs now

PR #1487, opened 2019-08-30
Present finder topic links to publishing-api

barrucadu:
These are defined in a few of the json files but not actually used
anywhere.  Apparently, the topic tagging was once managed by
panopticon (94d6879), but panopticon hasn't been around for quite a
while now...

---

https://govuk.zendesk.com/agent/tickets/3778731


barrucadu:
Actually it turns out we want a `parent` link for this after all

PR #1537, opened 2019-11-27
Replace govuk-lint with rubocop-govuk

laurentqro:
Replace deprecated govuk-lint with rubocop-govuk https://github.com/alphagov/rubocop-govuk, and fixes the shiny new linting errors.

### DIFF
--- a/app/controllers/attachments_controller.rb
+++ b/app/controllers/attachments_controller.rb
@@ -31,7 +31,7 @@ class AttachmentsController < ApplicationController
   def update
     document = fetch_document
     attachment = document.attachments.find(attachment_content_id)
-    attachment.update_attributes(attachment_params)
+    attachment.update_properties(attachment_params)
 
     if attachment.file.nil?
       save_updated_title(document, attachment)
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -112,7 +112,7 @@ private
   def document_error_messages
     heading = content_tag(:h4, "Please fix the following errors")
     errors = content_tag(:ul, class: "list-unstyled remove-bottom-margin") do
-      safe_join(@document.errors.full_messages.map { |message| content_tag(:li, message.html_safe) })
+      safe_join(@document.errors.full_messages.map { |message| content_tag(:li, message.html_safe) }) # rubocop:disable Rails/OutputSafety
     end
 
     heading + errors
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -28,7 +28,7 @@ class DocumentsController < ApplicationController
       redirect_to document_path(current_format.slug, @document.content_id)
     elsif @document.errors.any?
       flash.now[:errors] = document_error_messages
-      render :new, status: 422
+      render :new, status: :unprocessable_entity
     else
       flash.now[:danger] = unknown_error_message
       render :new
@@ -56,7 +56,7 @@ class DocumentsController < ApplicationController
       end
     else
       flash.now[:errors] = document_error_messages
-      render :edit, status: 422
+      render :edit, status: :unprocessable_entity
     end
   end
 
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -2,8 +2,10 @@ require "gds_api/publishing_api_v2"
 
 class DocumentsController < ApplicationController
   include ActionView::Context
+  include ActionView::Helpers::OutputSafetyHelper
   include ActionView::Helpers::TagHelper
   include ActionView::Helpers::TextHelper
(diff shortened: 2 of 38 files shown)


benthorner (commented):
What did I do to deserve this? 😉 

It looks like there are a couple of extra changes in here, so worth asking if they're intentional:

   - Previously we only linted specific directories, but now we're linting everything. I prefer this myself, because it makes it easier to auto-correct errors after running `rake lint`. Not that I write bad code, of course. Note that, unless it's [overridden in the Jenkinsfile](https://github.com/alphagov/content-publisher/blob/master/Jenkinsfile#L17), we'll only lint those old dirs on CI.

   - Previously we may have been running the Rails cops as well. I'm not sure. But I wonder if it's worth turning them on, as suggested [in the install instructions](https://github.com/alphagov/rubocop-govuk#installation). Depends on the number of errors to fix!

laurentqro:
@benthorner I added in Rails cops. The numerous resulting linting errors tell me that those were not turned on before.

I have this down to two linting errors:

```
Offenses:

app/controllers/documents_controller.rb:115:89: C: Rails/OutputSafety: Tagging a string as html safe may be a security risk.

      safe_join(@document.errors.full_messages.map { |message| content_tag(:li, message.html_safe) })

app/controllers/govspeak_controller.rb:15:35: C: Rails/OutputSafety: Tagging a string as html safe may be a security risk.
    render html: govspeak_preview.html_safe

```

They look acceptable to me, but would be keen to get your opinion on this.

I have done my best to make this PR more easily reviewable by looking at individual commits. i.e. 1 commit = 1 linting error.

app/helpers/publishing_helper.rb
```diff
@@ -1,30 +1,28 @@
 module PublishingHelper
-  def handle_remote_error
+  def handle_remote_error(document)
     begin
       yield
       true
     rescue GdsApi::HTTPErrorResponse => e
-      error_response_message(:base, e.message) if @publishable
+      error_response_message(document, :base, e.message)
```
benthorner, line 7: Are you sure we don't need the `if`?

app/presenters/actions_presenter.rb
```diff
@@ -18,21 +19,26 @@ def publish_button_visible?
 
   def publish_text
     if state == "published"
```
benthorner, line 21: Much better! I think we can simplify this even further by removing the `text` variable, since the `if` will implicitly return its last expression.

app/controllers/documents_controller.rb
```diff
@@ -104,23 +106,13 @@ def check_authorisation
...
-      },
-    )
-    errors = content_tag :ul, class: "list-unstyled remove-bottom-margin" do
-      list_items = @document.errors.full_messages.map do |message|
-        content_tag(:li, message.html_safe)
-      end
-      list_items.join.html_safe
+    heading = content_tag(:h4, "Please fix the following errors")
+    errors = content_tag(:ul, class: "list-unstyled remove-bottom-margin") do
+      safe_join(@document.errors.full_messages.map { |message| content_tag(:li, message.html_safe) })
```
benthorner: I think we can do `raw` instead of `html_safe` here, since I'm guessing `message` is just a string.

app/controllers/attachments_controller.rb
```diff
@@ -31,7 +31,7 @@ def edit
   def update
     document = fetch_document
     attachment = document.attachments.find(attachment_content_id)
-    attachment.update_attributes(attachment_params)
+    attachment.update_properties(attachment_params)
```
benthorner, line 34: Good fix - shame we didn't use `ActiveModel` in the first place 😠.

benthorner (commented):
Thanks for fixing all the errors @laurentqro! And the commits made it easy to review, too 😃. With the `html_safe` errors, I've commented on one specifically already; with the Govspeak one, I'd suggest [ignoring the error](https://github.com/alphagov/content-publisher/blob/master/app/controllers/govspeak_preview_controller.rb#L10), which should be fine since Govspeak should be sanitising stuff by default.

app/presenters/actions_presenter.rb
laurentqro: Good point!

app/helpers/publishing_helper.rb
laurentqro: The value of`@publishable` (a Document instance) is currently assigned in `set_errors`, and there only.

`handle_remote_error` needs this `if @publishable` because `error_response_message` calls the #errors method on it.

If we pass the Document instance to both `handle_remote_error` and `error_response_message`, I don't think we need presence check anymore.

What do you think?

benthorner: Sounds OK to me - worth a check 👍

barrucadu (approved):
LGTM

PR #1540, opened 2019-12-04
Specify email alert signup content item default filters

laurentqro:
We need to specify default filters in the email signup content item schema so that a user can sign up for alerts without having selected any facets on the search page.

https://github.com/alphagov/govuk-content-schemas/pull/940 will need to be merged and deployed first so that this PR can pass validation against the updated schema.

https://trello.com/c/OPAcIhIF/1193-cannot-sign-up-to-email-alerts-when-no-facets-are-selected


Tetrino (approved):
👍

PR #1548, opened 2019-12-17
Create marine notices specialist publisher 

laurentqro:
https://trello.com/c/TjKgnXWO/1210-create-a-functional-prototype-for-mca-specialist-finder

Related to:

* https://github.com/alphagov/govuk-content-schemas/pull/946
* https://github.com/alphagov/search-api/pull/1857

### DIFF
--- a/app/models/marine_notice.rb
+++ b/app/models/marine_notice.rb
@@ -1,0 +1,27 @@
+class MarineNotice < Document
+  validates :marine_notice_issued_date, presence: true, date: true
+  validates :marine_notice_type, presence: true
+  validates :marine_notice_vessel_type, presence: true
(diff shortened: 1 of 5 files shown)


bilbof (approved):
Nice work! I wonder if you could add tests for the model and view?

laurentqro:
> Nice work! I wonder if you could add tests for the model and view?

@bilbof, good point. I've added a model test, which includes a contract test. 👌 

Given we're already testing `Document`, `DocumentPresenter`, and `DocumentsController`, as well as the view for other specialist finders, I am unsure of the benefits of testing the view for the marine notice finder. Happy to hear your thoughts.

PR #1552, opened 2019-12-19
Add categories to Employment appeal tribunal decisions finder

AgaDufrat:
This adds 4 new categories

categories

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/3874840


JonathanHallam (approved):
Tested this on integration and it works fine

PR #1569, opened 2020-01-31
Remove redirect URL check when unpublishing

theseanything:
This removes the check to ensure that when unpublishing a document the redirect is pointing to live content. This because the publishing api currently doesn't support lookups on base path of multi-part content. Hence publishers are currently unable to redirect to multipart content such as guides. This behaviour is also inconsistent with other publishing applications.


sihugh (approved):
🚀

PR #1575, opened 2020-02-06
Add a stronger warning about republishing things

barrucadu:
The old warning makes it sound like drafts just get confused, but actually it's *everything*.

### DIFF
--- a/lib/tasks/republish.rake
+++ b/lib/tasks/republish.rake
@@ -4,11 +4,18 @@ require "csv"
 require "services"
 
 namespace :republish do
-  # Caution: These tasks should be avoided in production environnments  as
-  # they change the date information of draft documents and extra superfluous
-  # history to published documents.
+  ##########
   #
-  # If data is incorrect in Publishing API it should be fixed in Publishing API
+  # WARNING: do not run these tasks in production, unless absolutely necessary,
+  # as they will effectively shuffle the documents in the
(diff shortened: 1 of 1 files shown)


lib/tasks/republish.rake
```diff
@@ -4,9 +4,11 @@ require "csv"
 require "services"
 
 namespace :republish do
-  # Caution: These tasks should be avoided in production environnments  as
-  # they change the date information of draft documents and extra superfluous
-  # history to published documents.
+  # Do not run these tasks in production, unless absolutely necessary,
```
bilbof: Perhaps keeping "Caution:" or something like at the beginning would catch the eye :) Or perhaps a env flag one must set e.g. `DO_SOMETHING_DANGEROUS=true`

bilbof (approved):
Looks good 👍

edwardkerry (approved):
I would suggest even some capitals or WARNING: -like text

lib/tasks/republish.rake
barrucadu: I thought about a flag, but decided that we'd probably only want the flag in production.

It's fine to do in integration and staging, and if the flag is required *everywhere* there's the temptation to just copy and paste the command from integration to production without really thinking about it.

edwardkerry (approved):

PR #1576, opened 2020-02-06
Upload csv to s3 and link in email

benjamineskola:
Use S3 to transfer files, to be consistent with other emails and also compatible (in future) with Notify.

https://trello.com/c/yrMNgb9T/1714-3-specialist-publisher-make-documentlistexport-use-s3-instead-of-email-attachments

### DIFF
--- a/app/views/notifications_mailer/document_list.text.erb
+++ b/app/views/notifications_mailer/document_list.text.erb
@@ -1,3 +1,7 @@
 Hi <%= @user.name %>,
 
-Please find attached the CSV of <%= @document_klass.title.pluralize %><% if @query.present? %> matching the query "<%= @query %>"<% end %> you requested from GOV.UK specialist publisher.
+You requested a CSV of <%= @document_klass.title.pluralize %><% if @query.present? %> matching the query "<%= @query %>"<% end %> from GOV.UK specialist publisher.
+
+You can download this from the following address:
+
+<%= @url %>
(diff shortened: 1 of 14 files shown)


app/views/notifications_mailer/document_list.text.erb
```diff
@@ -1,3 +1,7 @@
 Hi <%= @user.name %>,
 
-Please find attached the CSV of <%= @document_klass.title.pluralize %><% if @query.present? %> matching the query "<%= @query %>"<% end %> you requested from GOV.UK specialist publisher.
+You requested a CSV of <%= @document_klass.title.pluralize %><% if @query.present? %> matching the query "<%= @query %>"<% end %> from GOV.UK specialist publisher.
+
+You can download this from the following address:
+
+<%= @url %>
```
cbaines, line 7: It would be good to say something in the commit message about why this change in general " Upload csv to s3 and link in email" is happening.

app/workers/document_list_export_worker.rb
```diff
@@ -46,4 +49,9 @@ def generate_csv(format, query)
       end
     end
   end
+
+  def upload_csv(filename, csv)
+    s3_file = S3FileUploader.save_file_to_s3(filename, csv)
```
cbaines, line 60: This doesn't exist right? At least not at this commit.

benjamineskola: Oh yeah. I added it in the next commit but actually those should be squashed together.

cbaines (approved):
Looks good to me :+1:

issyl0:
The steps here should be:

- Deploy Puppet (with secrets) all the way to Production.
- Deploy Specialist Publisher.

I'm hopeful we can do this today.

PR #1580, opened 2020-02-11
Add temporary rake task to redirect specific unpublished HealthExportCertificates

edwardkerry:
These documents were unpublished and set as "gone". There is a requirement for
them to be redirected. This Rake task can be removed after running.


lib/tasks/redirect_export_health_certificates.rake
```diff
@@ -0,0 +1,141 @@
...
+    "/export-health-certificates/export-a-live-equine-animal-destined-for-another-third-country-which-will-transit-the-european-union-certificate-8337",
+    "/export-health-certificates/export-wild-leporidae-farmed-rabbits-and-wild-land-mammals-other-than-ungulates-for-the-transit-through-or-storage-in-the-european-union-certificate-8339",
+    "/export-health-certificates/export-intermediate-products-to-be-used-for-the-manufacture-of-medicinal-and-cosmetic-products-intended-for-dispatch-to-or-for-transit-through-the-european-union-certificate-8341",
+    "/export-health-certificates/export-bone-horn-and-hoof-products-excluding-meal-intended-for-use-other-than-as-feed-material-organic-fertilisers-or-soil-improvers-for-dispatch-to-or-for-transit-through-the-european-union-certificate-8342",
+    "/export-health-certificates/export-untreated-wool-and-hair-to-the-european-union-certificate-8343",
+  ]
+  redirect_path = "/starting-to-export/licences"
+
+  paths.each do |path|
+    content_id = Services.publishing_api.lookup_content_ids(base_paths: path, exclude_unpublishing_types: [])
```
cbaines, line 136: You might be able to call this just once with all paths. That should return a hash from path to content_id for all the paths.

cbaines (approved):
I made one comment, but yeah, this looks good to me :+1:

PR #1593, opened 2020-02-25
Use GOV.UK Notify for ActionMailer messages

benjamineskola:
https://trello.com/c/BYnenCah/1774-specialist-publisher-use-notify-instead-of-amazon-ses


issyl0 (approved):
LGTM, pending secrets.

PR #1594, opened 2020-02-25
Remove leftover delivery_method parameter

benjamineskola:
This parameter was left over when configuring Notify in https://github.com/alphagov/specialist-publisher/pull/1593


issyl0 (approved):
Yay.

PR #1637, opened 2020-05-01
Update the marine_notice model spec

sihugh:
I changed the content schema to allow multiple options in https://github.com/alphagov/govuk-content-schemas/pull/984

I now want to remove the `string` option in the schema as per https://github.com/alphagov/govuk-content-schemas/pull/985 but I need to update this spec first.


laurentqro (approved):

laurentqro:
blast from the past 😄

PR #1693, opened 2020-08-25
Restore asset compress after Rails 6 upgrade

EGiataganas:
## What

Changes to `production.rb` to restore asset compression to previous output.

## Why

* The upgrade to Rails 6 disabled JS compression because it now defaults to using webpack.

## Visual differences

None.


pixeltrix (approved):

pixeltrix:
This is changing application.rb not production.rb - sorry I didn't notice before approving

PR #1697, opened 2020-08-28
Move International Development Funds to FCDO

edwardkerry:
The International Development Funds finder is moving to FCDO as the primary publishing organisation.

This PR updates the organisation ID and adds a summary to alert users to the change in ownership.


thomasleese (approved):

cbaines:
I think this is good to go, so merging.

PR #1699, opened 2020-08-28
Move Research For Development Outputs to FCDO

edwardkerry:
The Research for Development Output finder is moving to FCDO as the primary publishing organisation.

This PR updates the organisation ID and adds a summary to alert users to the change in ownership.


cbaines:
I think this is good to go, so merging.

cbaines (approved):

PR #1700, opened 2020-08-28
Add a Rake task for migrate DFID research outputs

thomasleese:
We're moving these to a new finder with a new document type, base path and metadata fields. This Rake task should migrate the documents one by one over to the new finder by republishing new editions of the document with the payload transformed to the new format.

[Trello Card](https://trello.com/c/wye58SZu/2109-update-x2-dfid-specialist-finders-to-fcdo)

### DIFF
--- a/lib/tasks/dfid.rake
+++ b/lib/tasks/dfid.rake
@@ -1,0 +1,26 @@
+namespace :dfid do
+  desc "Migrate research outputs"
+  task migrate_research_outputs: :environment do
+    content_ids = Republisher.content_ids_for_document_type("dfid_research_output")
+    content_ids.each do |content_id|
+      RepublishService.new.call(content_id) do |payload|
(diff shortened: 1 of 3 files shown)


lib/tasks/dfid.rake
```diff
@@ -0,0 +1,24 @@
+namespace :dfid do
+  desc "Migrate research outputs"
+  task migrate_research_outputs: :environment do
+    content_ids = Republisher.content_ids_for_document_type("dfid_research_output")
+    content_ids.each do |content_id|
```
thomasleese, line 5: We should find a way of ordering this by the `updated_at` so the ordering doesn't get messed up... `.reverse` could be used if we can't find a better way

thomasleese: I think this is resolved in d1dcef0.

edwardkerry (approved):

PR #1714, opened 2020-09-21
Country name change in schemas - Macedonia to North Macedonia

Tetrino:
Related to https://github.com/alphagov/govuk-content-schemas/pull/1009

### DIFF
--- a/lib/documents/schemas/export_health_certificates.json
+++ b/lib/documents/schemas/export_health_certificates.json
@@ -739,12 +739,6 @@
               "prechecked": false
             },
             {
-              "key": "macedonia",
-              "radio_button_name": "Macedonia",
-              "topic_name": "Macedonia",
-              "prechecked": false
-            },
-            {
               "key": "madagascar",
               "radio_button_name": "Madagascar",
               "topic_name": "Madagascar",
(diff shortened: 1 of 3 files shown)


Tetrino:
@cbaines as you are knowledgable in schema land, anything extra I need to do after this is brought in?

cbaines:
> @cbaines as you are knowledgable in schema land, anything extra I need to do after this is brought in?

Uhh, my first thought is about the existing data. Is that going to be changed/updated, and if so, how?

Tetrino:
That's my concern too. I'll peruse the docs and see what I can find.

Tetrino:
@cbaines Seems all I have to do is republish the specialist finders involved, I'd like to get this on integration to test this, I can revert if needs be.

cbaines:
> @cbaines Seems all I have to do is republish the specialist finders involved, I'd like to get this on integration to test this, I can revert if needs be.

Updating the finder sounds good, but I don't think updating the finder will update bits of content like https://www.gov.uk/export-health-certificates/export-porcine-animals-for-breeding-or-production-to-macedonia-certificate-7241 it's the existing content that I'm thinking about, as when the finder is changed there will then be a disparity between the content and the finder.

cbaines (approved):

Tetrino:
> Updating the finder sounds good, but I don't think updating the finder will update bits of content like https://www.gov.uk/export-health-certificates/export-porcine-animals-for-breeding-or-production-to-macedonia-certificate-7241 it's the existing content that I'm thinking about, as when the finder is changed there will then be a disparity between the content and the finder.

Looking at the underlying data in the publishing API it appears the document takes that information directly from the finder for most cases. However you are correct there are some extra steps that need to be taken to fix the metadata for existing documents. I'm working on a task for that now, and will merge this in when complete.

PR #1722, opened 2020-09-25
Add tasks to rename a country on all docs

benthorner:
https://trello.com/c/J6DOuBBp/990-country-updates

Closes: https://github.com/alphagov/publishing-api/pull/1855
Closes: https://github.com/alphagov/specialist-publisher/pull/1714

These should become part of the general process we have for doing
this across GOV.UK [1].

[1]: https://docs.publishing.service.gov.uk/manual/rename-a-country.html

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -218,6 +218,10 @@ class Document
     AllDocumentsFinder.all(page, per_page, query, document_type)
   end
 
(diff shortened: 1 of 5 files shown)


JonathanHallam (approved):
nice one, thanks

PR #1726, opened 2020-09-29
Fix locale select form options

MuriloDalRi:
This fixes #1725 where I added the language names to the select form. It was working but I refactored it and it broke.

Before:
Screenshot 2020-09-29 at 13 18 51

After:
Screenshot 2020-09-29 at 13 22 45


edwardkerry (approved):
makes sense 👍

PR #1732, opened 2020-10-09
Lint as part of the default rake task

kevindew:
We had expected that all apps were linting with the default rake task
and removed the linting step from the Jenkins library [1]. However it
seems this app was missed.

[1]: alphagov/govuk-jenkinslib#74

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,10 +4,10 @@ module ApplicationHelper
   end
 
   def locale_codes
-    I18n.t('language_names').keys
+    I18n.t("language_names").keys
   end
 
   def map_locale_names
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -4,10 +4,10 @@ def facet_options(form, facet)
   end
 
   def locale_codes
-    I18n.t('language_names').keys
+    I18n.t("language_names").keys
   end
 
   def map_locale_names
-    locale_codes.map { |l| [t("language_names.#{l}", locale: "en"), l] }.to_h
+    locale_codes.index_by { |l| t("language_names.#{l}", locale: "en") }
```
1pretz1, line 11: Much cleaner!

1pretz1 (approved):

PR #1743, opened 2020-10-27
Specialist Publisher - Bulk import of Protected food and Drink names

cintamani:
This PR introduces a new Rake task of which responsibility is to deal with bulk imports.

One import has been developed to bulk-import Protected Food and Drink names from a CSV file, of which examples were provided.

The import usage has been documented in the Rake task description.
When the process is run, it will provide useful messages regarding any document that could not be uploaded.
The code uses a `-ImportService` which is responsible to get the CSV::Row (or a hash of information) and save a document from them. This service is also responsible to construct useful error messages.
There is also a `-Parser` object which is used to parse the content of the Row data into the set of attributes that the ProtectedFoodDrinkName model expects.
The Rake task is responsible for parsing the CSV correctly, send each row in turn to the ImportService and report any error to stdout.

Test coverage is present, with a rake task test that functions as a full integration test for the feature.

Errors reporting looks like: 

```bash
ERROR - Document index: 1. Registered product name: . Title can't be blank
ERROR - Document index: 2. Registered product name: Irish Poteen/Irish Poitín. Status can't be blank
ERROR - Document index: 3. Registered product name: Irish Whiskey/Uisce Beatha Eireannach/Irish Whisky. Country can't be blank
ERROR - Document index: 4. Registered product name: Scotch Whisky. Date registration can't be blank
ERROR - Document index: 6. Registered product name: Tennessee whisky (Also spelled as “Tennessee whiskey”). Summary can't be blank. Register can't be blank
86 out of 91 records imported. 5 errors reported
```


spec/support/csvs/spirits_for_import.csv
```diff
@@ -0,0 +1,9 @@
+Registered product name,Status,Class or category of product,Protection type,Country of origin,Traditional term grapevine product category,Traditional term type,Traditional term language,Date of application,Date of UK registration,Date of original registration with the EU,Decision notice,Protection instrument,Date of publication of the instrument,Legislation,Summary,Product type,Internal notes,,,,,,,,
```
thomasleese, line 1: I think the convention is for these CSV files to be in `spec/fixtures`.

spec/support/load_rake_tasks.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.configure do |config|
```
thomasleese, line 1: Why do we need to do this?

lib/parsers/protected_food_drink_name_parser.rb
```diff
@@ -0,0 +1,229 @@
+class ProtectedFoodDrinkNameParser
```
thomasleese: I'd suggest structuring this class and the `ProtectedFoodDrinkNameImportService` one a little differently:

```
lib/importers/protected_food_drink_name.rb -> Importers::ProtectedFoodDrinkName
lib/importers/protected_food_drink_name/parser.rb -> Importers::ProtectedFoodDrinkName::Parser
```

I think because we don't have any other "parsers" or "services" in this app already, having a `lib/parsers` and `lib/services` directory with only a single file in feels a little strange. Plus using the structure above should mean the [autoloader works automatically](https://guides.rubyonrails.org/autoloading_and_reloading_constants.html) and we don't need to do any `require` statements.

thomasleese (commented):
This looks pretty good, I've just got a few comments about the file structure so it's consistent with our other apps.

spec/support/csvs/spirits_for_import.csv
cintamani: That was my first call but then I seen in the code that this is not correct. Currently, other files like this used in similar integration tests live in this folder (for example: https://github.com/alphagov/specialist-publisher/tree/master/spec/support/images ) while the `spec/fixtures` contains factories https://github.com/alphagov/specialist-publisher/tree/master/spec/fixtures

spec/support/load_rake_tasks.rb
cintamani: To load rake tasks and be able to test rake tasks

lib/parsers/protected_food_drink_name_parser.rb
cintamani: yes I was unsure on where to put this files :) will move them

spec/support/csvs/spirits_for_import.csv
thomasleese: Hmm you're right, that's unfortunately. Okay let's stick with this for now.

spec/support/load_rake_tasks.rb
thomasleese: I'd never seen it done like this before each test, I think we can avoid this and just put the `Rails.application.load_tasks` line in the `spec_helper.rb` like https://github.com/alphagov/content-publisher/blob/f633e044325f7d616e44ae1c6b041f0b9b52e1e0/spec/spec_helper.rb#L25. (`content-publisher` is one of our newer apps and often used as a guide for best practices)

cintamani: I can do that but is far away from best practices. 
If you end up calling `Rails.application.load_tasks` twice without a cleanup, your rake tasks will run twice at invocation in the code. 
If you load them once and don't clean up, they can be invoked only once. 
I solved bugs for it recently on a codebase so I am quite confident that this is the way to go. Also, this is not done before each test, only the ones with the correct tag will run this

thomasleese: Okay, that sounds reasonable 👍

thomasleese:
One of the tests is failing in our CI environment:

```
 ​1) rake import import:protected_food_and_drink_names with invalid data reports any data it could not import and the error preventing it from saving
    ​Failure/Error:
      ​expect { Rake::Task["bulk_imports:protected_food_and_drink_names"].invoke(csv_file) }.to \
        ​output(expected_output).to_stdout

      ​expected block to output "ERROR - Document index: 1. Registered product name: . Title can't be blank\nERROR - Document index: .... Summary can't be blank. Register can't be blank\n3 out of 8 records imported. 5 errors reported\n" to stdout, but output "ERROR - Document index: 1. Registered product name: . Title can't be blank\nERROR - Document index: ... as “Tennessee whiskey”). Register can't be blank\n3 out of 8 records imported. 5 errors reported\n"
      ​Diff:
      ​@@ -2,6 +2,6 @@
       ​ERROR - Document index: 2. Registered product name: Irish Poteen/Irish Poitín. Status can't be blank
       ​ERROR - Document index: 3. Registered product name: Irish Whiskey/Uisce Beatha Eireannach/Irish Whisky. Country can't be blank
       ​ERROR - Document index: 4. Registered product name: Scotch Whisky. Date registration can't be blank
      ​-ERROR - Document index: 6. Registered product name: Tennessee whisky (Also spelled as “Tennessee whiskey”). Summary can't be blank. Register can't be blank
      ​+ERROR - Document index: 6. Registered product name: Tennessee whisky (Also spelled as “Tennessee whiskey”). Register can't be blank
       ​3 out of 8 records imported. 5 errors reported
    ​# ./spec/lib/tasks/bulk_imports_spec.rb:28:in `block (4 levels) in <top (required)>'
    ​# /var/lib/jenkins/bundles/ruby/2.7.0/gems/webmock-3.9.1/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>
```

cintamani:
Ah Sorry Thomas! Will investigate that :)

lib/importers/protected_food_drink_name/parser.rb
```diff
@@ -0,0 +1,461 @@
...
+      end
+
+      def body
+        content = ""
+
+        # Product specification
+        if data["Product type"] != "Traditional term"
+          content += "## Product specification \n\n" \
+            "The product specification is not available on this site. " \
+            "Find out how to [get a product specification for a protected food name] " \
```
thomasleese:
```suggestion
            "Find out how to [get a product specification for a protected food name]" \
```

This space means the links aren't rendered correctly.

thomasleese (approved):

thomasleese (approved):

cintamani:
Thanks for the review @thomasleese  :) I'll leave you guys to do the merge, as I don't have the right permissions. Also, I see a Travis failure reported, shall we worry? I can't access the logs currently :)

PR #1745, opened 2020-10-28
Make summary optional for ProtectedFoodDrinkName

AlanGabbianelli:
This is a departmental request. See related [Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4197468).

Trello card: https://trello.com/c/XYFBR2Bt/2187-3-update-protected-food-names-frontend-template-make-summary-optional-for-this-finder

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -39,7 +39,7 @@ class Document
   alias_method :temporary_update_type?, :temporary_update_type
 
   validates :title, presence: true
-  validates :summary, presence: true
+  validates :summary, presence: true, unless: :protected_food_drink_name?
   validates :body, presence: true, safe_html: true, inline_attachments: true
   validates :update_type, presence: true, unless: :first_draft?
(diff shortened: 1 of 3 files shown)


callumknights (approved):

app/models/document.rb
```diff
@@ -39,7 +39,7 @@ def temporary_update_type
   alias_method :temporary_update_type?, :temporary_update_type
 
   validates :title, presence: true
-  validates :summary, presence: true
+  validates :summary, presence: true, unless: -> { protected_food_drink_name? }
```
ChrisBAshton: The syntax can be simplified slightly, like the examples below:

```suggestion
  validates :summary, presence: true, unless: :protected_food_drink_name?
```

ChrisBAshton (approved):
LGTM 👍 one small non-blocking suggestion.

app/models/document.rb
AlanGabbianelli: Ah, good point! 😉

PR #1749, opened 2020-11-08
Publishing in Bulk via Rake Task

cintamani:
This PR contains the code that implements a publishing in bulk feature.
The feature is accessible via Rake task. I used some code found in the `republish` task functionality. However, in the case of `publishing`, the code looks simpler as it does not have to validate states and states versions.
It will only work with documents that are currently in a Draft status.
This tasks also allow to run the publish all functionality on a subset of document types rather than all existing documents in `draft` mode.
Discussions has to happen still about the background job or in-line publiation. Looking at existing functionality, seems like a background job is the way to go.

### DIFF
--- a/app/lib/publisher.rb
+++ b/app/lib/publisher.rb
@@ -1,0 +1,58 @@
+require "services"
+
+class Publisher
+  def self.publish_all(types: nil)
+    new(types).publish_all
+  end
+
+  def publish_all
+    types.each do |document_type|
+      publish_document_type(document_type)
+    end
+  end
+
+private
+
+  attr_reader :types
+
+  def initialize(types)
+    @types = types.presence || document_types
+  end
+
+  def publish_document_type(document_type)
+    content_id_and_locale_pairs_for_document_type(
+      document_type,
+    ).each do |content_id, locale|
+      document = Document.find(content_id, locale)
+
+      unless Services.with_timeout(30) { document.publish }
+        Rails.logger.warn "Cannot publish document: #{content_id}:#{locale}"
+      end
+    end
+  end
+
+  def content_id_and_locale_pairs_for_document_type(document_type)
+    unless document_types.include?(document_type)
+      raise ArgumentError, "Unknown document_type: '#{document_type}'"
+    end
+
+    Services.with_timeout(30) do
+      Services.publishing_api.get_content_items(
+        document_type: document_type,
+        publication_state: "draft",
+        fields: %i[content_id locale],
+        per_page: 999_999,
+        order: "updated_at",
+      )["results"].map { |r| r.values_at("content_id", "locale") }
+    end
+  end
+
+  def document_types
+    @document_types ||= all_document_types
+  end
+
+  def all_document_types
+    Rails.application.eager_load!
+    Document.subclasses.map(&:document_type)
+  end
+end
--- a/lib/tasks/publish.rake
+++ b/lib/tasks/publish.rake
@@ -1,0 +1,15 @@
+# This tasks will publish documents that are in a draft publication state.
+
+require "services"
+
+namespace :publish do
+  desc "Publishes all documents of given types in a draft publication state.\n" \
+    "Usage: `rake publish:all[protected_food_drink_name,esi_fund]`"
+  task :all, [:document_types] => :environment do |_, args|
+    types = args.document_types.presence&.split(" ")&.compact
+
+    raise ArgumentError, "No type given." if types.empty?
+
+    Publisher.publish_all(types: types)
+  end
+end
--- a/spec/lib/publisher_spec.rb
+++ b/spec/lib/publisher_spec.rb
@@ -1,0 +1,39 @@
+require "rails_helper"
+
+RSpec.describe Publisher do
+  before do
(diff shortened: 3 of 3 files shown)


app/lib/publisher.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  def initialize(types)
+    @types = types.presence || document_types
+  end
+
+  def publish_document_type(document_type)
+    content_id_and_locale_pairs_for_document_type(
+      document_type,
+    ).each do |content_id, locale|
+      PublishWorker.perform_async(content_id, locale)
```
thomasleese: I reckon we can get away without doing this asynchronously. Publishing documents via the Publishing API is usually very quick and doing it synchronously will give us a better idea of progress (i.e. when the Rake tasks finishes, the requests have finished) plus it'll crash dramatically if an error does occur and give us better visibility.

lib/tasks/publish.rake
```diff
@@ -0,0 +1,13 @@
+# This tasks will publish documents that are in a draft publication state.
+
+require "services"
+
+namespace :publish do
```
thomasleese, line 5: I think we should call this something different because it's not quite doing the same as [pressing "Publish" in the app would do](https://github.com/alphagov/specialist-publisher/blob/master/app/services/document_publisher.rb) (it doesn't send emails or set the update type.

I'd suggest calling it `bulk_imports:publish` since it's closed related to the other bulk import task.

cintamani: no, this actually does exactly the same: https://github.com/alphagov/specialist-publisher/blob/master/app/controllers/documents_controller.rb#L64 as you can see here

cintamani: that service is called by: https://github.com/alphagov/specialist-publisher/blob/master/app/models/document.rb#L237 this just triggers it

app/lib/publisher.rb
cintamani: Your call, really :) I have no preference no history of why other similar features are done in asynch. Will get rid of the job :)

lib/tasks/publish.rake
thomasleese: Oh yes you're right, I must've misread something!

lib/tasks/publish.rake
```diff
@@ -0,0 +1,13 @@
...
+  desc "Publishes all documents in a draft publication state.\n" \
+    "Usage: `rake publish:all` or with types `rake publish:all protected_food_dring_name esi_fund`"
+  task :all, [:document_types] => :environment do |_, args|
+    types = args.document_types.presence&.split(" ")
```
thomasleese:
```suggestion
    types = args.extras
```

I'd suggest doing this then the invocation is `rake publish:all[protected_food_drink_name,esi_find]` which feels more like a traditional Rake task.

lib/tasks/publish.rake
```diff
@@ -0,0 +1,13 @@
+# This tasks will publish documents that are in a draft publication state.
+
+require "services"
+
+namespace :publish do
+  desc "Publishes all documents in a draft publication state.\n" \
+    "Usage: `rake publish:all` or with types `rake publish:all protected_food_dring_name esi_fund`"
```
cintamani:
```suggestion
    "Usage: `rake publish:all` or with types `rake publish:all[protected_food_drink_name,esi_fund]`"
```

lib/tasks/publish.rake
```diff
@@ -0,0 +1,13 @@
...
+  task :all, [:document_types] => :environment do |_, args|
```
cintamani, line 8:
```suggestion
  task all: :environment do |_, args|
```

lib/tasks/publish.rake
thomasleese:
```suggestion
    "Usage: `rake publish:all` or with types `rake publish:all[protected_food_drink_name esi_fund]`"
```

Without that it actually ends up publishing everything and then trying to run a task called `protected_food_drink_name`. I'd actually suggest that we make it not possible to run this task and publish everything, because it's highly unlikely we'd ever want to do that and has a risk that someone does it accidentally.

cintamani: Glad you said that :D I actually added the ability to call it on a specific set of types, because the original requirement was to publish all content with one task regardless! And I asked twice.  
Will remove the ability to do that now :)

app/lib/publisher.rb
```diff
@@ -0,0 +1,68 @@
...
+  def document_types
+    @document_types ||= all_document_types
+  end
+
+  def all_document_types
+    Rails.application.eager_load!
+    Document.subclasses.map(&:document_type)
+  end
+
+  def with_timeout(seconds)
```
thomasleese: I think we can use `Services.with_timeout` here now.

cintamani: Yes! And I did that change. I must have messed around too much with git again :( will fix it shortly

cintamani: Ah no git mess. I just didn't get rid of this implementation :) should be fine now. Will rebase again

thomasleese (approved):

PR #1758, opened 2020-11-24
Specialist food changes

beccapearce:
Make changes to the protected food specialist document:
- only show the fields register, country of origin and date of registration on the search page
- update the schema to show summaries on the search page
-  change the value in the register from `Foods: designated origin and geographic origin` to `Foods: designated origin and geographical indication` and `Foods: traditional speciality` to `Foods: traditional speciality guaranteed`
- Add an option `Names protected by international treaty` to the `register` and `protection_type`
- Make `Traditional term grapevine product category` able to accept multiple values
- Rename the field `country` to `country_of_origin`
- Add a field to record `reason_for_protection`
- Change the `date_application` `date_registration` and `date_registration_eu` to be a date instead of a string

Trello card: https://trello.com/c/x2TYiSqT/2181-5-make-changes-to-protected-food-names-document-and-finder

### DIFF
--- a/lib/importers/protected_food_drink_name/parser.rb
+++ b/lib/importers/protected_food_drink_name/parser.rb
@@ -12,7 +12,7 @@ module Importers
           status: status,
           class_category: class_category,
           protection_type: protection_type,
-          # reason_for_protection: reason_for_protection, #TODO
+          reason_for_protection: reason_for_protection,
           country_of_origin: country,
           traditional_term_grapevine_product_category: traditional_term_grapevine_product_category,
           traditional_term_type: traditional_term_type,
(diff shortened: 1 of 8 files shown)


lib/importers/protected_food_drink_name/parser.rb
```diff
@@ -13,7 +13,7 @@ def get_attributes
           class_category: class_category,
           protection_type: protection_type,
           # reason_for_protection: reason_for_protection, #TODO
```
thomasleese: I think we should be able to uncomment this now, as the field is there. It was left commented out until we'd done this work!

thomasleese (approved):

PR #1773, opened 2020-12-07
Use StringEx to generate slugs

thomasleese:
See individual commits for more details, but this library seems to support more characters than the built in Rails transliteration and should avoid any issues we've been seeing with empty string outputs.

[Trello Card](https://trello.com/c/oedYOWVD/2247-updates-to-protected-food-drink-names-finder)


spec/models/protected_food_drink_name_spec.rb
```diff
@@ -16,4 +16,12 @@
       expect(document.base_path).to eq("/protected-food-drink-names/sterea-ellada")
     end
   end
+
+  context "with Cryllic characters" do
```
benjamineskola: "Cyrillic" :)

thomasleese: Oops, I've pushed up a fix!

spec/models/protected_food_drink_name_spec.rb
```diff
@@ -16,4 +16,12 @@
       expect(document.base_path).to eq("/protected-food-drink-names/sterea-ellada")
```
ChrisBAshton, line 16: Am I right in assuming that the Stringex approach generates the same slug as the manual Greek transliteration approach?

ChrisBAshton (approved):
LGTM 👍  I wondered if this might change how English slugs are generated too, but the [README](https://github.com/rsl/stringex) looks pretty predictable, so I wouldn't have thought we'd run into that.

And it looks as though this code is only called on "first draft", so we don't have to worry about existing slugs, do we?

spec/models/protected_food_drink_name_spec.rb
thomasleese: Yeah, that's correct 👍

thomasleese:
> And it looks as though this code is only called on "first draft", so we don't have to worry about existing slugs, do we?

That's right, this will only affect new editions.

thomasleese:
This supersedes https://github.com/alphagov/specialist-publisher/pull/1771.

PR #1780, opened 2020-12-16
Add a rake task to move medical alerts from one category to another

benjamineskola:
This could probably be made more general and thus more useful in future (it's based on the country-rename tasks) but that seems like overkill for the current situation

https://trello.com/c/xnGP5kS0/2236-5-mhra-request-to-rename-specialist-publisher-alert-categories

### DIFF
--- a/lib/tasks/medical_safety_alerts.rake
+++ b/lib/tasks/medical_safety_alerts.rake
@@ -1,0 +1,12 @@
+namespace :medical_safety_alerts do
+  desc "Change type of medical alerts"
+  task :change_alert_type, %i[old new] => :environment do |_, args|
+    MedicalSafetyAlert.find_each do |doc|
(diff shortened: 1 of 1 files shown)


edwardkerry (approved):
this makes sense, could we add an example of what an alert_type should look like? 
`"Alert type"` vs `:alert_type`, to make sure we don't accidentally update it with the wrong type?
May not be needed if this is a one off task!

lib/tasks/retag_editions.rake
```diff
@@ -0,0 +1,12 @@
+namespace :medical_alerts do
```
thomasleese: I think it would be good for this to match the filename. Being extra picky, should we call this `medical_safety_alerts` to match the model name exactly?

PR #1802, opened 2021-01-19
Add new land use option

andysellick:
- requested by a user in support ticket 4431297
- should add a new option to the filters under 'Land use' on https://www.gov.uk/countryside-stewardship-grants


kevindew (approved):
Nice one

PR #1807, opened 2021-01-22
Create a "medicines recall/notification" medical alert type

benjamineskola:
Two categories of medical safety alert are to be merged into one. This creates the single new category; the next step will be to move the old documents and users into this category, and finally to remove the old categories.

https://trello.com/c/GvEQmeaW/2328-further-changes-to-mhra-finder-categories

cf. #1710 for an alert type being added, and #1780 for the scripts to move documents between alert types.


AlanGabbianelli (approved):
LGTM together with https://github.com/alphagov/govuk-content-schemas/pull/1037

PR #1812, opened 2021-01-27
Allow signing up to employment tribunal decisions by category

thomasleese:
This makes it possible to pick a category (or "Jurisdiction code") when signing up to emails. This was apparently something that used to work, but I can't see it in the history of this file.

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4441728)


GDSNewt (approved):
That is weird that we would have no record of it changing, nonetheless this makes sense to me 👍

PR #1823, opened 2021-03-01
Remove duplicate guatemala entry

thomasleese:
It's listed twice by accident, but we don't need that to be the case.


ChrisBAshton (approved):
Is it worth thinking about adding some validation to prevent this kind of thing from happening again?

thomasleese:
> Is it worth thinking about adding some validation to prevent this kind of thing from happening again?

That's a good point, I'll make a note to write up a card for this.

PR #1845, opened 2021-04-08
Move Service Standard Reports to CDDO

1pretz1:
The PR does two things, it transfers ownership from GDS to CDDO of:

* the [Service Standard Reports finder][1]
* [Service Standard Reports themselves][2] and updates relevant page content 

For these change to take effect we will have to republish the Service Standard Reports finder 
and run the rake task contained within this PR.

## Primary organisation change

Screenshot 2021-04-08 at 11 20 12

## Content change

Screenshot 2021-04-07 at 17 13 25

## Finder change

Screenshot 2021-04-07 at 16 34 34

Trello:
https://trello.com/c/rKlR91M9/2463-content-tagging-for-gds-mog-change-cddo-specialist-publisher-content

[1]: https://www.gov.uk/service-standard-reports
[2]: https://www.gov.uk/service-standard-reports/file-your-confirmation-statement-alpha-assessment-report

### DIFF
--- a/lib/tasks/gds_to_cddo_tmp.rake
+++ b/lib/tasks/gds_to_cddo_tmp.rake
@@ -1,0 +1,19 @@
+desc "Migrate Service Standard Reports from GDS to CDDO "
+task gds_to_cddo_tmp: :environment do
+  content_ids = Republisher.content_id_and_locale_pairs_for_document_type("service_standard_report")
+
+  content_ids.each do |content_id, locale|
+    RepublishService.new.call(content_id, locale) do |payload|
+      gds_markdown = "From: | [Government Digital Service](https://www.gov.uk/government/organisations/government-digital-service)"
+      cddo_markdown = "From: | [Central Digital and Data Office](https://www.gov.uk/government/organisations/central-digital-and-data-office)"
+
+      payload[:details][:body].each do |data|
+        if data[:content].present?
+          data[:content] = data[:content].gsub(gds_markdown, cddo_markdown)
+        end
+      end
+
+      payload
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/tasks/gds_to_cddo_tmp.rake
```diff
@@ -0,0 +1,19 @@
+desc "Migrate Service Standard Reports from GDS to CDDO "
+task gds_to_cddo_tmp: :environment do
+  content_ids = Republisher.content_id_and_locale_pairs_for_document_type("service_standard_report")
+
+  content_ids.each do |content_id, locale|
+    RepublishService.new.call(content_id, locale) do |payload|
+      gds_markdown = "From: | [Government Digital Service](https://www.gov.uk/government/organisations/government-digital-service)"
```
edwardkerry, line 7: Is it OK for these to be `www.gov.uk` - how are they presented on the other environments?

1pretz1: I think so yes, as it mirrors the current behaviour in the other test environments (they all have Production URL's). In a perfect world these would be environment specific but I don't think that's possible to persist due to the data sync.

edwardkerry (approved):

sengi:
It looks like the database migration in this change updated all past Service Standard reports to say "From: Central Digital and Data Office" rather than just those which were published after the organisational change.

https://govuk.zendesk.com/agent/tickets/5176455

1pretz1:
Thanks @sengi - the original Trello card did specify all documents but doesn't look like this was intended. I've replied with a more detailed response in the linked Zendesk ticket.

sengi:
Thanks @1pretz1 🙌

PR #1867, opened 2021-05-20
Update FCERM finder to filter on date of completion

brucebolt:
The FCERM (Flood and Coastal Erosion Risk Management) finder currently uses the published date as a filter facet. This changes that to the date of completion, which is more useful to the users of this finder.

Make live by running this rake task:
```
publishing_api:publish_finder[flood_and_coastal_erosion_risk_management_research_reports]
```

Trello card: https://trello.com/c/yNBVwh6S

### DIFF
--- a/lib/documents/schemas/flood_and_coastal_erosion_risk_management_research_reports.json
+++ b/lib/documents/schemas/flood_and_coastal_erosion_risk_management_research_reports.json
@@ -176,7 +176,7 @@
 
     {
       "key": "date_of_start",
-      "name": "Start date",
+      "name": "Date of start",
       "type": "date",
       "display_as_result_metadata": true,
       "filterable": false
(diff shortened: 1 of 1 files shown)


thomasleese (commented):
Changes look good, I can't remember if this is in the card, but I wonder if should change the `start_date` name to `Date of start` as well so they're consistent when they show up in the result metadata?

thomasleese (approved):

brucebolt:
> Changes look good, I can't remember if this is in the card, but I wonder if should change the `start_date` name to `Date of start` as well so they're consistent when they show up in the result metadata?

Good idea.  I've made the change.

Strangely, the label was already 'Date of start' in the search results.  I'm not sure where that came from?  🤔 
Screenshot 2021-05-20 at 13 22 37

thomasleese:
> Strangely, the label was already 'Date of start' in the search results. I'm not sure where that came from? 🤔

Oh that's weird, yeah, no idea what this field even affects then!

PR #1873, opened 2021-06-08
Update medical safety alerts name

gclssvglx:
Change the medical saftey alerts name from 'Alerts and recalls for drugs and medical devices'
to 'Alerts, recalls and safety information: drugs and medical devices'.

See https://govuk.zendesk.com/agent/tickets/4587261


gclssvglx:
There are a [few](https://github.com/search?q=org%3Aalphagov+1e9c0ada-5f7e-43cc-a55f-cc32757edaa3&type=code) places where this content item is referenced, but it looks like this is the only that needs changing.

owenatgov (approved):

PR #1832, opened 2021-03-16
Review and revise README to refer to GOV.UK Docker

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This follows the pattern established in [1].

Please see the commit messages for more details.

[1]: https://github.com/alphagov/content-publisher/pull/2257

PR #1912, opened 2021-09-08
Remove screenshots from README

ChrisBAshton:
Following the pattern established in https://github.com/alphagov/content-publisher/pull/2257,
and a precedent rolled out to other apps, we no longer include
screenshots in READMEs because they:

- Are often out of date
- Take up a lot of space
- Their aim can be achieved by clicking on one of the live examples
  instead.

I think this got missed from #1832.


benthorner (approved):
🧹

PR #1910, opened 2021-09-07
Create new specialist document type: OIM Project

hannako:
### What

Extend specialist publisher to allow a new type of document called `OIM Projects` to be published.

### Why

Department [request](https://govuk.zendesk.com/agent/tickets/4690504)

### Related PR's
- [govuk_content_schemas](https://github.com/alphagov/govuk-content-schemas/pull/1071)
- [search_api](https://github.com/alphagov/search-api/pull/2326)
- [email-alert-api](https://github.com/alphagov/email-alert-api/pull/1664)

### Screenshots
Screenshot 2021-09-08 at 15 42 10
Screenshot 2021-09-08 at 15 42 20

[trello](https://trello.com/c/vN0SFEyb/2700-create-new-specialist-document-and-finder-office-for-the-internal-market-projects)

PR #1933, opened 2021-10-29
Create new specialist document type: Product safety alert

ChrisBAshton:
This is for the new "Product safety alerts, unsafe products and
recalls" finder. I've followed https://docs.publishing.service.gov.uk/apps/specialist-publisher/creating-a-new-specialist-document-type.html,
but also taken inspiration from the most recent PR #1910.

Screenshots:

## Index page

> Screenshot 2021-10-29 at 10 50 27

## New page

> Screenshot 2021-10-29 at 10 52 15

Trello: https://trello.com/c/ojsFViiX/2746-create-new-specialist-document-and-finder-product-safety-alerts-unsafe-products-and-recalls-3

### DIFF
--- a/app/models/product_safety_alert.rb
+++ b/app/models/product_safety_alert.rb
@@ -1,0 +1,29 @@
+class ProductSafetyAlert < Document
+  validates :product_alert_type, presence: true
+  validates :product_risk_level, presence: true
+  validates :product_category, presence: true
+  validates :product_measure_type, presence: true
+  validates :product_recall_alert_date, allow_blank: true, date: true
+
+  FORMAT_SPECIFIC_FIELDS = %i[
+    product_alert_type
+    product_risk_level
+    product_category
+    product_measure_type
+    product_recall_alert_date
+  ].freeze
+
+  attr_accessor(*FORMAT_SPECIFIC_FIELDS)
+
+  def initialize(params = {})
+    super(params, FORMAT_SPECIFIC_FIELDS)
+  end
+
+  def self.title
+    "Product Safety Alert"
+  end
+
+  def primary_publishing_organisation
+    "a0ee18e7-9e1e-4ba1-aed5-f3f287dce752"
+  end
+end
--- a/lib/documents/schemas/product_safety_alerts.json
+++ b/lib/documents/schemas/product_safety_alerts.json
@@ -1,0 +1,144 @@
+{
+  "content_id": "a22b3f1f-2c91-49a1-a469-ff21d465c543",
+  "base_path": "/product-safety-alerts",
+  "format_name": "Product safety alerts, unsafe products and recalls",
+  "name": "Product safety alerts, unsafe products and recalls",
+  "description": "Find product safety alerts, unsafe products and recalls",
+  "filter": {
+    "format": "product_safety_alert"
+  },
+  "signup_content_id": "bb6047de-3854-4774-a5d5-fc66fed4f792",
+  "signup_copy": "You'll get an email each time an alert is updated or a new alert is published.",
+  "organisations": ["a0ee18e7-9e1e-4ba1-aed5-f3f287dce752"],
+  "topics": [],
+  "email_filter_by": "product_alert_type",
+  "email_filter_facets": [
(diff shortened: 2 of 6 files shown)


app/models/product_safety_alert.rb
```diff
@@ -0,0 +1,33 @@
...
+    product_recall_alert_date
+  ].freeze
+
+  attr_accessor(*FORMAT_SPECIFIC_FIELDS)
+
+  def initialize(params = {})
+    super(params, FORMAT_SPECIFIC_FIELDS)
+  end
+
+  def taxons
```
1pretz1: Not sure we need to define the `taxons` method here as the parent model has the same method, with the same output (empty array): https://github.com/alphagov/specialist-publisher/blob/main/app/models/document.rb#L84-L86.

lib/documents/schemas/product_safety_alerts.json
```diff
@@ -0,0 +1,144 @@
...
+  "description": "Find product safety alerts, unsafe products and recalls",
+  "filter": {
+    "format": "product_safety_alert"
+  },
+  "signup_content_id": "bb6047de-3854-4774-a5d5-fc66fed4f792",
+  "signup_copy": "You'll get an email each time an alert is updated or a new alert is published.",
+  "organisations": ["a0ee18e7-9e1e-4ba1-aed5-f3f287dce752"],
+  "topics": [],
+  "email_filter_by": "product_alert_type",
+  "email_filter_facets": [
```
1pretz1, line 15: I've just looked at the associated card and it mentions that there should (?) be a subscription option for multiple facets, rather than just the 'Alert type', as per:

```
**Facets for the Finder:**
Users need to be able to filter and/or subscribe by :
- Alert Type
- Risk Level
- Product Category
- Measure Type
- Recall/alert date (before and after picker, like https://www.gov.uk/drug-device-alerts)
```

The comment isn't really clear on what subscription options are needed (and/or). Has there been an active choice here?

1pretz1 (commented):
Looks good @ChrisBAshton - thanks for providing screenshots 📸 

Just have one minor suggestion and a query.

lib/documents/schemas/product_safety_alerts.json
ChrisBAshton: Great spot - thanks @1pretz1! I've pushed up a commit that I think fixes that (lack of documentation, but I've explained my reasoning in the commit). Would be great to get a second pair of eyes on that.
 
I _think_ the new commit means we can close https://github.com/alphagov/email-alert-api/pull/1672 (based on https://docs.publishing.service.gov.uk/apps/specialist-publisher/creating-a-new-specialist-document-type.html#configure-the-email-sign-up-page) 🤔

1pretz1: Ah, nice! The `all_selected_facets` option does seem less user friendly though as the user doesn't get a check box / confirmation of what facets they're subscribing to (until they've actually subscribed), for example https://www.gov.uk/residential-property-tribunal-decisions/email-signup (all facets option) vs https://www.gov.uk/cma-cases/email-signup. I think this happens due to the [logic in Finder Frontend](https://github.com/alphagov/finder-frontend/blob/main/app/presenters/subscriber_list_params_presenter.rb#L36-L38).

However, if want to enable email subscriptions for all the facets then this seems the only option. Hmm, should we ask BEIS to clarify if they actually want all facets, or just go with the `all_selected_facets`?

> I think the new commit means we can close alphagov/email-alert-api#1672

I believe you're correct. Could we also remove the `email_filter_facets` hash from this schema if we use `all_selected_facets`?

ChrisBAshton: Good shout. We're checking with BEIS. On hold until then.

ChrisBAshton: BEIS have decided to go with the standard `product_alert_type`. Could you re-review please, @1pretz1 ?

1pretz1 (approved):
LGTM! 🍏

PR #1935, opened 2021-11-03
Reword summary for UKMCAB

benjamineskola:
https://trello.com/c/Eq7yU0wm/2756-changes-to-ukmcab-specialist-finder-3


hannako (changes requested):
You need to add the new fields [here](https://github.com/alphagov/specialist-publisher/blob/main/lib/documents/schemas/uk_market_conformity_assessment_bodies.json#L108)

And then republish the finder, see step 3 [here](https://docs.publishing.service.gov.uk/apps/specialist-publisher/creating-a-new-specialist-document-type.html#deploy-and-publish)

ps sorry I reviewed your PR as jenkins! 🙈

benjamineskola:
> You need to add the new fields [here](https://github.com/alphagov/specialist-publisher/blob/main/lib/documents/schemas/uk_market_conformity_assessment_bodies.json#L108)

Added in #1937 since it feels separate from the change of description.

hannako (approved):
👍

PR #1969, opened 2022-02-07
Create DRCF digital markets research finder

baisa:
Following CMA request to create a specialist finder to bring together research from several organisations.

[Trello card](https://trello.com/c/0LHrDag7/157-create-new-specialist-finder-for-competition-and-markets-authority)

### DIFF
--- a/app/views/metadata_fields/_drcf_digital_markets_researches.html.erb
+++ b/app/views/metadata_fields/_drcf_digital_markets_researches.html.erb
@@ -1,0 +1,62 @@
+<%= render layout: "shared/form_group", locals: { f: f, field: :digital_market_research_category, label: "Category" } do %>
+  <%= f.select :digital_market_research_category,
+      f.object.facet_options(:digital_market_research_category),
+      {},
+      {
+        class: 'form-control',
+        multiple: true,
+        data:
+        {
+          placeholder: 'Select a category'
+        }
+      }
+  %>
+<% end %>
+<%= render layout: "shared/form_group", locals: { f: f, field: :digital_market_research_publisher, label: "Publisher" } do %>
+  <%= f.select :digital_market_research_publisher,
+      f.object.facet_options(:digital_market_research_publisher),
+      {},
+      {
+        class: 'select2 form-control',
+        multiple: true,
+        data:
+        {
+          placeholder: 'Select a publisher'
+        }
+      }
+  %>
+<% end %>
+<%= render layout: "shared/form_group", locals: { f: f, field: :digital_market_research_area, label: "Research area" } do %>
+  <%= f.select :digital_market_research_area,
+      f.object.facet_options(:digital_market_research_area),
+      {},
+      {
+        class: 'select2 form-control',
+        multiple: true,
+        data:
+        {
+          placeholder: 'Select a research area'
+        }
+      }
+  %>
+<% end %>
+<%= render layout: "shared/form_group", locals: { f: f, field: :digital_market_research_topic, label: "Research topic" } do %>
+  <%= f.select :digital_market_research_topic,
+      f.object.facet_options(:digital_market_research_topic),
+      {},
+      {
+        class: 'select2 form-control',
+        multiple: true,
+        data:
+        {
+          placeholder: 'Select a research topic'
+        }
+      }
+  %>
+<% end %>
+<%= render "shared/date_fields",
+  f: f,
+  field: :digital_market_research_publish_date,
+  format: :drcf_digital_markets_research,
+  label: "Research publish date"
+%>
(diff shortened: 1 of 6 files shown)


benjamineskola (approved):

lib/documents/schemas/drcf_digital_markets_researches.json
```diff
@@ -0,0 +1,115 @@
+{
+  "content_id": "380def5e-bed2-4501-967e-334767ac270d",
+  "base_path": "/find-digital-market-research",
+  "format_name": "DRCF digital markets research",
+  "name": "DRCF digital markets research",
+  "description": "DRCF digital markets research finder",
+  "filter": {
+    "format": "find_digital_market_research"
```
sihugh: This should be `drcf_digital_markets_research` to match the content type

app/views/metadata_fields/_drcf_digital_markets_researches.html.erb
```diff
@@ -0,0 +1,61 @@
...
+      f.object.facet_options(:digital_market_research_category),
+      {},
+      {
+        class: 'select2 form-control',
+        data:
+        {
+          placeholder: 'Select a category'
+        }
+      }
+  %>
```
kevindew, line 14: As I understand we were changing this to be a single select rather than a select 2?

spec/fixtures/factories.rb
```diff
@@ -645,4 +650,20 @@
...
+  factory :drcf_digital_markets_research, parent: :document do
+    base_path { "/find-digital-market-research/example-document" }
+    document_type { "drcf_digital_markets_research" }
+    transient do
+      default_metadata do
+        {
+          "digital_market_research_category" => "ad-hoc-research",
+          "digital_market_research_publisher" => "gambling-commission",
+          "digital_market_research_area" => "media-and-entertainment",
+          "digital_market_research_topic" => "future-connectivity",
```
kevindew: I imagine that these should be arrays since multiple can be selected.

spec/features/creating_a_drcf_digital_markets_research_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+    expect(page.status_code).to eq(422)
+
+    expect(page).to have_css(".elements-error-summary")
+    expect(page).to have_css(".form-group.elements-error")
+    expect(page).to have_css(".elements-error-message")
+
+    expect(page).to have_content("Please fix the following errors")
+    expect(page).to have_content("Title can't be blank")
+    expect(page).to have_content("Summary can't be blank")
+    expect(page).to have_content("Body can't be blank")
```
kevindew, line 35: Is this missing a category validation?

spec/features/creating_a_drcf_digital_markets_research_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+  scenario "creating the new DRCF digital markets research with no data" do
+    visit "/drcf-digital-markets-researches/new"
+
+    click_button "Save as draft"
+
+    expect(page.status_code).to eq(422)
+
+    expect(page).to have_css(".elements-error-summary")
+    expect(page).to have_css(".form-group.elements-error")
+    expect(page).to have_css(".elements-error-message")
```
kevindew: I can see this is in some of the other tests but not too sure of the value of it, they just tell us some html elements exists which isn't particularly of value to a user - would suggest removing these lines.

kevindew (commented):
The tests look satisfyingly consistent with the other ones in this repo - nice job.

spec/features/creating_a_drcf_digital_markets_research_spec.rb
baisa: Yes, I have added it, thanks!

spec/fixtures/factories.rb
baisa: Quite right! 👍

app/views/metadata_fields/_drcf_digital_markets_researches.html.erb
baisa: I have added `include_blank: true`, as single select would break the specs, but maybe I am missing something.

kevindew: Sorry, this was my mistake. I thought it still remained a select multiple because it was a select2. I now notice there isn't the `multiple: true` attribute.

Is the precedence in this app to always use a select2 for any select inputs? It seems a bit heavy duty to me given there's only 2 options here and a non-enhanced select would be more appropriate.

baisa: Yes it looks like specialist-publisher uses select2 exclusively.

kevindew: Cool - lets keep that then. Sorry for the misunderstanding

kevindew (approved):

PR #1975, opened 2022-02-15
Fix orphaned labels, add missing H1s and link missing validation messages

davidgisbey:
### Description

In some places labels have not been correctly associated with the corresponding input. This means that they're not read out on screen readers. This PR fixes this issue.

It also uses fixes a few pages not having an H1 and removes some placeholders.

I've extended the existing pattern for error messaging. While it's not optimal, it will be addressed by another card in the backlog.

The build for this is currently failing due to copy changes causing an e2e test to fail. If we decide to keep the copy changes this PR updates the test to use the new copy and skips the test https://github.com/alphagov/publishing-e2e-tests/pull/446 

Once this PR is deployed and merged the test will need to be unskipped 

### Screenshots

While most of the changes aren't visible. I've tweaked the design of versioning section on the edit report page so that error messaging is consistent and it behaves properly for screen readers. Both of the screenshots have triggered a validation for version type not being selected.

|Before|After|
|------------|------------|
|image|image|

### Trello Card

https://trello.com/c/smhLoJnE/246-upgrade-orphaned-labels-specialist

### DIFF
--- a/app/views/shared/_minor_major_update_fields.html.erb
+++ b/app/views/shared/_minor_major_update_fields.html.erb
@@ -1,17 +1,31 @@
 <% unless document.first_draft? %>
-  <div class="checkbox add-vertical-margins">
-    <%= f.radio_button :update_type, :minor, class: 'js-update-type-minor' %>
-    <%= f.label :update_type_minor %>
-    <p class="help-block">Only use for minor changes like fixes to typos, links, GOV.UK style or metadata.</p>
-  </div>
-  <div class="checkbox add-vertical-margins">
-    <%= f.radio_button :update_type, :major, class: 'js-update-type-major' %>
-    <%= f.label :update_type_major %>
-    <p class="help-block">This will notify subscribers to <%= current_format.title.pluralize %>.</p>
-  </div>
-  <div class="<%= document.update_type != 'major' ? 'js-hidden' : nil %> js-change-note">
-    <%= f.label :change_note %>
-    <%= f.text_area :change_note, class: 'form-control short-textarea' %>
-    <p class="help-block">This will be publically viewable on GOV.UK</p>
+  <div class="form-group <%= 'elements-error' if field_has_errors(@document, :update_type) %>">
+    <%= f.label :update_type, for: "#{document.document_type}_update_type_minor" do %>
+      Update type
+      <% field_errors(document, :update_type).each do |msg| %>
+        <br>
+        <span class="elements-error-message add-label-margin"><%= msg.html_safe %></span>
+      <% end %>
+    <% end %>
+
+    <div class="checkbox">
+      <%= f.radio_button :update_type, :minor, class: 'js-update-type-minor', aria: { describedby: "hint-id" } %>
+      <%= f.label :update_type_minor do %>
+        Minor
+      <% end %>
+      <p class="help-block">Only use for minor changes like fixes to typos, links, GOV.UK style or metadata.</p>
+    </div>
+    <div class="checkbox">
+      <%= f.radio_button :update_type, :major, class: 'js-update-type-major', aria: { describedby: "hint-id" } %>
+      <%= f.label :update_type_major do %>
+        Major
+      <% end %>
+      <p class="help-block">This will notify subscribers to <%= current_format.title.pluralize %>.</p>
+    </div>
+    <div class="<%= document.update_type != 'major' ? 'js-hidden' : nil %> js-change-note">
+      <%= f.label :change_note %>
+      <%= f.text_area :change_note, class: 'form-control short-textarea' %>
+      <p class="help-block">This will be publicly viewable on GOV.UK</p>
+    </div>
   </div>
 <% end %>
--- a/app/assets/stylesheets/_forms.scss
+++ b/app/assets/stylesheets/_forms.scss
@@ -97,6 +97,10 @@ form {
   .actions {
     margin: 20px 0;
   }
+
+  .checkbox {
+    margin: 0;
+  }
 }
 
 .summary-length-info {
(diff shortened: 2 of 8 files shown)


app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
+  <div class="form-group <%= 'elements-error' if field_has_errors(@document, :update_type) %>">
+    <%= f.label :update_type, for: 'aaib_report_update_type_minor' do %>
+      Update type
+      <% field_errors(document, :update_type).each do |msg| %>
+        <br>
+        <span class="elements-error-message add-label-margin"><%= msg.html_safe %></span>
+      <% end %>
+    <% end %>
+
+    <div class="checkbox">
```
davidgisbey, line 11: The radios here had the checkbox class applied here for styling purposes. I've left it in for now as I imagine we will probs be moving specialist to the GOV.UK Design System sooner rather than later as there's not a tonne of views in the app. This will be removed as part of that.

Happy to go in and fix stuff like this if people would prefer it.

app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
+      <%= f.radio_button :update_type, :minor, class: 'js-update-type-minor' %>
```
DilwoarH:
```suggestion
      <%= f.radio_button :update_type, :minor, class: 'js-update-type-minor', aria: { describedby: "hint-id" } %>
```
We should probably add [aria-describedby](https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html) here.

DilwoarH: Here is a GOVUK example: https://design-system.service.gov.uk/patterns/question-pages/#asking-multiple-questions-on-a-page

app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
+      <%= f.label :update_type_minor do %>
+        Minor
+      <% end %>
+      <p class="help-block">Only use for minor changes like fixes to typos, links, GOV.UK style or metadata.</p>
+    </div>
+    <div class="checkbox">
+      <%= f.radio_button :update_type, :major, class: 'js-update-type-major' %>
```
DilwoarH: same as [above](https://github.com/alphagov/specialist-publisher/pull/1975/files#r816650749)

app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
+      <%= f.label :update_type_major do %>
+        Major
+      <% end %>
+      <p class="help-block">This will notify subscribers to <%= current_format.title.pluralize %>.</p>
+    </div>
+    <div class="<%= document.update_type != 'major' ? 'js-hidden' : nil %> js-change-note">
+      <%= f.label :change_note %>
+      <%= f.text_area :change_note, class: 'form-control short-textarea' %>
+      <p class="help-block">This will be publically viewable on GOV.UK</p>
```
ollietreend: This is a [spelling mistake](https://dictionary.cambridge.org/spellcheck/english/?q=publically). I realise you didn't write this copy, but might as well fix it when we spot it!

```suggestion
      <p class="help-block">This will be publicly viewable on GOV.UK</p>
```

app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
-    <%= f.radio_button :update_type, :major, class: 'js-update-type-major' %>
-    <%= f.label :update_type_major %>
-    <p class="help-block">This will notify subscribers to <%= current_format.title.pluralize %>.</p>
-  </div>
-  <div class="<%= document.update_type != 'major' ? 'js-hidden' : nil %> js-change-note">
-    <%= f.label :change_note %>
-    <%= f.text_area :change_note, class: 'form-control short-textarea' %>
-    <p class="help-block">This will be publically viewable on GOV.UK</p>
+  <div class="form-group <%= 'elements-error' if field_has_errors(@document, :update_type) %>">
+    <%= f.label :update_type, for: 'aaib_report_update_type_minor' do %>
```
ollietreend: Is `aaib_report_update_type_minor` supposed to be hardcoded in here?

My understanding is that this partial is used across all document types.

app/views/shared/_minor_major_update_fields.html.erb
ollietreend: It's not the best class name, but I agree it's probably not worth fixing at this point. 👍🏻

ollietreend (commented):
This is looking good 👍🏻 just a couple of comments from me

app/views/shared/_minor_major_update_fields.html.erb
davidgisbey: 🤦 absolutely. Good spot

app/views/shared/_minor_major_update_fields.html.erb
davidgisbey: Thanks 🧑‍🎓

ollietreend (approved):
Looks good 👍🏻

app/views/shared/_minor_major_update_fields.html.erb
```diff
@@ -1,17 +1,31 @@
...
+    <%= f.label :update_type, for: "#{document.document_type}_update_type_minor" do %>
+      Update type
+      <% field_errors(document, :update_type).each do |msg| %>
+        <br>
+        <span class="elements-error-message add-label-margin"><%= msg.html_safe %></span>
+      <% end %>
+    <% end %>
+
+    <div class="checkbox">
+      <%= f.radio_button :update_type, :minor, class: 'js-update-type-minor', aria: { describedby: "hint-id" } %>
```
DilwoarH, line 12: hey - the `hint-id` was meant to be the id of the hint block. I'll raise a PR to fix it

PR #2010, opened 2022-04-06
Add short names for product recall finder

kevindew:
When showing results on the finder we were previously showing rather
clunky names such as:

"Product alert type", "Product measure type", "Product recall alert
date"

As it seems these were generated automatically from the "key" value,
rather than defaulting to the "name" value - as one might expect.

This adds the short names that match the filters for a cleaner, more
consistent, user experience.

## Before:

Screenshot 2022-04-06 at 12 49 23

## After:

Screenshot 2022-04-06 at 12 42 27


beccapearce (approved):
Look good to me

PR #2011, opened 2022-04-06
Correct format for product_recall_alert_date

beccapearce:
This field wasn't having any effect because it was specified to have a
format of oim_project - presumably this was a copy and paste error.

This was causing the product_recall_alert_date to not show up in the finder.


spec/features/creating_a_product_safety_alert_report_recall.rb
```diff
@@ -0,0 +1,103 @@
+require "spec_helper"
+
+RSpec.feature "Creating a DRCF digital markets research", type: :feature do
```
kevindew: Ooh copy+paste issue to fix ☝️

kevindew (approved):
Super

PR #2028, opened 2022-05-11
Update puma config

johnny-m-young:
This is needed for when our apps are re-platformed to Kubernetes and are run on puma in production.
A similar change has been made to [Collections in this PR](https://github.com/alphagov/collections/commit/20395d80c0a3dde729cb7243a9b6d7628a99d577)


kevindew:
The code change looks good.

However you've got a rogue .DS_Store file in here.

You can configure your machine to ignore all .DS_Store files (for reference we use .gitignore file in repos to only ignore files created by codebase not those on local dev machines) as an example: https://stackoverflow.com/questions/7335420/global-git-ignore

For me I have a file `~/.gitignore_global` which has contents of:

```
.DS_Store
.byebug_history
```

and then I add that globally with `git config --global core.excludesFile '~/.gitignore_global'`

johnny-m-young:
Ah yes, I didn't see that. Thanks Kevin. I've updated it now to exclude the .DS_Store file and set up a global ignore on my machine.

kevindew (approved):
Super thanks 👍

PR #2032, opened 2022-05-16
Increase RSpec test coverage to >95% for CD

johnny-m-young:
For the ticket on [enabling CD for specialist publisher](https://trello.com/c/NfdjEbQi/261-enable-continuous-deployment-for-specialist-publisher)

This PR contains changes to improve the RSpec test coverage, pushing it up past the 95% threshold required for CD. The following commits are to improve coverage:
- [republish_service.rb specs](https://github.com/alphagov/specialist-publisher/commit/efba05b467089b75e17b1cc011c600ad2794afd3)
- [publishing_api.rake specs](https://github.com/alphagov/specialist-publisher/commit/6005784d27ed75a01731060f6ae8a0c467edca99)
- [parser.rb specs](https://github.com/alphagov/specialist-publisher/commit/5f35d58164775853c9c60f8e3cdbfad08ae59aa2)
- [attachments_controller.rb specs](https://github.com/alphagov/specialist-publisher/commit/1fbd940b43185b4c82812f5cf7bbb642e9238297)

This PR also fixes issues with some RSpec tests: one to avoid altering the system state by [mocking the Rails config](https://github.com/alphagov/specialist-publisher/commit/6e7a7f4f959ef05e452865e77576ee5306a16c51), as this was perpetuating the altered state to subsequent tests, causing them to fail; and one to [stop reloading rake tasks between specs](https://github.com/alphagov/specialist-publisher/commit/02f7e503fd62c3377d5f8842e71f5d1a40eac160), as this was causing issues with fluctuating test coverage reports due to coverage data of rake tasks being wiped with the reload.

Finally, it [consolidates the loading of SimpleCov into one spec helper](https://github.com/alphagov/specialist-publisher/commit/01bbdc172cd4e1daabe9d77645bc37c70f877c09).

### DIFF
--- a/spec/controllers/attachments_controller_spec.rb
+++ b/spec/controllers/attachments_controller_spec.rb
@@ -67,7 +67,7 @@ RSpec.describe AttachmentsController, type: :controller do
 
     it "redirect to the specialist document edit page" do
       document = CmaCase.find(document_content_id, document_locale)
-      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+      allow(subject).to receive(:fetch_document).and_return(document)
 
       stub_any_publishing_api_put_content
       stub_any_publishing_api_patch_links
@@ -75,17 +75,30 @@ RSpec.describe AttachmentsController, type: :controller do
       stub_request(:post, "#{Plek.find('asset-manager')}/assets")
         .to_return(body: JSON.dump(asset_manager_response), status: 201)
 
-      post :create, params: { document_type_slug: document_type_slug, document_content_id_and_locale: "#{document_content_id}:#{document_locale}", attachment: attachment }
+      post :create, params: {
+        document_type_slug: document_type_slug,
+        document_content_id_and_locale: "#{document_content_id}:#{document_locale}",
+        attachment: attachment,
+      }
 
       expect(document.attachments.count).to eq(3)
-      expect(response).to redirect_to(edit_document_path(document_type_slug: document_type_slug, content_id_and_locale: "#{document_content_id}:#{document_locale}"))
+      expect(response).to redirect_to(edit_document_path(
+                                        document_type_slug: document_type_slug,
+                                        content_id_and_locale: "#{document_content_id}:#{document_locale}",
+                                      ))
     end
 
     it "shows an error if no attachment is uploaded" do
       document = CmaCase.find(document_content_id, document_locale)
-      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+      allow(subject).to receive(:fetch_document).and_return(document)
 
-      post :create, params: { document_type_slug: document_type_slug, document_content_id_and_locale: "#{document_content_id}:#{document_locale}", attachment: no_file_attachment }
+      post :create, params: {
+        document_type_slug: document_type_slug,
+        document_content_id_and_locale: "#{document_content_id}:#{document_locale}",
+        attachment: no_file_attachment,
+      }
+
+      expect(flash[:danger]).to be_present
       expect(response).to redirect_to(new_document_attachment_path(document_type_slug: document_type_slug))
     end
   end
@@ -94,9 +107,13 @@ RSpec.describe AttachmentsController, type: :controller do
     it "renders the edit attachment form" do
       document = CmaCase.find(document_content_id, document_locale)
       attachment = document.attachments.find(attachment_content_id)
-      allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+      allow(subject).to receive(:fetch_document).and_return(document)
 
-      get :edit, params: { document_type_slug: document_type_slug, document_content_id_and_locale: "#{document_content_id}:#{document_locale}", attachment_content_id: attachment_content_id }
+      get :edit, params: {
+        document_type_slug: document_type_slug,
+        document_content_id_and_locale: "#{document_content_id}:#{document_locale}",
+        attachment_content_id: attachment_content_id,
+      }
 
       expect(assigns(:attachment)).to eq(attachment)
       expect(response).to render_template :edit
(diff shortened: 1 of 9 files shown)


spec/lib/tasks/publishing_api_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        error_message = %r{#<RuntimeError: Could not find file: lib/documents/schemas/aaib_reportings.json>}
+
+        expect { Rake::Task["publishing_api:publish_finder"].invoke("aaib_reportings") }.to output(error_message).to_stdout
+      end
+    end
+
+    context "correct file name given" do
+      context "no server error present" do
+        it "publishes a single finder to the Publishing API" do
+          content_id = "b7574bba-969f-4c49-855a-ae1586258ff6"
```
davidgisbey: Do we ever expect this to change? I'm assuming not, but if there's a chance it will, we could consider grabbing it from the schema like you did in the above test 

```
schema = Dir.glob("lib/documents/schemas/aaib_reports.json")
content_id = MultiJson.load(File.read(schema))["content_id"]
```

Please feel free to ignore if irrelevant!

spec/lib/tasks/publishing_api_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      stub_publishing_api_has_content(cma_cases, hash_including(document_type: CmaCase.document_type))
+    end
+
+    it "patches links for all CMA cases" do
+      message = %r{Links patched for 10 cma_case documents}
+
+      expect {
+        Rake::Task["publishing_api:patch_document_type_links"].invoke("cma_case")
+      }.to output(message).to_stdout
+    end
```
davidgisbey, line 98: I wonder if we should have an expectation here that it actually hits the `patch_links` endpoint for each case rather than only relying on the count

```
cma_cases.each do |cma_case|
  assert_publishing_api_patch_links(cma_case['content_id'])
end
```
Wdyt?

spec/lib/importers/protected_food_drink_name/parser_spec.rb
```diff
@@ -55,4 +55,95 @@
...
+          "Product type" => "Wine",
+        },
+        { "Protection type" => "Protected Designation of Origin (PDO)" },
+        { "Protection type" => "Traditional Specialities Guaranteed (TSG)" },
+        { "Protection type" => "Traditional Term" },
+        { "Protection type" => "Name protected by international treaty" },
+        { "Protection type" => "American viticultural area" },
+        { "Protection type" => "US spirit drink" },
+      ]
+    end
```
davidgisbey, line 127: tiny nit but newline needed

spec/controllers/attachments_controller_spec.rb
```diff
@@ -96,34 +109,151 @@
...
+        expect(flash[:success]).to eq(success_message)
+        expect(response).to redirect_to(edit_document_path(
+                                          document_type_slug: document_type_slug,
+                                          content_id_and_locale: "#{document_content_id}:#{document_locale}",
+                                        ))
+      end
+
+      context "error updating the attachment" do
+        it "shows an error message and redirects to edit attachment page" do
+          allow_any_instance_of(Document).to receive(:save).and_return(false)
```
davidgisbey: Generally it's better to use `#allow` than `#allow_any_instance_of` where possible as it allows you to mock stub with a lot more specificity 

We can do 

```
 allow(document).to receive(:save).and_return(false)
``` 

here to make sure the correct document is being updated

spec/controllers/attachments_controller_spec.rb
```diff
@@ -96,34 +109,151 @@
...
+        {
+          title: "updated test attachment upload",
+        }
+      end
+
+      it "updates the attachment title" do
+        allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
+        success_message = "Attachment successfully updated"
+
+        post :update, params: {
```
davidgisbey: this isn't anything to do with you, but generally updates use should be PATCH'ing rather than POST'ing. Although, i'm not sure we need to worry about it in this PR as we're just adding coverage.

spec/controllers/attachments_controller_spec.rb
```diff
@@ -96,34 +109,151 @@
...
+        expect(response).to redirect_to(edit_document_path(
+                                          document_type_slug: document_type_slug,
+                                          content_id_and_locale: "#{document_content_id}:#{document_locale}",
+                                        ))
+      end
+
+      context "error updating the attachment" do
+        it "shows an error message and redirects to edit attachment page" do
+          allow_any_instance_of(Document).to receive(:save).and_return(false)
+          allow_any_instance_of(AttachmentsController).to receive(:fetch_document).and_return(document)
```
davidgisbey: ` allow(AttachmentsController).to receive(:fetch_document).and_return(document)`

davidgisbey (approved):
Great work! I've left a few nits, but definitely nothing blocking

spec/controllers/attachments_controller_spec.rb
johnny-m-young: When I change it to this the test breaks. I can however do 

`allow(subject).to receive(:fetch_document).and_return(document)`

shall I change it to this?

spec/lib/tasks/publishing_api_spec.rb
johnny-m-young: Yea I agree with this and will add it to the test

spec/controllers/attachments_controller_spec.rb
johnny-m-young: Ah yes, I only used post as that was used previously and didn't think to check. I can change all of the update tests to patch requests

spec/controllers/attachments_controller_spec.rb
davidgisbey: Tbh i think either is fine. It's a pretty nitty comment. On previous code bases i've worked on (and some of ours) linters flag `allow_any_instance_of` and it fails the build. I personally prefer subject in this case, but i think it's largely preference.

spec/controllers/attachments_controller_spec.rb
davidgisbey: Honestly, it's nothing to do with the scope of the card so i definitely wasn't expecting it to be changed. It's just something i noticed that we could potentially fix at some point. POST/PATCH being used incorrectly isn't the end of the world though.

PR #2073, opened 2022-08-15
Authenticate users against doc types their org can access

leenagupte:
The check on root path checks for users with "editor" permissions checks that the user's organisation can access the first finder in the list of schemas. This schema is AAIB reports and not all users will be able to access them.
Rather than checking against the first document model, check against the first model that the user's org can access.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance]
(https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

### DIFF
--- a/app/controllers/passthrough_controller.rb
+++ b/app/controllers/passthrough_controller.rb
@@ -14,7 +14,7 @@ class PassthroughController < ApplicationController
 private
 
   def first_permitted_format
-    @first_permitted_format ||= document_models.sort_by(&:name).find do |document_class|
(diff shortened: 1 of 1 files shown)


leenagupte:
Closing as it turns out the code is working as expected. If the user's organisation cannot access any of the finder document types, then the user cannot access the application. This is because specialist publisher does not have a home page. The first page you see when you log in is the admin page of the first (alphabetical) document type that you can access. If you can't access any document types, the application has nothing to show you.

PR #2110, opened 2022-09-20
Add disease finder

beccapearce:
Add a new specialist document: animal disease cases

Trello card: https://trello.com/c/bXI3zWdc/584-create-new-specialist-finder-for-animal-disease-cases-and-disease-control-zones

### DIFF
--- a/app/validators/open_before_closed_validator.rb
+++ b/app/validators/open_before_closed_validator.rb
@@ -6,8 +6,8 @@ class OpenBeforeClosedValidator < ActiveModel::EachValidator
 private
 
   def before_closed?(record)
-    opened = record.try(:opened_date) || record.try(:oim_project_opened_date)
-    closed = record.try(:closed_date) || record.try(:oim_project_closed_date)
+    opened = record.try(:opened_date) || record.try(:oim_project_opened_date) || record.try(:disease_case_opened_date)
+    closed = record.try(:closed_date) || record.try(:oim_project_closed_date) || record.try(:disease_case_closed_date)
 
     opened.blank? || closed.blank? || opened <= closed
   end
--- a/spec/fixtures/factories.rb
+++ b/spec/fixtures/factories.rb
@@ -68,6 +68,11 @@ FactoryBot.define do
     organisation_content_id { "957eb4ec-089b-4f71-ba2a-dc69ac8919ea" }
   end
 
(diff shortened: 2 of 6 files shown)


app/validators/open_before_closed_validator.rb
```diff
@@ -6,8 +6,8 @@ def validate_each(record, attribute, _value)
 private
 
   def before_closed?(record)
-    opened = record.try(:opened_date) || record.try(:oim_project_opened_date)
-    closed = record.try(:closed_date) || record.try(:oim_project_closed_date)
+    opened = record.try(:opened_date) || record.try(:oim_project_opened_date) || record.try(:disease_case_opened_date)
+    closed = record.try(:closed_date) || record.try(:oim_project_closed_date) || record.try(:disease_case_closed_date)
```
ollietreend, line 10: Would there be any drawback to re-using `opened_date` and `closed_date` metadata fields on this new document type? 🤔 Or do they need to have a unique name for each document type?

spec/fixtures/factories.rb
```diff
@@ -666,4 +671,21 @@
...
+    document_type { "animal_disease_case" }
+    transient do
+      default_metadata do
+        {
+          "disease_type" => %w[bird-flu],
+          "zone_restriction" => "no-longer-in-force",
+          "zone_type" => %w[surveillance],
+          "virus_strain" => "h5nx",
+          "opened_date" => "2022-08-18T10:12:26+00:00",
+          "closed_date" => "2022-09-18T10:12:26+00:00",
```
ollietreend: Are these field names correct?

```suggestion
          "disease_case_opened_date" => "2022-08-18T10:12:26+00:00",
          "disease_case_closed_date" => "2022-09-18T10:12:26+00:00",
```

ollietreend (commented):
This looks good. Just wanted to ask about the 'opened' and 'closed' date fields.

app/validators/open_before_closed_validator.rb
beccapearce: Yes I thought the same thing (hence the legacy opened and closed dates in the other file 🙈), I think it does need to be different since the public facing page wouldn't load when it was opened and closed dated so I think somewhere it makes it unhappy... which is a shame since all the validations etc would be the same

ollietreend (approved):

PR #2159, opened 2022-11-16
Update intro to tribunal decision finder

richard-gds:
https://govuk.zendesk.com/agent/tickets/5115052
https://trello.com/c/OyOeZNvt/4266-update-to-content-content-change-request

1) England and Wales are having an issue with their database and users can’t currently access copies, apart from some of the most requested ones which are kept on a separate database. There’s currently no expected date for the system to start working again.

2) I’ve updated the link for Scotland, as the responsibility for this has shifted over to a different tribunal building.


ollietreend (approved):
This looks good to me – so long as the content itself has already been approved 👍🏻

PR #2169, opened 2022-11-30
Create new licence document

1pretz1:
This PR is concerned with setting up the basic fields / schema / templates needed to create a licence in Specialist Publisher, there will be follow up tasks, e.g adding the ability for these documents to be rendered in Frontend / adding model validations etc. 

We need to setup the new document type as we're looking to migrate away from the current [Licence Finder](https://www.gov.uk/licence-finder). We'll be running user testing against the new licence finder created by Specialist Publisher.

The test build will fail until https://github.com/alphagov/publishing-api/pull/2211 is merged. 

## Screenshots

Screenshot 2022-12-13 at 23 34 30

Screenshot 2022-12-13 at 23 35 02

Screenshot 2022-12-13 at 19 25 38

Screenshot 2022-12-13 at 19 28 13

Trello:
https://trello.com/c/mGJ5DIsI/1659-create-licence-document-type-in-specialist-publisher

### DIFF
--- a/lib/documents/schemas/licence_transactions.json
+++ b/lib/documents/schemas/licence_transactions.json
@@ -32,6 +32,7 @@
       "type": "text",
       "preposition": "of sector",
       "display_as_result_metadata": true,
+      "show_option_select_filter": true,
       "filterable": true,
       "allowed_values": [
         { "label": "Agriculture, forestry and fishing", "value": "agriculture-forestry-fishing" },
@@ -56,6 +57,7 @@
       "type": "text",
       "preposition": "regarding",
       "display_as_result_metadata": true,
+      "show_option_select_filter": true,
       "filterable": true,
       "allowed_values": [
         {  "label": "Operate as a bookmaker", "value": "operate-as-a-bookmaker" },
--- a/lib/documents/schemas/licence_transactions.json
+++ b/lib/documents/schemas/licence_transactions.json
@@ -1,0 +1,91 @@
+{
+  "content_id": "b8327c0c-a90d-47b6-992b-ea226b4d3306",
+  "base_path": "/find-licences",
+  "format_name": "Find and apply for licences",
+  "name": "Find and apply for licences",
+  "description": "Find licences for business activities and other activities",
+  "filter": {
+    "format": "licence_transaction"
+  },
+  "show_summaries": true,
+  "organisations": ["af07d5a5-df63-4ddc-9383-6a666845ebe9"],
+  "document_noun": "licence",
+  "facets": [
+    {
+      "key": "country",
+      "name": "Country",
+      "type": "text",
+      "preposition": "in country",
+      "display_as_result_metadata": true,
+      "filterable": true,
+      "allowed_values": [
+        { "label": "England", "value": "england" },
+        { "label": "Wales", "value": "wales" },
+        { "label": "Scotland", "value": "scotland" },
+        { "label": "Northern Ireland", "value": "northern-ireland" }
+      ]
+    },
+    {
+      "key": "sector",
+      "name": "Sector",
+      "type": "text",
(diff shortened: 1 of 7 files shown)


spec/features/licence_transaction_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    fill_in "Summary", with: "This is the summary of a licence"
+    fill_in "Body", with: "## Header#{"\n\nThis is the long body of a licence" * 2}"
+    select "England", from: "Country"
+    select "Education", from: "Sector"
+    select "Use CCTV systems", from: "Activity"
+    fill_in "Continuation link", with: "https://www.gov.uk"
+    fill_in "Will continue on", with: "on GOV.UK"
+
+    click_button "Save as draft"
+    assert_publishing_api_put_content(content_id)
```
baisa, line 52: Maybe consider passing payload value here?

baisa (commented):
Looks good @1pretz1 ! Happy to approve when this is ready to merge 👍  (We just need to remember to set primary_publishing_organisation and remove pre production flag).

spec/features/licence_transaction_spec.rb
1pretz1: Nice suggestion, I did ponder about testing the payload (like some other features) but decided against it for a couple of reasons:

1) It is a bit strange to be testing API calls / payload in a feature test, as the expectations should just be what appears on the page versus what the underlying code is doing (api calls, model updates etc). Feels like asserting a status and a generic call is kind of okay, but asserting a full payload seems too much.

2) The payload / data structure is already tested in other places, e.g the [model is validated against Publishing API's schema]( https://github.com/alphagov/specialist-publisher/pull/2169/files) and the [shared payload presenter has a spec]( https://github.com/alphagov/specialist-publisher/blob/main/spec/presenters/document_presenter_spec.rb).

I've also tested various scenarios against the feature test, like removing a model attribute, updating a field name etc which all results in the feature test failing so pretty confident we don't need to assert the payload too.

What are your thoughts?

1pretz1:
Thanks for the review @baisa! I believe I've addressed your comment, feel free to disagree though! 

True, we'll need to update the `primary_publishing_organisation` and remove pre production flag if / when the finder goes officially live. However, for User Research, we'll be keeping the primary org as GDS and setting the pre-production flag to ensure the finder doesn't get published after we merge the PR.

lib/documents/schemas/licence_transactions.json
```diff
@@ -0,0 +1,94 @@
...
+  "base_path": "/find-licences",
+  "format_name": "Find and apply for licences",
+  "name": "Find and apply for licences",
+  "description": "Find licences for business activities and other activities",
+  "filter": {
+    "format": "licence_transaction"
+  },
+  "show_summaries": true,
+  "organisations": ["af07d5a5-df63-4ddc-9383-6a666845ebe9"],
+  "document_noun": "licence",
```
baisa, line 13: Just a note regarding how results order is determined - we can customise it by adding for example `"default_order": "title",` , see more examples here https://github.com/alphagov/specialist-publisher/search?q=default_order

spec/features/licence_transaction_spec.rb
baisa: Sounds reasonable 👍

baisa (approved):
Ok sure @1pretz1 , we can merge with pre production flag if we want this deployed for User Research 👍

lib/documents/schemas/licence_transactions.json
1pretz1: Ah interesting, didn't know that, thanks! I can't see an obvious ordering to apply so I'll leave this for now but will keep this in mind if there is a user need.

PR #2184, opened 2022-12-15
Set license_transaction rendering app as frontend

edwardkerry:
This PR sets the rendering app for the LicenseTransaction to"frontend". It removes the default hard-coded rendering_app from the DocumentPresenter, and adds a method to the parent Document class to default the rendering_app to "government-frontend.

[trello](https://trello.com/c/iPGhPygc/1675-use-frontend-as-the-rendering-app-for-licences-in-specialist-publisher)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -116,6 +116,10 @@ class Document
     truncate(@base_path, length: 250, omission: "")
   end
 
+  def rendering_app
+    "government-frontend"
+  end
+
   def document_type
     self.class.document_type
   end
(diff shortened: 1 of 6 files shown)


leenagupte (approved):
👍

1pretz1 (approved):

1pretz1:
This could be merged but depends on https://trello.com/c/AsWZLDoG/1652-update-frontend-to-render-new-licence-document-type to actually render the licence pages, so we'll hold fire until we pick that work up again.

PR #2197, opened 2023-01-03
Update the content schemas rake task to build_schemas

beccapearce:
The name of the rake task to rebuild the content-schemas has been changed from build to build_schemas when it was moved out of the govuk-content-schemas repo and into the publishing repo


davidgisbey (approved):
Might be a tiny nit so feel free to ignore. Maybe we should link to an example in the PublishingAPI now so it's clear it lives there now. Feel free to ignore!

beccapearce:
Yes I agree but I'm not sure we have an example to link to just yet 🤔, I started making the changes but couldn't see links to replace all the links we had

PR #2207, opened 2023-01-10
Import level two licence sectors into Specialist Publisher

leenagupte:
Trello: https://trello.com/c/TiYLM2hE

# What's changed and why?

It has been agreed with Content Designers and User Researchers to tag migrated licences in Specialist Publisher to only level two sectors (from Licence Finder).

The hierarchical structure for sectors (3 levels) is being dropped and licence documents will just be tagged to the 2nd level. This is due to limitations to support hierarchical tagging in Specialist Publisher / Finder Frontend and to mitigate bespoke work just for User Research.

Activities do not need to be imported as activities don't make complete sense. They mostly map 1-2-1 with individual licences but have different names to the licences.

Country tagging will remain and is already configured in Specialist Publisher.

The tagging of licence documents to the level sectors will be accomplished in a future data migration.

### DIFF
--- a/lib/data/licence_transaction/industry_sectors.txt
+++ b/lib/data/licence_transaction/industry_sectors.txt
@@ -1,0 +1,75 @@
+Accommodation                                                                    
+Adult and other education                                                      
+Advertising and marketing services                                               
+Agricultural materials wholesale                                                 
+Agriculture                                                                      
+Architectural and other building services                                        
+Arts and entertainment                                                           
+Arts, antiques and second-hand goods retail                                      
+Arts, sports and recreation goods manufacturing                                  
+Arts, sports and recreation goods wholesale                                      
+Bars                                                                             
+Books, stationery and gifts retail                                               
+Books, stationery and gifts wholesale                                            
+Building construction                                                            
+Building materials and hardware wholesale
+Building materials manufacturing
+Building trades
+Business support services
+Catering
+Chemical and petroleum products manufacturing
+Clothing, footwear and accessories manufacturing
+Clothing, footwear and accessories retail, hire and repair
+Clothing, footwear and accessories wholesale
+Complementary therapy services
+Computer services
+Cosmetics, medicinal and chemical products wholesale
+Design and photographic services
+Electric and electronic appliance retail, hire and repair
+Electric and electronic equipment wholesale
+Electrical equipment manufacturing
+Electronic equipment and computer manufacturing
+Energy utilities
+Film production and distribution
+Financial services
+Fishing and hunting
+Food, drink and tobacco manufacturing
+Food, drink and tobacco retail
+Food, drink and tobacco wholesale
+Forestry
+Formal education
+Gambling facilities operation
+Health services
+Home and garden products retail, hire and repair
+Internet services
+Jewellery, watch and clock manufacturing
+Jewellery, watch and clock retail and repair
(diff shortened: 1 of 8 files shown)


1pretz1 (approved):
Looks good!

app/services/licence_transaction/import_industry_sectors.rb
```diff
@@ -4,6 +4,10 @@ def self.call
   end
```
1pretz1: Minor: we should use encapsulation for the private methods

spec/services/licence_transaction/import_industry_sectors_spec.rb
```diff
@@ -1,21 +1,21 @@
 require "spec_helper"
 
 RSpec.describe LicenceTransaction::ImportIndustrySectors do
```
1pretz1: I think we're now testing the wrong method (should be `call`) and as a result we're missing some tests for the features that have been added (reading / writing to JSON file). If we find it useful to have `imported_json_data` as its own public method then we should probably test both.

1pretz1 (commented):
Changes look good, nice automation! I've just made a couple of comments. We could also create a rake task for this import which is a bit nicer than having to run commands in a console.

spec/services/licence_transaction/import_industry_sectors_spec.rb
leenagupte: You caught me! I'll have a bit of refactor and find a nice way to test the JSON file.

leenagupte:
@1pretz1 Thanks for your review. I've addressed your comments and fixed up.

app/services/licence_transaction/import_industry_sectors.rb
```diff
@@ -0,0 +1,43 @@
+class LicenceTransaction::ImportIndustrySectors
```
1pretz1: This feels like it should live in lib as it's only used by a single rake task and not part of the main application code

1pretz1 (approved):
Nice work! Approving as my only comment is minor.

app/services/licence_transaction/import_industry_sectors.rb
leenagupte: Hmmm, I thought about this, and figured it was really an import service with the potential to be run a few times. Maybe even refactored to be used in the scripts to update the sector names and re-tag licences. 

The name is probably wrong though. I think it should have been named "IndustrySectorsImporter"

1pretz1: I still think it should live in `lib` unless it is being used by the application at runtime as everything else in the services directory is related to the main application code. Interestingly, we do have an `importers` directory in `lib`: https://github.com/alphagov/specialist-publisher/tree/589ad0fee62548e86e979aec74841044e841f959/lib.

Agreed on the proposed name 👍

leenagupte: Agreed on name and location. I'll move the class and test files into lib. I noticed the same "importer" folder when I was reviewing your PR! https://github.com/alphagov/specialist-publisher/pull/2210#discussion_r1072103633

leenagupte:
@1pretz1 I've renamed the files and moved everything to under "lib" as agreed. The changes are ready for a re-review.

1pretz1 (approved):
Nice!

PR #2210, opened 2023-01-12
Create migration to import licences from Publisher 

1pretz1:
More info in the commits ->

Successful run in Staging: https://deploy.blue.staging.govuk.digital/job/run-rake-task/218466/console

Running the script again (after already importing licences): https://deploy.blue.staging.govuk.digital/job/run-rake-task/218468/console

### Finder Frontend

Screenshot 2023-01-13 at 11 28 49

### Specialist Publisher

Screenshot 2023-01-13 at 11 30 50

Screenshot 2023-01-13 at 11 30 39

Trello:
https://trello.com/c/RvKrQhgS/1676-create-import-mechanism-to-migrate-publisher-licences-into-specialist-publisher


lib/common_licence_identifiers.rb
```diff
@@ -0,0 +1,240 @@
+# Mainstream Publisher (411 licence)
+# arr = Edition.where(state: "published", _type: "LicenceEdition”).pluck(:licence_identifier)
+
+# Licence Finder (228 licence that exist in Publisher)
+# irb(main):017:0> Licence.where(:gds_id.in => arr).pluck(:gds_id)
+
+class CommonLicenceIdentifiers
+  def self.all
```
leenagupte: Why will this list be needed? Can you added something to the commit message? For example, I assume it's because we're only importing licences that are in both Publisher and Licence Finder?

Is that because there is only sector information for the licences that are in both?

Are there any licences that are in Publisher and have a licence-identifier, but are not in Licence Finder? If so, what is the plan for these licences? Can this information be added to the commit message.

leenagupte: I wonder if we should agree an approach for where to store licence data like this. For example, I created a [data file](https://github.com/alphagov/specialist-publisher/pull/2207/commits/af9908a65752fa3b87d74213190ad41cbb25f450), this is a list. I think we should pick one or the other (or a third option) so all of the data is in one place.

app/models/licence_transaction.rb
```diff
@@ -17,6 +17,10 @@ def taxons
     []
   end
 
+  def change_note_required?
+    false
```
leenagupte: I'm a bit concerned that we'll forget to remove this and cause problems with not having change notes in the future.

Could a default change note be set instead? i.e. something like "Imported from Publisher"?

leenagupte: I know that change notes are not being displayed on licences at the moment, but it could be useful to keep them for email alerts. This would also match the other specialist document types that do record them.

lib/licence_importer.rb
```diff
@@ -0,0 +1,81 @@
...
+  def initialize(filter_by_licence_identifiers:)
+    @filter_by_licence_identifiers = filter_by_licence_identifiers
+  end
+
+  def import
+    licences = Services.publishing_api.get_content_items(
+      document_type: "licence", page: 1, per_page: 500, states: "published",
+    )["results"]
+
+    migratable_licences = filter_by_licence_identifiers ? filtered_licences(licences) : licences
```
leenagupte: So `filter_by_licence_identifiers` expects a boolean to be passed in to the constructor? So if `filter_by_licence_identifiers` is true, only import the licences that exist in both Publisher and Licence Finder, but if it's false it import all licences?

What are the scenarios for each option? For example, when would `filter_by_licence_identifiers` be set to false? Can a note be added to the commit message to explain the difference?

lib/licence_importer.rb
```diff
@@ -0,0 +1,81 @@
...
+        licence_transaction_continuation_link: details["continuation_link"],
+        licence_transaction_licence_identifier: details["licence_identifier"],
+        imported: true,
+      )
+
+      if new_licence.valid?
+        original_content_id = licence["content_id"]
+        new_content_id = new_licence.content_id
+
+        save_draft(original_content_id, new_licence)
```
leenagupte: I don't think `original content_id` needs to be passed here as I don't believe it's being used in the `save_draft` method.

lib/licence_importer.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def filtered_licences(licences)
+    common_licences = CommonLicenceIdentifiers.all
+
+    licences.select do |licence|
+      common_licences.include?(licence["details"]["licence_identifier"])
+    end
+  end
+
+  def save_draft(_original_content_id, new_licence)
```
leenagupte: Related to one of the comments above, is `_original_content_id` needed?

lib/licence_importer.rb
leenagupte: The document model already has a `save` method that already calls `put_content` and `patch_links` in publishing_api. So `new_licence.save` could have been called instead.

I'm assuming `save` wasn't used because it calls `DocumentLinksPresenter` and that only presents organisations, primary_publishing_organisation and taxons as links to publishing-api?

Could a note about this be added to the commit message?

lib/licence_importer.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      if new_licence.valid?
+        original_content_id = licence["content_id"]
+        new_content_id = new_licence.content_id
+
+        save_draft(original_content_id, new_licence)
+
+        save_links(original_content_id, new_content_id)
+
+        publish(new_content_id)
```
leenagupte: The trello card linked to the PR description says to only create draft documents. Has that changed? Is that why "publish" is being called here?

lib/licence_importer.rb
```diff
@@ -0,0 +1,81 @@
+require "common_licence_identifiers"
+
+class LicenceImporter
```
leenagupte: I think this class should be namespaced with "LicenceTransaction" to separate it from the other generic classes. ie. `LicenceTransaction::LicenceImporter`

I've also just seen the "importer" directory within "lib" so I think this class would fit there.

I think the `LicenceTransaction::ImportIndustrySectors` being created in PR #2207 should/will also be moved there so that all of the import code is in one place.

leenagupte: I think it would have helped with the review if the creation of the class had been broken into smaller commits so that some of the decisions could have been more easily explained and understood.

lib/tasks/import_licences.rake
```diff
@@ -0,0 +1,10 @@
+require "licence_importer"
+
+# Imports all published licences or you can filter by licence identifier stored
+# in `../common_licence_identifiers.rb`.
+
+desc "imports licences"
```
leenagupte: I think this rake task should be namespaced, so that we can add all of the related licence transaction import rake tasks to the same file.

I've gone with a `licence_transaction` namespace in the import sectors PR and called the file `licence_transactions.rake`. Happy to discuss and agree on names.

lib/tasks/import_licences.rake
```diff
@@ -0,0 +1,10 @@
...
+task :import_licences, [:filter_by_licence_identifiers] => :environment do |_, args|
+  filter_by_licence_identifiers = args[:filter_by_licence_identifiers] || false
+  LicenceImporter.new(filter_by_licence_identifiers:).import
```
edwardkerry: is this missing the filter_by_licence_identifiers variable?
`LicenceImporter.new(filter_by_licence_identifiers: filter_by_licence_identifiers).import`

spec/lib/licence_importer_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+
+  def publising_api_patch_links_response
+    {
+      organisations: %w[af07d5a5-df63-4ddc-9383-6a666845ebe9],
+      primary_publishing_organisation: %w[af07d5a5-df63-4ddc-9383-6a666845ebe9],
+      taxons: [],
+    }
+  end
+
+  def publishing_api_licences_response
```
leenagupte: If the actual values in the content item don't matter, it's possible to use GovukSchemas to load an example content item. Example [here](https://github.com/alphagov/finder-frontend/blob/e0e8f32000f708b2b2a67983d49f45d9f04c27f8/spec/support/content_helper.rb#L14).

spec/lib/licence_importer_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+    stub_publishing_api_has_content(
+      publishing_api_response,
+      { document_type: "licence", page: 1, per_page: 500, states: "published" },
+    )
+    stub_publishing_api_has_links(
+      { content_id: "6ad67774-62aa-4413-b9eb-7f3aaa897115", links: publising_api_patch_links_response },
+    )
+  end
+
+  context "when a licence is valid" do
```
leenagupte: All of the tests you've added a great, and it feels like they cover all of the scenarios. I think they could be re-jigged a bit to make it clear what's being tested. For example, I think the tests here are only for when `filter_by_licence_identifiers` is false? I think the test names should be updated to reflect the specific scenario being tested.

Perhaps rather than dividing the tests by validity, they could be grouped by the `filter_by_licence_identifiers` value?

spec/lib/licence_importer_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+      )
+      patch_links_request = stub_publishing_api_patch_links(
+        new_content_id, { links: expected_patch_links_payload }
+      )
+
+      described_class.new(filter_by_licence_identifiers: false).import
+
+      expect(put_content_request).to have_been_requested
+      expect(patch_links_request).to have_been_requested
+      expect(publish_request).to have_been_requested
```
leenagupte: As well as expecting the publishing-api to be called, I think it would be good to check the number of Licences is increasing by the expected amount. 

Something like: 
- https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/change-matcher 
- https://github.com/alphagov/content-publisher/blob/d1263f1867c79bc0392cb5cda6868dcbd06ae8bb/spec/jobs/scheduled_publishing_job_spec.rb#L20-L24

app/models/document.rb
```diff
@@ -7,7 +7,7 @@ class Document
   include DateHelper
   include PublishingHelper
 
-  attr_reader :update_type
+  attr_reader :update_type, :imported
```
leenagupte, line 10: Could this status be used to make sure that the licence isn't imported twice, for example if the script is run a second time? This could happen if the script fails the first time, and has to be re-run.

leenagupte (commented):
This is a really good start. Good job figuring out that the licence data can be pulled in from publishing-api!

I've got some inline comments and questions. 

Also I think there needs to be some agreement offline about how importer code should be structured to make everything consistent and easy to find and remove later.

Finally, I think we need to think about how to prevent Licence Transactions being duplicated in Publishing API if the script is run multiple times.

lib/tasks/import_licences.rake
leenagupte: No, the Ruby 3 styles allow you to use `(filter_by_licence_identifiers:)` rather than `(filter_by_licence_identifiers: filter_by_licence_identifiers)` if the key and value have the same name. [1]
I think Rubocop enforces this style.


[1]: https://blog.saeloun.com/2021/09/28/ruby-allow-value-omission-in-hash-literals.html

edwardkerry: Ah, not seen that previously (or not often enough to remember it - I will now). Thanks  Leena.

lib/common_licence_identifiers.rb
1pretz1: Have added more information to the commit message to cover the questions above, in summary:

Yes for your first two questions.

> Are there any licences that are in Publisher and have a licence-identifier, but are not in Licence Finder? If so, what is the plan for these licences? 

Yes, these will be imported in a later piece of work, at this stage there is no mapped out plan as we're only at the User Research stage. Once we know the product we're building then we can plan the work to import the other licences.  

> I wonder if we should agree an approach for where to store licence data like this. For example, I created a [data file](https://github.com/alphagov/specialist-publisher/pull/2207/commits/af9908a65752fa3b87d74213190ad41cbb25f450), this is a list. I think we should pick one or the other (or a third option) so all of the data is in one place.

We discussed async and agreed to format the licence identifier data as `txt` file for consistency.

app/models/licence_transaction.rb
1pretz1: Sure, have added a change note 👍 

> Could a default change note be set instead? i.e. something like "Imported from Publisher"?

Hmm, change notes are meant to represent a public facing change, rather than documenting an internal migration so I was hesitant to to add that sort of note, however it doesn't matter at this stage so I have added that change note... for now and we can discuss further at a later date 😄 

> I know that change notes are not being displayed on licences at the moment, but it could be useful to keep them for email alerts.

We don't currently support email alerts for licences unless we choose to enable them on the finder but fair point.

lib/licence_importer.rb
1pretz1: I initially wrote this so we could easily import all published licences (check if any errors etc), however this is not needed at this stage so have removed the functionality, it's easy to add back so no biggie.

lib/licence_importer.rb
1pretz1: Ah good point, I missed the unused argument. Have now removed.

Yep, save wasn't used as it limited the links we could save (hardcoded values). I've added more information to the commit message.

lib/licence_importer.rb
1pretz1: Nope, good spot!

lib/licence_importer.rb
1pretz1: The Trello card is probably out of date, we need to publish the documents so they appear in the finder (don't seem to support draft finders). Have added this information to the commit message.

lib/licence_importer.rb
1pretz1: > I think this class should be namespaced with "LicenceTransaction" to separate it from the other generic classes. ie. LicenceTransaction::LicenceImporter
I've also just seen the "importer" directory within "lib" so I think this class would fit there.
I think the LicenceTransaction::ImportIndustrySectors being created in PR https://github.com/alphagov/specialist-publisher/pull/2207 should/will also be moved there so that all of the import code is in one place.

Discussed async and have moved to namespaced directory.

> I think it would have helped with the review if the creation of the class had been broken into smaller commits so that some of the decisions could have been more easily explained and understood.

Noted, I've added more information to the commit message to explain the 'why'.

lib/tasks/import_licences.rake
1pretz1: Discussed async and done!

spec/lib/licence_importer_spec.rb
1pretz1: I did consider this but I didn't trust the example data to be up to date and we get back slightly different data with the APIs than what would be found in the schema example. It's probably overkill but it feels safe / realistic to have the actual API responses.

spec/lib/licence_importer_spec.rb
1pretz1: I've removed `filter_by_licence_identifiers` option, as this is the only way of importing licences so don't believe any changes need to be made.

spec/lib/licence_importer_spec.rb
1pretz1: Hmm, don't think we can do that as Specialist Publisher doesn't have a database and so we're unable to query   the changed state. I think having the request expectations is suffice as it ensures we're calling the Publishing API for the intended requests 👍

app/models/document.rb
1pretz1: That would be good to have but would need extra work implementing, as this flag would need to be persisted on the content item. We could potentially use the `base_path` or other unique identifier to work out if the document has been imported before (require an extra request to Publishing API) and skip if so, however we'd also need to consider the scenario where the content has changed, in which case we'd either need to skip, update the new content on the existing licence or unpublish the old one first. 

~Definitely something to consider but I think this is out of scope for this initial piece of work. What do you think?~

Thinking a bit more, as we don't handle if Publishing API is unavailable midway through, it may be a pain if the tasks fails before UR. I think the document won't publish anyway if they already exist in the stack as there will be a conflict of base_path's.

I'll look into some validation for this now 👍

1pretz1:
Hey @leenagupte , thanks for the review! I believe I've addressed all your comments.

The PR will need updating (only the spec) if https://github.com/alphagov/specialist-publisher/pull/2216 is merged before this.

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -0,0 +1,83 @@
...
+      def call
+        filtered_licences.each do |licence|
+          details = licence["details"]
+          new_licence = ::LicenceTransaction.new(
+            locale: "en",
+            base_path: "/find-licences#{licence['base_path']}",
+            title: licence["title"],
+            summary: licence["description"],
+            body: details["licence_overview"].first["content"],
+            publication_state: "draft",
```
leenagupte, line 13: Minor, but shouldn't this match the publication state from publishing-api? Especially as only published licences are being retrieved and the licence transactions are being published!

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -0,0 +1,83 @@
...
+
+          new_content_id = new_licence.content_id
+
+          save_draft(new_licence)
+
+          save_links(licence["content_id"], new_content_id)
+
+          publish(new_content_id)
+
+          puts "Published: #{new_licence.base_path}"
```
leenagupte, line 47: Not a blocker, but I'm surprised the linter hasn't picked this up. It's definitely something that Rubocop warns about: https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/Output

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+require "importers/licence_transaction/licence_importer"
+
+RSpec.describe Importers::LicenceTransaction::LicenceImporter do
+  let(:new_content_id) { "0cc89dd8-1055-4e6b-8f64-9a772dbe28db" }
+
+  before do
+    allow(SecureRandom).to receive(:uuid).and_return(new_content_id)
+
+    stub_publishing_api_has_content(
+      publishing_api_response,
```
leenagupte, line 13: Minor, it's feels a little strange that `publishing_api_response` is defined in every test but used in the global stub. I initially looked for a method similar to `publishing_api_licences_response` and got a little confused when I couldn't find it.

It might be easier to read if a default value was set here, i.e. 
```
let(:publishing_api_response) { publishing_api_licences_response }
```

This value could be modified directly in the tests as needed.

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -31,6 +31,11 @@ def call
             next
           end
 
+          if new_licence_already_imported?(new_licence.base_path)
+            puts "Skipping as licence: #{new_licence.base_path} is already imported"
+            next
+          end
+
```
leenagupte, line 38: This is nice!

Very minor, but I think there's a typo in the commit message. I think this should say option 3:

> I also didn't choose option 2 as this would require multiple editions to be
published / change history that we don't want or need, as it would be
better to have a fresh edition on import.

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -182,4 +210,62 @@ def publishing_api_licences_response
       },
     ]
   end
+
+  def publishing_api_existing_licences_response
+    [
+      {
+        "auth_bypass_ids" => [],
+        "base_path" => "/find-licences/art-therapist-registration",
```
leenagupte, line 220: Nice! I see what you did there 🎉

spec/lib/licence_importer_spec.rb
leenagupte: As the test is only checking one licence, I think this is fine.

leenagupte (commented):
This is looking great! 🎉 
Thanks for updating the commit messages, the decisions are really clear now.
There is one typo in a commit that I've mentioned in the inline comments that I think really needs to be fixed as it affects the meaning of the point you're making.

I love the new functionality for skipping imports too. 

There are a couple of inline comments, but they are all minor. 
Well done!

lib/importers/licence_transaction/licence_importer.rb
1pretz1: No as we need to create a draft first before we can publish, as we can't just publish a newly created document without their being a draft (put_content and then publish). The results is only one published edition which is the correct state.

spec/lib/importers/licence_transaction/licence_importer_spec.rb
1pretz1: Hmm we want to have a bespoke value for each and not repeat the stub. If we modified the value outside of a let block then we'd need to repeat the stub. I've moved the `let` to have a default value at the global level should which should make it clearer?

lib/importers/licence_transaction/licence_importer.rb
1pretz1: Good spot!

lib/importers/licence_transaction/licence_importer.rb
1pretz1: I did have a little play but couldn't get the correct logging behaviour so leaving as is. This is only rule I find quite annoying / adds little value in these sort of scenarios. We do use `puts` in various other places in the app too.

leenagupte: I got caught by the linter in the licence-finder PR and ended up having to use `Rails.logger.info`. In development it output to the log file. But yes, it's very irritating.

leenagupte (approved):
Fantastic work! 🏅

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+  def expected_put_content_payload
+    {
+      base_path: "/find-licences/art-therapist-registration",
+      title: "Art therapist registration",
+      description: "You need to register with the Health and Care Professions Council (HCPC) to practise as an art therapist in the UK",
+      document_type: "licence_transaction",
+      change_note: "Imported from Publisher",
+      schema_name: "specialist_document",
+      publishing_app: "specialist-publisher",
+      rendering_app: "government-frontend",
```
leenagupte: Should this now be "frontend" now that https://github.com/alphagov/specialist-publisher/pull/2184 has been merged?

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -0,0 +1,189 @@
...
+            text: "Fines and penalties",
+            level: 2,
+            id: "fines-and-penalties",
+          },
+        ],
+      },
+      routes: [
+        {
+          path: "/find-licences/art-therapist-registration",
+          type: "exact",
```
leenagupte: Ditto here, I think this should be "prefix" now.

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -186,4 +212,62 @@ def publishing_api_licences_response
...
+        "public_updated_at" => "2023-01-18T17:28:50Z",
+        "published_at" => "2023-01-18T17:43:46Z",
+        "publishing_app" => "specialist-publisher",
+        "publishing_api_first_published_at" => "2023-01-18T17:28:50Z",
+        "publishing_api_last_edited_at" => "2023-01-18T17:42:02Z",
+        "redirects" => [],
+        "rendering_app" => "government-frontend",
+        "routes" => [
+          {
+            "path" => "/find-licences/art-therapist-registration", "type" => "exact"
```
leenagupte: Should the type here be "prefix" as well?

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -186,4 +212,62 @@ def publishing_api_licences_response
...
+        "first_published_at" => "2023-01-18T17:28:50Z",
+        "last_edited_at" => "2023-01-18T17:42:02Z",
+        "phase" => "live",
+        "public_updated_at" => "2023-01-18T17:28:50Z",
+        "published_at" => "2023-01-18T17:43:46Z",
+        "publishing_app" => "specialist-publisher",
+        "publishing_api_first_published_at" => "2023-01-18T17:28:50Z",
+        "publishing_api_last_edited_at" => "2023-01-18T17:42:02Z",
+        "redirects" => [],
+        "rendering_app" => "government-frontend",
```
leenagupte: Should the rendering app be "frontend" now?

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -116,7 +116,7 @@ def expected_put_content_payload
       change_note: "Imported from Publisher",
       schema_name: "specialist_document",
       publishing_app: "specialist-publisher",
-      rendering_app: "government-frontend",
+      rendering_app: "frontend",
```
leenagupte, line 119: I think all the changes in this commit should be fixed up to the original commit.

leenagupte (commented):
The changes to the rendering app and route type are good, but I think they should be fixed up to the original commit.

leenagupte (approved):

PR #2212, opened 2023-01-15
Dockerfile cleanup + enable bootsnap.

sengi:
- Use bootsnap to reduce startup resource usage spike.
- Parameterise Ruby version.
- Make better use of WORKDIR.
- Use env vars from the base image where appropriate, instead of hardcoding paths.
- Remove unnecessary use of `bundle exec`.
- Update .dockerignore.

Tested: app boots successfully with `docker run`.


Dockerfile
```diff
@@ -1,31 +1,28 @@
...
+ENV GOVUK_NOTIFY_API_KEY=unused
 
+WORKDIR $APP_HOME
+COPY Gemfile* .ruby-version ./
 RUN bundle install
-
-COPY . /app
-
-# This is not a valid key but it needs to be in the correct format for the ruby client
-ENV GOVUK_NOTIFY_API_KEY=test-b56ea330-006a-459b-8af3-8015bb51a0e7-77b95465-3c02-4b09-9371-44e33cbdf9c6
```
theseanything, line 15: is this comment still valid? Does the dummy value need be a specific format still?

sengi: Oh good question - forgot to mention that I'd tested this. Works just fine with the shorter (and less weirdly-specific-looking) dummy value.

marcpomfret (approved):

PR #2216, opened 2023-01-19
Present LicenceTransaction route type as 'prefix'

edwardkerry:
Licences previously published by LicencePublisher used 'prefix' rather than 'exact' routes.

A 'prefix' route tells Router to direct all paths that start with a certain pattern to a certain application. A prefixed route for e.g '/find-licences/new-licence' would direct all request starting with that path, e.g 'licences/new-licence/council' to the same application as '/licences/new-licence'.

The full list of routes for a licence page are dynamically generated as they rely on postcode lookups, so we do not know the full list of possible routes.

This commit allows LicenceTransactions to present as 'prefix' routes, whilst maintaining other documents to maintain 'exact' routing.

[trello](https://trello.com/c/51k4iW5A/1677-update-specialist-publisher-licence-to-use-prefix-routes)

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -341,6 +341,10 @@ class Document
     }
   end
 
+  def route_type
+    "exact"
+  end
+
   def self.exportable?
     false
   end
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
Looks good to me.

PR #2223, opened 2023-01-26
Reorder licence transaction facet so Industry is first

edwardkerry:
This PR swaps the order of facets for Licences, so Industry appears before Location in the rendering app.

before
Screenshot 2023-01-26 at 14 18 12

after
Screenshot 2023-01-26 at 16 33 23

It also updates the order in the admin UI for parity.

[trello](https://trello.com/c/lkm9322o/1786-move-industry-filter-above-location-for-new-licence-finder)


leenagupte (approved):
👍

1pretz1 (approved):

PR #2262, opened 2023-04-18
Editor content tweaks

KludgeKML:
Adds content tweaks mentioned here https://docs.google.com/document/d/1b64gBYFuuSGBPZ8B4Poy0UH57NKuLDVBn9WkLi7dANk/edit#heading=h.jz5m30507l7y to the editor.

These changes affect all document editor pages in Specialist publisher.

- Alter New document type titles so they'll be downcased (except initials, and special case for EU Withdrawal Act 2018)

Before:

Screenshot 2023-04-20 at 09 38 24

After:
Screenshot 2023-04-18 at 15 13 49

https://trello.com/c/BWE1XxzY/1937-backend-devs-review-effort-of-suggested-name-changes-to-backend-of-specialist-publisher-tool

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -308,6 +308,11 @@ class Document
     title.parameterize.pluralize
   end
 
+  # Singularize and downcase but maintaining initials
+  def self.title_for_new_document
+    title.singularize.split(" ").map { |p| p.upcase == p ? p : p.downcase }.join(" ")
+  end
+
   def content_id_and_locale
     "#{content_id}:#{locale}"
   end
--- a/app/views/documents/new.html.erb
+++ b/app/views/documents/new.html.erb
@@ -1,4 +1,4 @@
-<%= content_for :page_title, "New #{current_format.title.singularize}" %>
+<%= content_for :page_title, "New #{current_format.title_for_new_document}" %>
 
 <% content_for :breadcrumbs do %>
   <li><%= link_to "Your documents", documents_path(current_format.slug) %></li>
@@ -9,7 +9,7 @@
   GOVUK.formChangeProtection.init($('.new_document'), 'You have unsaved changes that will be lost if you leave this page.');
 <% end -%>
 
-<h1>New <%= current_format.title.singularize %></h1>
+<h1>New <%= current_format.title_for_new_document %></h1>
 
 <div class="row">
   <div class="col-md-8">
--- a/app/models/statutory_instrument.rb
+++ b/app/models/statutory_instrument.rb
@@ -32,6 +32,11 @@ class StatutoryInstrument < Document
     "EU Withdrawal Act 2018 statutory instrument"
   end
 
+  # Keep name of act titleized
+  def self.title_for_new_document
+    title
+  end
+
   def links
     super.merge(
       organisations: organisations | [primary_publishing_organisation],
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -8,7 +8,7 @@
 
 <div class="row">
   <div class="sidebar col-md-3">
-    <%= link_to "Add another #{current_format.title}", new_document_path(current_format.slug), class: 'action-link' %>
+    <%= link_to "Add another #{current_format.title_for_new_document}", new_document_path(current_format.slug), class: 'action-link' %>
     <%= form_tag(documents_path(current_format.slug), method: :get, class: "add-vertical-margins well") do %>
       <%= label_tag "query", "Search" %>
       <%= text_field_tag("query", params[:query], class: "form-control add-bottom-margin") %>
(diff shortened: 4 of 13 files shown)


nnagewad (approved):
From a design perspective looks good

app/models/document.rb
```diff
@@ -308,6 +308,15 @@ def self.slug
...
   end
 
+  # Downcase but maintaining initials
+  def self.title_for_new_document
+    title.split(" ").map { |p| p.upcase == p ? p : p.downcase }.join(" ")
+  end
+
+  def self.title_starts_with_initials?
+    title[0...2].upcase == title[0...2]
+  end
```
kevindew: It doesn't look like this method is used anywhere? I assume it was a first attempt and not used in implementation.

app/models/document.rb
```diff
@@ -308,6 +308,15 @@ def self.slug
     title.parameterize.pluralize
   end
 
+  # Downcase but maintaining initials
+  def self.title_for_new_document
+    title.split(" ").map { |p| p.upcase == p ? p : p.downcase }.join(" ")
+  end
```
kevindew, line 314: It'd be nice to have a unit test for this. 

I also worry are there not more cases where this downcases something that should be title case? I see you got the statutory instrument so I'm guessing you went through them all.

app/views/documents/new.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= content_for :page_title, "New #{current_format.title.singularize}" %>
```
kevindew, line 1: I'm assuming the `singularize` method was used for changing a plural to singular. Do we have a problem with any that are named with plurals?

kevindew (commented):
Changes look decent, just a few comments to resolve.

app/models/document.rb
KludgeKML: Oh, yes - I thought I'd taken that out.

KludgeKML: Fixed.

app/views/documents/new.html.erb
KludgeKML: Ah yes. The default names almost all seem to be plurals as far as I can see - I should make sure this is moved into the title_for_new_document method.

app/models/document.rb
KludgeKML: Yeah, I went through them all - and I updated the tests that already existed. I think it's hard to know from the code what should be downcased and not, so it's probably down to a content person to check when a new type is added. I'll add a unit test for the method.

KludgeKML: Unit tests added

app/views/documents/new.html.erb
KludgeKML: I've moved this into the title_for_new_document method so that it behaves identically for both the link to this page and the title.

kevindew:
I've taken a bit more of a look @KludgeKML and I'm not convinced this is the right approach.

I think there are more cases of titles that should be capitalised than identified. For example: [Countryside Stewardship grants](https://www.gov.uk/countryside-stewardship-grants), [Drug Safety Update](https://www.gov.uk/drug-safety-update). So I think the approach here is a bit blunter than it should be, and instead to do it, we should reference the specialist finders for what the frontend capitalisation is.

I also notice there are various references in the code to titles within sentences (https://github.com/alphagov/specialist-publisher/blob/main/app/presenters/actions_presenter.rb#L46 and https://github.com/alphagov/specialist-publisher/blob/10fbe34513678979230d2ff78e8a614334319642/app/mailers/notifications_mailer.rb#L10) and it doesn't seem consistent to me to use lowercase on the "New <item>" page but capitalise there. 

I think it might be better to just update all the `title` methods to be the lowercase equivalents. E.g:

```
   def self.title
-    "AAIB Report"
+    "AAIB report"
   end

   def self.title
-    "Animal disease case"
+    "animal disease case"
   end

   def self.title
-    "Asylum Support Decisions"
+    "asylum support decisions"
   end

   def self.title
-    "Business Finance Support Scheme"
+    "business finance support scheme"
   end

   def self.title
-    "CMA Case"
+    "CMA case"
   end

   def self.title
-    "Countryside Stewardship Grant"
+    "Countryside Stewardship grant"
   end
...
```

As I think the grammatically correct lowercase usages far out numbers those with capitalis (caveat: I've not exhaustively checked).

If I understand right it'd leave just two views: https://github.com/alphagov/specialist-publisher/blob/10fbe34513678979230d2ff78e8a614334319642/app/views/documents/index.html.erb#L1-L7 and https://github.com/alphagov/specialist-publisher/blob/10fbe34513678979230d2ff78e8a614334319642/app/views/layouts/application.html.erb#L11-L26

Then for these we'll need a method on the model or a helper method that can do the equivalent of:

```
def title_for_heading(title)
  capitalised = title[0].upcase == title[0] ? title.capitalize : title
  capitalised.pluralize
end
```

Sorry to be a pain with this, just looks like this goes quite deep. 

There is a part of me that wonders is "New Licence" that bad?

KludgeKML:
No need to apologise, this is the whole point of checking, after all! :)

I don't think it's critical - the request comes from our content person on this project, and I think it's more to follow our style guide than from any serious problem it would cause the department users.

PR #2269, opened 2023-04-25
Update industries and tagging

1pretz1:
This PR does the following:

* updates licence transaction schema file with the new industries (source: ['Final version!' tab](https://docs.google.com/spreadsheets/d/1ZfGKnVKmfB5yGBWMx3PvwKq8KEj6Lt1XzcWXCJWBMf0/edit#gid=1567905948))
* updates import code to use CSV tagging vs tagging from licencefinder API
 
Industries and their tagged licences have deviated significantly from the tagging in the licencefinder tool and so we're now importing using a CSV which contains an industry / location mapping for each licence. This simplifies the migration process (no longer need a bespoke API endpoint), enables us to import all the other licences (rather than ~%60 of them that have a mapping in licence finder) and enables us to easily update the tagging going forward, as it would just involve updating [this spreadsheet](https://docs.google.com/spreadsheets/d/1yCUgzOiCXEuQS88B2O3Xc_iqQKq2mVra0hr_zhNWEcU/edit#gid=0), downloading as CSV, opening a PR and then running the import again.  

I've added CSV validation to ensure that each industry / locations matches a value defined in the schema which mitigate the risk of typos etc. 

There are some licences that are not currently mapped and so won't be imported as it stands. I've discussed this with Jessie (PM) and the action was to not migrate them for now, make a note of these and address them later. There is a Trello card capturing them: https://trello.com/c/kwBQLstL/2057-decide-approach-for-unmapped-licences-that-are-excluded-from-the-import.  

Tested running this in staging which imported correctly (total imported = 389 licences).

### Post import screenshots

Screenshot 2023-04-25 at 12 59 24
Screenshot 2023-04-25 at 12 59 32

Trello:
https://trello.com/c/hdqGmGDA/1939-add-new-industries-and-update-migration-code-to-use-new-mapping
https://trello.com/c/73ZybQsj/1940-import-all-licences


KludgeKML (approved):
Nice. LGTM

lib/data/licence_transaction/licences_and_tagging.csv
```diff
@@ -0,0 +1,409 @@
+Locations,Link,Industry 1,Industry 2,Industry 3,Industry 4,Industry 5,Industry 6,Industry 7,Industry 8,Industry 9,Industry 10,Industry 11,Industry 12,Industry 13,Industry 14,Industry 15,Industry 16,Industry 17,Industry 18,Industry 19
+Northern Ireland,https://www.gov.uk/licence-to-abstract-and-or-impound-water-northern-ireland,Environment: water,Agriculture: animals,,,,,,,,,,,,,,,,,
+Northern Ireland,https://www.gov.uk/consent-for-leaflet-distribution-northern-ireland,Advertising and marketing,Arts and recreation,"Beauty, hairdressing, nails and spa treatments","Hospitality, restaurants, bars and catering",Legal services,"Gambling and betting, including arcades, bingo and racing","Finance, insurance and insolvency",Events,Media and broadcasting,,,,,,,,,,
```
leenagupte, line 3: Should the link just be the base path, so it uses up the content item for the environment it's being run it?

lib/data/licence_transaction/licences_and_tagging.csv
```diff
@@ -0,0 +1,409 @@
+Locations,Link,Industry 1,Industry 2,Industry 3,Industry 4,Industry 5,Industry 6,Industry 7,Industry 8,Industry 9,Industry 10,Industry 11,Industry 12,Industry 13,Industry 14,Industry 15,Industry 16,Industry 17,Industry 18,Industry 19
```
leenagupte, line 1: And there are 19 industries as that is the current maximum in the spreadsheet?

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -18,8 +25,8 @@ def call
             public_updated_at: licence["public_updated_at"],
             first_published_at: licence["first_published_at"],
             update_type: "major",
-            licence_transaction_location: [],
-            licence_transaction_industry: [],
+            licence_transaction_location: tagging["locations"],
+            licence_transaction_industry: tagging["industries"],
```
leenagupte, line 38: So does this only add the tagging to new licences? What happens to the existing licences?
Perhaps there needs to be another method that dumps the tagging for existing licences and the re-tags them?

I think that's what was happening before too, so perhaps it's something that was overlooked, or I've forgotten how this works.

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -80,11 +85,35 @@ def licences
...
-          licence_identifiers.include?(licence["details"]["licence_identifier"])
+      def locations(grouped_licence)
+        grouped_licence.flat_map do |licence|
+          licence["Locations"].split(", ").map(&:parameterize)
+        end
+      end
+
+      def industries(grouped_licence)
+        grouped_licence.flat_map do |licence|
+          licence[2..].compact.map(&:parameterize)
```
leenagupte, line 119: Nice!

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -80,11 +85,35 @@ def licences
...
+        end
+      end
+
+      def grouped_licences
+        CSV.foreach(licence_tagging_path, headers: true).group_by do |licence|
+          licence["Link"]
```
leenagupte, line 125: So are there duplicate rows, or is this just to mitigate? Does that mean we're tagging to all of the industries in both rows, rather than overwriting the first row?

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -134,6 +136,8 @@ def expected_put_content_payload
           licence_transaction_continuation_link: "http://www.hpc-uk.org/apply",
           licence_transaction_licence_identifier: licence_identifier,
           licence_transaction_will_continue_on: "the Health and Care Professions Council (HCPC) website",
+          licence_transaction_industry: %w[healthcare],
+          licence_transaction_location: %w[england wales scotland northern-ireland],
```
leenagupte, line 167: Why have these rows been added here? Was this just missed in the original tests? I assume so, as I assume that we were always sending the industry and location information in the metadata.

lib/documents/schemas/licence_transactions.json
```diff
@@ -267,4 +267,4 @@
       "filterable": false
     }
   ]
-}
\ No newline at end of file
+}
```
leenagupte: This should be fixed up to where the lint violation occurred.

lib/data/licence_transaction/common_licence_identifiers.txt
```diff
@@ -1,228 +0,0 @@
-1245-7-1
```
leenagupte, line 1: I think deleting these files are good. I think all of the removed files could have been deleted in a separate commit to make this one easier to review.

lib/importers/licence_transaction/tagging_csv_validator.rb
```diff
@@ -0,0 +1,46 @@
...
+      def initialize(licences_tagging)
+        @licences_tagging = licences_tagging
+      end
+
+      def valid?
+        tagging_validation_errors.empty?
+      end
+
+      def errors
+        tagging_validation_errors.each { |e| puts "#{e}\n\n" }
```
leenagupte: If there are errors, I'd append a message about how to rectify the errors, e.g. "Please make sure you update the licence-transaction schema file first" and then put the link to the instructions.

spec/lib/importers/licence_transaction/tagging_csv_validator_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "spec_helper"
+require "importers/licence_transaction/tagging_csv_validator"
+
+RSpec.describe Importers::LicenceTransaction::TaggingCsvValidator do
+  it "returns true if tagging is valid" do
+    tagging = [
+      {
+        "base_path" => "/licence",
+        "locations" => %w[england],
+        "industries" => %w[street-trading-and-markets],
```
leenagupte, line 10: If we think the tags are likely to change I think it would worth creating fixture files for the CSV and schema.

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -80,11 +85,35 @@ def licences
         )["results"]
       end
 
-      def filtered_licences
-        licence_identifiers = File.new(common_licences_path).readlines(chomp: true)
+      def licences_tagging
```
leenagupte, line 97: I think this method and `licence_tagging` are too similarly named, and are therefore easy to misread and get confused.

spec/lib/importers/licence_transaction/licence_importer_spec.rb
```diff
@@ -16,6 +16,26 @@
     )
   end
 
+  context "when the csv tagging is invalid" do
+    before do
+      allow_any_instance_of(Pathname).to receive(:join)
```
leenagupte: Rather than stubbing Pathname, perhaps the schema and  csv files can be passed to Licence Importer, with default files used if they are not? That would allow an easier use of fixture files for all of the tests.

leenagupte (commented):
I think the CSV mapping is great idea, and a much easier way to tag industries.

I've added a few comments inline. On the whole though I think this is great.

lib/data/licence_transaction/licences_and_tagging.csv
1pretz1: The code uses the existing mainstream base_path, it only uses the spreadsheet link to check it exists (which pulls out the path of the URL to check: https://github.com/alphagov/specialist-publisher/pull/2269/files#diff-9fefdf86044e4f7f4075d0f74f7b114aecd6316621ac1cc8ba0c17621699e39aR94). I tried to modify the spreadsheet as little as possible to avoid introducing human errors.

lib/data/licence_transaction/licences_and_tagging.csv
1pretz1: Yep, for licences like https://www.gov.uk/notification-to-process-personal-data they’re tagged to 19 industries (https://docs.google.com/spreadsheets/d/1yCUgzOiCXEuQS88B2O3Xc_iqQKq2mVra0hr_zhNWEcU/edit#gid=0)

lib/importers/licence_transaction/licence_importer.rb
1pretz1: Currently (and previously) licences that already exist are skipped from the subsequent on a licence by licence basis (https://github.com/alphagov/specialist-publisher/blob/90ce29350e145210038565b23ab8a3227bb62d54/lib/importers/licence_transaction/licence_importer.rb#L43-L46). 

Before we launch, licences get deleted every night in the test environments and so any updates to the tagging file will be applied as the licence will be effectively ‘new’. 

If we wanted to update the code to update existing licences (e.g after we’ve run the import in production) then we’d need to update the code to do that. This feels like something we could do if we think that industry tagging is likely to change on bulk, post launch but that is probably out of scope of this piece of work.

lib/importers/licence_transaction/licence_importer.rb
1pretz1: There are ‘duplicate’ rows in the spreadsheet due to different licences actually pointing to the same licence page (intentionally). As I’ve not tried to modify the spreadsheet too much manually, the code takes this into consideration. 

The approach I took is just to add all the unique industries together for each link.

spec/lib/importers/licence_transaction/licence_importer_spec.rb
1pretz1: Previously, we were stubbing out the licences-api response as an empty array and so there were no industries / locations in the test examples (https://github.com/alphagov/specialist-publisher/blob/e7169d721eeef6ab5057a617d0f7834ecb52321f/spec/lib/importers/licence_transaction/licence_importer_spec.rb#L18), now we’re using the licences_and_tagging.csv file which defines industries and locations.

lib/documents/schemas/licence_transactions.json
1pretz1: Sure thing, good spot!

lib/data/licence_transaction/common_licence_identifiers.txt
1pretz1: I've moved them to a separate commit 👍

lib/importers/licence_transaction/tagging_csv_validator.rb
1pretz1: Good point, I’ve added a descriptive message which points to a Trello card that outlines the steps needed to update the tagging (and industry list if required). https://trello.com/c/2SBbuD8N/1969-instructions-for-updating-industry-facet-values-and-the-their-licence-tagging

lib/importers/licence_transaction/licence_importer.rb
1pretz1: Sure, I’ve renamed `licence_tagging`

spec/lib/importers/licence_transaction/tagging_csv_validator_spec.rb
1pretz1: I see your point, however I think it’s v low effort to update the test (on the basis that exact tag changes) when you already have a PR open for the industry updates (vs making the code / tests more complex to take in a fixture file). Thoughts?

spec/lib/importers/licence_transaction/licence_importer_spec.rb
1pretz1: Sure, I did ponder this (influencing application code just for tests). However, now passing the CSV file which removes the stubbing 👍

lib/importers/licence_transaction/licence_importer.rb
leenagupte: Thanks for the reminder.

lib/importers/licence_transaction/licence_importer.rb
leenagupte: Out of curiosity, do you have an example? I'd love to see it in action.

I assume the content designers are aware of the duplication? It's something they may want to look into.

spec/lib/importers/licence_transaction/tagging_csv_validator_spec.rb
leenagupte: I think if the tags are pretty stable now, and the tests only need to be updated rarely then it's fine. If they keep going through many iterations you may want to revisit this.

lib/importers/licence_transaction/licence_importer.rb
1pretz1: Sure, I did spot checked the duplicates and they looked reasonable as either a licence page covers two licences (like https://www.gov.uk/licences-use-music-online-mobile) or there are separate rows per location (like https://www.gov.uk/greenhouse-gas-emissions-permit). The actual licence list came from a BEIS audit, rather than something we generated so I'd expect (hope) they'd be correct, we just added the proposed industries for each licence. I've let Jessie know though 👍 

I've just pulled the links that have more than one row below, you can search for the links in the [spreadsheet](https://docs.google.com/spreadsheets/d/1ZfGKnVKmfB5yGBWMx3PvwKq8KEj6Lt1XzcWXCJWBMf0/edit#gid=1760190301) under the 'Matching industries to BEIS's licence audit' tab.

```
=> ["https://www.gov.uk/approval-and-accreditation-of-vocational-training-courses-scotland",
 "https://www.gov.uk/gas-electricity-supply-licence",
 "https://www.gov.uk/scheduled-monument-consent-england-scotland-wales",
 "https://www.gov.uk/greenhouse-gas-emissions-permit",
 "https://www.gov.uk/fireworks-manufacture-registration",
 "https://www.gov.uk/licences-use-music-online-mobile",
 "https://www.gov.uk/registration-of-independent-health-care-service-england-wales",
 "https://www.gov.uk/second-hand-dealers-registration-scotland"]
```

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -80,11 +85,35 @@ def licences
...
-          licence_identifiers.include?(licence["details"]["licence_identifier"])
+      def locations(grouped_licence)
+        grouped_licence.flat_map do |licence|
+          licence["Locations"].split(", ").map(&:parameterize)
+        end
+      end
+
+      def industries(grouped_licence)
+        grouped_licence.flat_map do |licence|
+          licence[2..].compact.map(&:parameterize)
```
leenagupte, line 119: Given the problem discussed previously about licences being duplicated in the source data, perhaps add a `uniq` here so that we aren't saving duplicate tags. e.g. `licence[2..].compact.uniq.map(&:parameterize)`

lib/importers/licence_transaction/licence_importer.rb
leenagupte: Ok, that's fair.

lib/importers/licence_transaction/tagging_csv_validator.rb
```diff
@@ -0,0 +1,55 @@
...
+      def parsed_licence_schema
+        schema_file_path = Rails.root.join("lib/documents/schemas/licence_transactions.json")
+        json_blob = File.new(schema_file_path).read
+        JSON.parse(json_blob)
+      end
+
+      def unrecognised_tags_instructions
+        <<~HEREDOC
+          Please read the instructions (under heading 'Update tagging') in the following link to resolve the unrecognised
+          tags errors: https://trello.com/c/2SBbuD8N/1969-how-to-correct-unrecognised-tags-when-importing-licences
```
leenagupte, line 51: It's slightly evil to make people copy and paste link text from the command line rather than printing out the instructions, but ok 🙈

lib/importers/licence_transaction/tagging_csv_validator.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+
+      def valid?
+        tagging_validation_errors.empty?
+      end
+
+      def errors
+        tagging_validation_errors.each { |e| puts "#{e}\n\n" }
+
+        puts unrecognised_tags_instructions unless valid?
```
leenagupte: You could put `valid?` check at the top of the method to avoid any extra processing.

i.e.

```ruby
  def errors
    return if valid?

    tagging_validation_errors.each { |e| puts "#{e}\n\n" }
    puts unrecognised_tags_instructions
  end

```

leenagupte (commented):
I like changes made to pass in the sectors file 👍 

There are a couple of minor inline suggestions. But on the whole it looks good!

lib/importers/licence_transaction/licence_importer.rb
1pretz1: We already are [here](https://github.com/alphagov/specialist-publisher/pull/2269/files/660f5aa5a619ba781086be015ecb0198bd7271f4..720620a1b69ac05349a40a5f5464d14a219099aa#diff-9fefdf86044e4f7f4075d0f74f7b114aecd6316621ac1cc8ba0c17621699e39aR92-R93). Having a unique at that level ensures that there are no unique values in the final list (vs on line 110 where it would only ensure uniqueness within each row).

lib/importers/licence_transaction/tagging_csv_validator.rb
1pretz1: Haha maybe but you get all the extra context too without having to write a dissertation in the code + this will be short lived code

lib/importers/licence_transaction/tagging_csv_validator.rb
1pretz1: Sure thing 👍 Done!

leenagupte (approved):
Looks great!

PR #2274, opened 2023-05-02
Ignore empty strings for licence import

1pretz1:
continuation_link / will_continue_on can be empty strings which previously led to the identifier not being saved if these were present. This commit updates the conditional to `present?` which ignores empty strings.

We're also not saving will_continue_on / continuation_link if their value is an empty string which is cleaner.


KludgeKML (approved):
LGTM

PR #2277, opened 2023-05-10
Add validation to licence fields

1pretz1:
Adds validation rules as specified in the [Trello card](https://trello.com/c/uf3E7dCl/1974-add-validation-to-new-licence-document-fields). 

More info in the commits ->

## Before

Screenshot 2023-06-22 at 18 13 58

## After

Screenshot 2023-06-26 at 15 22 25
Screenshot 2023-06-26 at 15 16 48
Screenshot 2023-06-26 at 15 17 45
Screenshot 2023-06-26 at 15 17 59
Screenshot 2023-06-26 at 15 19 21
Screenshot 2023-06-26 at 15 19 45
Screenshot 2023-06-26 at 15 20 14

Trello:
https://trello.com/c/uf3E7dCl/1974-add-validation-to-new-licence-document-fields


app/validators/licence_identifier_unique_validator.rb
```diff
@@ -0,0 +1,15 @@
+class LicenceIdentifierUniqueValidator < ActiveModel::EachValidator
+  def validate_each(record, attribute, value)
+    return if value.blank?
+
+    existing_licence_identifiers = []
+
+    record.class.find_each do |licence|
```
KludgeKML, line 7: How fast is this? Is it making 309 api calls every time we try to save a licence?

KludgeKML (approved):
LGTM (Pending content approval of warning messages)

app/validators/licence_identifier_unique_validator.rb
KludgeKML: (Update - it's more like 8 API calls, as they're done in batches of 50, and there isn't currently a good optimisation we can do here. Appears to be quick in integration, we should just accept this is not perfect but necessary and keep an eye on whether it causes slowdowns in use. Marking as resolved).

app/views/metadata_fields/_licences.html.erb
```diff
@@ -28,11 +28,17 @@
   %>
 <% end %>
 
-<%= render layout: "shared/form_group", locals: { f: f, field: :licence_transaction_will_continue_on, label: "Name of website where you apply" } do %>
+<div class="additional-field-context-section">
+  <h2>How users apply</h3>
```
MartinJJones: Small thing, the browser will likely fix this anyway when rendering the page, but the opening tag here is a `h2`, and the closing tag is a `h3`.

MartinJJones (commented):
Changes look good to me overall, just one small thing to fix 👍

app/views/metadata_fields/_licences.html.erb
1pretz1: Good spot, now fixed!

app/validators/link_or_identifier_validator.rb
```diff
@@ -0,0 +1,37 @@
+class LinkOrIdentifierValidator < ActiveModel::Validator
+  def validate(record)
+    if link_and_identifier_exists?(record) || link_and_identifier_missing?(record)
```
KludgeKML: Could there be a better name for these functions? Because it reads like: trigger an error if either link and id exists, or link and id are missing.

1pretz1: Sure, we can rejig the naming if it's not clear, however your assumption of 'trigger an error if either link and id exists, or link and id are missing.'  is correct. We only want to trigger an alert if:

a) the link AND identifier fields are both filled in or 
b) the link AND identifier fields are both blank. 

A user filling in one OR the other is the correct procedure.

How about, `both_link_and_identifier_exist? || both_link_and_identifier_missing?` - thoughts? Happy for other suggestions too!

KludgeKML: Yeah, I like those new names. Nice.

MartinJJones (approved):
Frontend changes look good to me 👍

KludgeKML (approved):
Looks good to me (with caveat of the variable names)

app/validators/link_or_identifier_validator.rb
1pretz1: Done!

PR #2288, opened 2023-05-22
Set licence primary_publishing_organisation from current user.

KludgeKML:
- As individual licences will be owned by several different organisations, we can't hardcode ownership as with other documents.
- Update to use the has_organisations? option as used by Statutory Instruments
- Set the initial Primary Publishing Organisation to the creating user's organisation.

https://trello.com/c/Y6GnGLLM/2007-use-current-users-department-as-primary-publisher-for-licences-in-specialist-publisher

### In a new licence (edited by a GDS member, so GDS is auto-selected as Primary Publishing Organisation)
Screenshot 2023-05-25 at 14 14 58

### Editing an existing licence (created by another department, so GDS is not auto-selected as Primary Publishing Organisation), with additional organisations
Screenshot 2023-05-25 at 14 36 23

### DIFF
--- a/app/helpers/organisations_helper.rb
+++ b/app/helpers/organisations_helper.rb
@@ -2,4 +2,10 @@ module OrganisationsHelper
   def organisations_options
     Organisation.all.map { |o| [o.title, o.content_id] }
   end
+
+  def primary_publishing_organisation_options(form)
+    return {} if form.object.primary_publishing_organisation.present?
+
+    { selected: current_user.organisation_content_id }
+  end
 end
--- a/app/models/licence_transaction.rb
+++ b/app/models/licence_transaction.rb
@@ -1,4 +1,6 @@
 class LicenceTransaction < Document
+  validates :primary_publishing_organisation, presence: true
+
   FORMAT_SPECIFIC_FIELDS = %i[
     licence_transaction_continuation_link
     licence_transaction_industry
@@ -7,10 +9,12 @@ class LicenceTransaction < Document
     licence_transaction_will_continue_on
   ].freeze
 
-  attr_accessor(*FORMAT_SPECIFIC_FIELDS)
+  attr_accessor(*FORMAT_SPECIFIC_FIELDS, :organisations, :primary_publishing_organisation)
 
   def initialize(params = {})
     super(params, FORMAT_SPECIFIC_FIELDS)
+    @primary_publishing_organisation = params[:primary_publishing_organisation]
+    @organisations = params[:organisations]
   end
 
   def taxons
@@ -21,9 +25,15 @@ class LicenceTransaction < Document
     "Licence"
   end
 
-  def primary_publishing_organisation
-    # Set to GDS for testing
-    "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+  def has_organisations?
+    true
+  end
+
+  def links
+    super.merge(
+      organisations: organisations | [primary_publishing_organisation],
+      primary_publishing_organisation: [primary_publishing_organisation],
+    )
   end
 
   def route_type
--- a/lib/documents/schemas/licence_transactions.json
+++ b/lib/documents/schemas/licence_transactions.json
@@ -11,7 +11,7 @@
   },
   "show_summaries": true,
   "organisations": [
-    "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+    "aa750cdf-7925-429d-a2b3-0d9fa47d2c48"
   ],
   "document_noun": "licence",
   "facets": [
--- a/lib/importers/licence_transaction/licence_importer.rb
+++ b/lib/importers/licence_transaction/licence_importer.rb
@@ -39,6 +39,8 @@ module Importers
             licence_transaction_will_continue_on: details["will_continue_on"].presence,
             licence_transaction_continuation_link: details["continuation_link"].presence,
             licence_transaction_licence_identifier: licence_identifier(details),
+            primary_publishing_organisation: tagging["primary_publishing_organisation"] || ::LicenceTransaction.schema_organisations.first,
+            organisations: tagging["organisations"]&.split(";") || [],
             imported: true,
           )
 
--- a/app/views/metadata_fields/_licences.html.erb
+++ b/app/views/metadata_fields/_licences.html.erb
@@ -28,7 +28,6 @@
   %>
 <% end %>
 
-
 <%= render layout: "shared/form_group", locals: { f: f, field: :licence_transaction_will_continue_on, label: "Name of website where you apply" } do %>
   <%= f.text_field :licence_transaction_will_continue_on, class: 'form-control' %>
 <% end %>
@@ -39,4 +38,31 @@
 
 <%= render layout: "shared/form_group", locals: { f: f, field: :licence_transaction_licence_identifier, label: "Licence identifier" } do %>
   <%= f.text_field :licence_transaction_licence_identifier, class: 'form-control' %>
+<% end %>
+
+<%= render "shared/form_group", f: f, field: :primary_publishing_organisation, label: "Publishing organisation" do %>
+  <%= f.select :primary_publishing_organisation,
+      organisations_options,
+      {},
+      {
+        class: "select2 form-control",
+        multiple: false,
+        data: {
+          placeholder: "Select a primary publishing organisation"
+        }
+      }
+  %>
+<% end %>
+<%= render "shared/form_group", f: f, field: :organisations, label: "Other associated organisations" do %>
+  <%= f.select :organisations,
+      organisations_options,
+      {},
+      {
+        class: "select2 form-control select-all-disabled",
+        multiple: true,
+        data: {
+          placeholder: "Select organisations"
+        }
+      }
+  %>
 <% end %>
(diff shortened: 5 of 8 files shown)


app/models/licence_transaction.rb
```diff
@@ -21,9 +22,14 @@ def self.title
...
+  def schema_organisations
+    return [] unless @creator_organisation
+
+    [@creator_organisation]
+  end
+
   def primary_publishing_organisation
-    # Set to GDS for testing
-    "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+    @creator_organisation
```
leenagupte: What happens if no `primary_publishing_organisation` is set? Perhaps we should keep a default here?

leenagupte (commented):
This looks good. I have one inline question.

app/models/licence_transaction.rb
KludgeKML: I _think_ that can't happen, because the only way you can get to document.save (where this is called) on a new licence is going through the controller's create action. 

That said, I clearly made a different decision here: https://github.com/alphagov/specialist-publisher/pull/2288/files#diff-0221e5c51589e073e450a8584ee6385eed859ac34eaaec48f886d4a2786d9f3bR25-R29 .. (although that method is returning an array, so it needs some massaging).

I think on balance it won't hurt to be cautious about a default (although I think in practise the safeguard will never be needed). I'll update.

KludgeKML: Fixed

leenagupte: Thanks. Can you add a note to the commit or a comment saying that this guid is for GDS?

KludgeKML: Done!

leenagupte (dismissed):
👍

app/models/licence_transaction.rb
```diff
@@ -21,9 +22,17 @@ def self.title
     "Licence"
   end
 
+  def schema_organisations
```
1pretz1: I think we probably shouldn't be using a method called `schema_organisations` as the organisation is not being pulled by the schema in this case so is a little confusing. Maybe just 'organisations'? We'd also need to update the [parent method](https://github.com/alphagov/specialist-publisher/blob/7f67d74dba73813b9740e628806fc775f435de9c/app/models/document.rb#L189) and the references to it within the codebase

lib/documents/schemas/licence_transactions.json
```diff
@@ -10,9 +10,6 @@
     "format": "licence_transaction"
   },
   "show_summaries": true,
-  "organisations": [
```
1pretz1, line 13: I think we'll still want the organisations attribute as this also defines the finders organisation (from: ... on /find-licences). However, this is probably a good opportunity to swap the organisation from GDS -> BEIS as we only set this to GDS as we were drafting the prototype.

app/models/licence_transaction.rb
```diff
@@ -21,9 +22,17 @@ def self.title
...
+    return [] if @creator_organisation.blank?
+
+    [@creator_organisation]
+  end
+
   def primary_publishing_organisation
-    # Set to GDS for testing
-    "af07d5a5-df63-4ddc-9383-6a666845ebe9"
+    # Return GDS organisation ID if not supplied
+    return "af07d5a5-df63-4ddc-9383-6a666845ebe9" if @creator_organisation.blank?
```
1pretz1: If a creating organisation doesn't exists then I think it probably makes sense to set the publishing org to BEIS (rather than us). GDS was just used for testing and should probably be swapped now?

1pretz1 (commented):
Nice work so far Keith. The actual setting of the primary publishing organisation works but there is a slight issue with the current implementation.

In both methods (schema_organisations and primary_publishing_organisation) we're using the `@creator_organisation` var, however this is only available when we're initially creating a document and so subsequent editing (or [the time it is published](https://github.com/alphagov/specialist-publisher/blob/06d7073276c54083757fd55c275b5c7e2672039c/app/services/document_publisher.rb#L9)) will result in the publishing [organisation being overwritten](https://github.com/alphagov/specialist-publisher/blob/06d7073276c54083757fd55c275b5c7e2672039c/app/presenters/document_links_presenter.rb#L11) as we have a guard that sets the `primary_publishing_organisation` if the var is blank (which it will be after first creation) to GDS when the method is called.

To mitigate this, we'll need to fetch the documents organisation from Publishing API, rather than from an instance var after the initial draft has been created. 

If we want to guard against a user having no organisation (this does seem possible from Signon so nice catch) and setting it to a default then we'll need to only apply this on first creation. 

Something to note is that I imagine we don't want to update the publishing organisation, for subsequent editions, even if the current users org is different which I believe is how it works for other apps.  

Could also a test case for the above scenario to catch this going forward 👍

leenagupte:
@KludgeKML I'm going to dismiss my review as @1pretz1 has raised some really good points that should be addressed.

app/models/licence_transaction.rb
KludgeKML: Hmm. BEIS doesn't exist any more, does it? Who's taken over this responsibility? DBT? (DSIT?)

lib/documents/schemas/licence_transactions.json
KludgeKML: Cool (but same question as below - BEIS no longer a thing, so is it DBT now?)

1pretz1: Ah yes it is now DBT, doh!

app/models/licence_transaction.rb
1pretz1: See comment above

KludgeKML:
@leenagupte / @1pretz1 

I've reworked this based on the fact that there's already a couple of document types in SP that handle organisations and primary publishing organisation (Statutory Instruments / Flood and Coastal Erosion Risk Management Research Reports), so there's existing support for it. As an addition I've added a tweak to the form creation that sets the PPO to the current user's PPO if there isn't already one for the document. If the user's organisation is blank, this of course won't work, but they'll still be forced to pick one because the validation will kick in, so here we've covered the "user has no org" problem.

Note that to get the test suite to pass I've had to add a feature to the importer, so it's possible we've also encroached on this card: https://trello.com/c/9qWrtRto/2084-update-import-code-to-use-tagging-csv-organisation-as-owning-organisation

lib/importers/licence_transaction/licence_importer.rb
```diff
@@ -39,6 +39,8 @@ def call
             licence_transaction_will_continue_on: details["will_continue_on"].presence,
             licence_transaction_continuation_link: details["continuation_link"].presence,
             licence_transaction_licence_identifier: licence_identifier(details),
+            primary_publishing_organisation: tagging["primary_publishing_organisation"] || ::LicenceTransaction.schema_organisations.first,
```
1pretz1, line 42: I'd probably want validation for a missing primary publishing organisation, rather than automatically using the default org but this can be addressed at a later point as it isn't a concern of this PR 👍

1pretz1 (approved):
Nice work Keith! This is much better approach, also great work on automatically selecting the users org when creating a document ⭐ 

Don't have any suggestions so approving 💯

app/helpers/organisations_helper.rb
```diff
@@ -2,4 +2,10 @@ module OrganisationsHelper
   def organisations_options
     Organisation.all.map { |o| [o.title, o.content_id] }
   end
+
+  def primary_publishing_organisation_options(form)
```
1pretz1, line 6: Maybe we could add this to the other doc types that use an org select? (possibly in another PR)

KludgeKML: Yes, I was wondering about that. I guess the question is - for the other two types, is the lack of a default a feature? (if the field value isn't usually the current user's org, then the field being empty by default plus the validation for presence acts as a gate, the user can't accidentally save/publish without deliberately putting in the value).

lib/importers/licence_transaction/licence_importer.rb
KludgeKML: Agreed - we have the ticket for actually doing the import - it'll build on this I guess, and add proper validation of primary publishing org in the CsvTaggingValidator class.

app/helpers/organisations_helper.rb
1pretz1: I'm not sure whether it is an explicit feature or lack of feature in this instance, however the general pattern in our publishing apps is to autofill the primary publishing organisation from the users current org (e.g in Whitehall, Content Publisher ...) so I'd assume the change would be expected. 

I also think that if we have those concerns then we probably don't want to do it just for licences 👍 

In conclusion, as there is precedent in our publishing apps, my vote is 'do it' 😄

KludgeKML: Okay, cool - once this is merged I'll make another PR with the two other document types covered as well.

leenagupte: I agree with @1pretz1, this seems like a good feature to add to the other two specialist documents that also allow organisations to be set.

leenagupte (commented):
This looks good. Have you managed to get it running locally? Can you add some screenshots to the PR description? Thanks.

app/helpers/organisations_helper.rb
KludgeKML: (I've also confirmed with Content 2L that this is a desirable feature, so will definitely add it to the other document types in a follow-up PR once this is merged.)

KludgeKML:
@leenagupte Screenshots added.

leenagupte (approved):

PR #2295, opened 2023-05-26
Adding a rake task to unpublish and redirect a finder.

minhngocd:
There is request coming from departments to decommisssion a finder and redirect it to a different part of service.gov.uk. The rake task will enable us to get this work done in time, with the aim to expand and support unpublishing documents later on.

https://trello.com/c/ZBF26F1U/1322-redirect-and-potentially-archive-old-finder-by-30th-may

### DIFF
--- a/lib/tasks/unpublish.rake
+++ b/lib/tasks/unpublish.rake
@@ -1,0 +1,35 @@
+require "services"
+
+namespace :unpublish do
+  desc "Unpublish a Finder by slug and redirect the finder page to specific URL"
+  task :redirect_finder, %i[finder_slug redirect_url] => :environment do |_, args|
+    schema_file = Dir.glob("lib/documents/schemas/*.json").find do |file|
+      File.foreach(file).grep(/"base_path": "\/#{args.finder_slug}"/).any?
+    end
+
+    if schema_file.nil?
+      puts "Could not find any finders with that slug. Please check again."
+      next
+    end
+
+    schema = MultiJson.load(File.read(schema_file))
+    puts "=== Finder found ==="
+    puts "Slug: #{args.finder_slug}"
+    puts "Finder name: #{schema['name']}"
+    puts "Content ID: #{schema['content_id']}"
+    puts "Redirecting to: #{args.redirect_url}"
+
+    begin
+      response = GdsApi.publishing_api.unpublish(
+        schema["content_id"],
+        type: "redirect",
+        alternative_path: args.redirect_url,
+        discard_drafts: true,
+      )
+      puts "Publishing API response #{response.code}: #{response.raw_response_body}"
+      puts "Finder unpublished"
+    rescue GdsApi::HTTPServerError => e
+      puts "Error unpublishing finder: #{e.inspect}"
+    end
+  end
+end
--- a/lib/tasks/unpublish.rake
+++ b/lib/tasks/unpublish.rake
@@ -1,20 +1,13 @@
 require "services"
 
 namespace :unpublish do
-  desc "Unpublish a Finder by slug and redirect the finder page to specific URL"
-  task :redirect_finder, %i[finder_slug redirect_url] => :environment do |_, args|
-    schema_file = Dir.glob("lib/documents/schemas/*.json").find do |file|
-      File.foreach(file).grep(/"base_path": "\/#{args.finder_slug}"/).any?
-    end
+  desc "Unpublish a Finder by file name and redirect the finder page to specific URL"
+  task :redirect_finder, %i[finder_file redirect_url] => :environment do |_, args|
+    require "finder_loader"
 
-    if schema_file.nil?
-      puts "Could not find any finders with that slug. Please check again."
-      next
-    end
-
-    schema = MultiJson.load(File.read(schema_file))
+    schema = FinderLoader.new.finder(args.finder_file).first[:file]
     puts "=== Finder found ==="
-    puts "Slug: #{args.finder_slug}"
+    puts "Slug: #{schema['base_path']}"
     puts "Finder name: #{schema['name']}"
     puts "Content ID: #{schema['content_id']}"
     puts "Redirecting to: #{args.redirect_url}"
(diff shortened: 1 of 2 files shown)


lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,42 @@
...
+    schema_file = Dir.glob("lib/documents/schemas/*.json").find do |file|
+      File.foreach(file).grep(/"base_path": "\/#{args.finder_slug}"/).any?
+    end
+
+    if schema_file.nil?
+      puts "Could not find any finders with that slug. Please check again."
+      next
+    end
+
+    schema = MultiJson.load(File.read(schema_file))
```
ollietreend: I think you might be doing things the 'hard' way here. The [publishing_api:publish_finder][1] task makes use of [FinderLoader][2] which seems a simpler way to do it.

I think something like this should work:

```suggestion
  task :redirect_finder, %i[name redirect_url] => :environment do |_, args|
    begin
      schema = FinderLoader.new.finder(args.name)[0].fetch(:file)
    rescue StandardError => e
      puts "Error: #{e.inspect}"
      next
    end
```

[1]: https://github.com/alphagov/specialist-publisher/blob/caf19121fa3a82eec4f0ff828b88a075e09fe275/lib/tasks/publishing_api.rake#L18-L37
[2]: https://github.com/alphagov/specialist-publisher/blob/caf19121fa3a82eec4f0ff828b88a075e09fe275/app/lib/finder_loader.rb

lib/tasks/unpublish.rake
```diff
@@ -0,0 +1,42 @@
...
+    puts "Finder name: #{schema['name']}"
+    puts "Content ID: #{schema['content_id']}"
+    puts "Redirecting to: #{args.redirect_url}"
+    puts "Are you sure you want to unpublish this finder? (y/n)"
+
+    user_approval = $stdin.gets.strip.downcase
+    if user_approval != "y"
+      puts "Cancelled action. Not unpublishing anything."
+      next
+    end
```
ollietreend: As we discussed in Slack: I don't think we gain user consent before performing the unpublish action.

Rake tasks haven't traditionally run in interactive TTYs in GOV.UK, so the convention is for Rake tasks to 'just do the thing' rather than wait for user input.

If a developer misfires and accidentally unpublishes the wrong thing, they always have the option of republishing it with the `publishing_api:publish_finder` task.

ollietreend (commented):
Nice work @minhngocd. I've just added a couple of comments - one of which we've already discussed though!

lib/tasks/unpublish.rake
minhngocd: The problem with this is, it loads via filename, and not the slug - which is not consistantly the same thing in specialist finder. from a support perspective i just thought finding by slug would be the easiest since we can copy from the URL. I could also add the functionality to FinderLoader - what do you think @ollietreend ?

lib/tasks/unpublish.rake
minhngocd: Sure i can remove

lib/tasks/unpublish.rake
ollietreend: @minhngocd I think we should try to make this Rake task consistent with the other one – otherwise, it could be more confusing if the command to re-publish a finder requires a different param than the one to unpublish it.

I don't have a strong opinion about whether the `FinderLoader` should support finding by slug as well as (or instead of) finding by the JSON file name. Given that's how the existing `publishing_api:publish_finder` task works, I'd be inclined to say it isn't a problem – or at least it hasn't been inconvenient enough to have triggered someone to change it already. And since these Rake tasks aren't things we need to run every day, I feel like it's probably fine.

However, if you think that proposed behaviour would be a valuable addition, then I agree it's probably something to change in `FinderLoader` as then it'd be consistent across both Rake tasks. It might be something to split into a separate "improve `FinderLoader#finder`" PR so as not to block this one.

minhngocd: i would push for FinderLoader to load by slug, given the filename inconsistencies e.g. 
base path: find-licenses, filename: licence_transactions, 
base path: drug-device-alerts, filename: medical_safety_alerts 

and at least 3 more examples. 
But sure, i'll update this to use FinderLoader and put another PR out for FinderLoader

ollietreend: Eeek 😱 I didn't realise the filenames/slugs were as inconsistent as that!

Yeah - definitely open a PR for improving `FinderLoader` then.

spec/lib/tasks/unpublish_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    it "unpublishes finder with redirect" do
+      content_id = MultiJson.load(File.read("lib/documents/schemas/aaib_reports.json"))["content_id"]
+      stub_publishing_api_unpublish(content_id, body: {
+        type: "redirect",
+        alternative_path: "https://service.gov.uk",
+        discard_drafts: true,
+      })
+      task.invoke("aaib_reports", "https://service.gov.uk")
+      expect(output.string).to include("Publishing API response 200")
+      expect(output.string).to include("Finder unpublished")
```
ollietreend, line 23: It'd be good to assert that the Publishing API actually gets called - not just the text output. Or is that implicitly done by `stub_publishing_api_unpublish` already?

The spec for `publishing_api:publish_finder` includes some assertions – I'm wondering if there's one for 'unpublish' that we can use, too:

https://github.com/alphagov/specialist-publisher/blob/caf19121fa3a82eec4f0ff828b88a075e09fe275/spec/lib/tasks/publishing_api_spec.rb#L62-L64

minhngocd: It is implicitly done since in the before block it's stubbing out everything to error, and this test specifically stubs the specific call to return successfully. but i can look for an explicit assert

ollietreend (approved):
Given what you just said about the implicit assertion, I think this PR should be good to go 👍🏻 

Although it would be nicer if there was some way to make the assertion more explicit. Just to cover us in case the `before` block changes and this test inadvertently stops being so useful.

spec/lib/tasks/unpublish_spec.rb
minhngocd: all good - updated the tests with the explicit assert - it reads better now 👍

PR #2343, opened 2023-07-10
Skip validation for licence with link but no text

1pretz1:
Currently ~40 licences in Publisher don't have will continue on text but do have a link as this validation doesn't exist in Publisher. Now we've implemented it here, we need to skip that certain validation for the import to successfully import all licences. Next time these licences are edited, they will need the link text to save and publish so the situation will resolve in time.

Tested on integration and all licence now imported successfully. 

Trello:
https://trello.com/c/FUHCUnxO/2096-skip-validation-for-will-continue-on-text-when-importing


KludgeKML (approved):
LGTM

PR #2358, opened 2023-07-24
Added the "not applicable" market sector option to CMA case schema

ryanb-gds:
The Competition and Markets authority have recently started including Office for the Internal Market projects and Subsidy Advice Unit referrals in their CMA case specialist finder. However, the market sector field in the CMA case data type does not apply to these new types. Therefore, we are adding a new "not applicable" option to the CMA case market sector field.

Trello card: https://trello.com/c/SdjSO9PB


beccapearce (approved):
Looks good, this will need the same changes made in publishing-api and potentially search-api as well :)

ryanb-gds:
> Looks good, this will need the same changes made in publishing-api and potentially search-api as well :)

I've already applied to publishing API. Will look into search, thanks for the heads up

PR #2390, opened 2023-09-07
Add orgs to metadata if present

KludgeKML:
Where documents may have Primary Publishing Organisation / Organisations, show these in the metadata so that editors don't have to edit the document to see the owning orgs.

https://trello.com/c/q5FMBg6c/2167-add-owning-organisations-to-licence-overview-page-in-specialist-publisher

### DIFF
--- a/app/views/documents/show.html.erb
+++ b/app/views/documents/show.html.erb
@@ -27,6 +27,14 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if @document.class.has_organisations? %>
+        <dt>Publishing organisation</dt>
+        <dd><%= organisation_name(@document.primary_publishing_organisation) %></dd>
+        <dt>Other associated Organisations</dt>
+            <% @document.organisations.each do |org| %>
+            <dd><%= organisation_name(org) %></dd>
+          <% end %>
+      <% end %>
       <dt>Bulk published</dt>
       <dd><%= @document.bulk_published %></dd>
       <dt>Publication state</dt>
--- a/app/helpers/organisations_helper.rb
+++ b/app/helpers/organisations_helper.rb
@@ -1,6 +1,10 @@
 module OrganisationsHelper
+  def all_organisations
+    @all_organisations ||= Organisation.all
+  end
+
   def organisations_options
-    Organisation.all.map { |o| [o.title, o.content_id] }.sort_by { |title, _id| title.downcase.strip }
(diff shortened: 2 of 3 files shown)


app/views/documents/show.html.erb
```diff
@@ -27,6 +27,14 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if @document.class.has_organisations? %>
+        <dt>Primary Publishing Org</dt>
```
1pretz1: Minor: could we match the title against the input field labels, either by changing these or the input labels? Currently we have: "Publishing organisation" and "Other associated organisations". I do acknowledge that some field labels have fallen out of sync with the metadata counterpart but would be quick and easy to change in this case. Thoughts?

app/views/documents/show.html.erb
```diff
@@ -27,6 +27,14 @@
...
+        <dd><%= organisation_name(@document.primary_publishing_organisation) %></dd>
+        <dt>Organisations</dt>
+            <% @document.organisations.each do |org| %>
+            <dd><%= truncate(organisation_name(org), length: 140) %></dd>
```
1pretz1: Wondered why we are truncating at that specific point but then realised it's the precedent for the other fields 👍

1pretz1 (approved):
Looks good, one minor comment but non blocking so approving and leaving to your judgement 👍

app/views/documents/show.html.erb
```diff
@@ -27,6 +27,14 @@
           <% end %>
         <% end %>
       <% end %>
+      <% if @document.class.has_organisations? %>
```
1pretz1, line 30: Post review thoughts: could we also save the organisations in a variable (`orgs = Organisation.all`) and pass them to the name helper? As currently we fetch all the organisations from Publishing API for each organisation to retrieve the name.

app/views/documents/show.html.erb
KludgeKML: Yeah, I actually don't know if it's as vital. I might remove it - I don't think the other ones are much more open in what they accept, so maybe truncating them is vital. We control org names, though, so we're not expecting any surprises.

app/views/documents/show.html.erb
KludgeKML: Oh, yes - definitely.

app/views/documents/show.html.erb
KludgeKML: Yeah, we could memoize the orgs in a helper. I'd forgotten that we make a full call for each run of Organisation.all.

PR #2415, opened 2023-10-19
Upgrade ruby to 3.2.2

beccapearce:


ryanb-gds (approved):
This looks suspiciously easy 🤔

beccapearce:
Yeah I'm a bit suspicious!

But I've also checked the changelog and nothing seems like it should affect us. 🤞🏻

PR #2422, opened 2023-11-06
Add Google Tag Manager snippet

beccapearce:
Google Tag Manager (GTM) can be included on all pages of the app by configuring the following environment variables:

- GOOGLE_TAG_MANAGER_ID
- GOOGLE_TAG_MANAGER_AUTH
- GOOGLE_TAG_MANAGER_PREVIEW

All three variables are optional.

If none are set, GTM will not be included on the page.

If only GOOGLE_TAG_MANAGER_ID is set, the default environment of the GTM container will be used.

If GOOGLE_TAG_MANAGER_AUTH and GOOGLE_TAG_MANAGER_PREVIEW are set, the specified GTM container environment will be used. This is used for managing the rollout of changes to integration so they can be tested before rolling out to production.

trello card: https://trello.com/c/Wcuq3R1S/2167-add-ga4-page-view-tracking-to-apps-that-we-own


ollietreend (commented):
We might want to consider adding:

```ruby
gem "govuk_publishing_components"
```

to the Gemfile, since it's not currently declared as an explicit dependency.

Right now it's only in the project as a sub-dependency of Govspeak.

This is what Whitehall does: https://github.com/alphagov/whitehall/blob/6a2953a47e662095bb8e9cabee40930de0a9c302/Gemfile#L27

beccapearce:
Ah yes you did mention it! I'll add it in!

Thank you! I'll check the other PRs tomorrow as well

ryanb-gds (approved):

PR #2459, opened 2023-12-11
Add format validation for licence identifier

1pretz1:
We've previously been informed of a rendering issue that was caused by the publisher inputting random text into the licence identifier field. Previously there was only a uniqueness validator for the field which accepted any format, now we're validating that the format is correct.

The accepted format is XXX-Y-B or XXXX-Y-B where:
- X is a formality code (3 or 4 digits)
- Y is a country code (1 to 7, 1 digit)
- B is a number in sequence (if more than 1 'type' [quite rare], 1 digit)

## Screenshots

Screenshot 2024-02-23 at 15 23 52
Screenshot 2024-02-23 at 15 23 57

Trello:
https://trello.com/c/uFIvraIX/2288-validate-licence-identifer-format

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -54,6 +54,7 @@ en:
               link_and_identifier_exists: Enter either the website where users can apply for the licence or the licence identifier
             licence_transaction_licence_identifier:
               taken: This licence identifier is already in use
+              invalid: Enter a licence identifier. This is a 5 or 6 digit number
             licence_transaction_industry:
               blank: Select an industry
             licence_transaction_location:
--- a/app/models/licence_transaction.rb
+++ b/app/models/licence_transaction.rb
@@ -18,9 +18,13 @@ class LicenceTransaction < Document
 
   validates_with LinkOrIdentifierValidator
   validates :primary_publishing_organisation, presence: true
-  validates :licence_transaction_licence_identifier, licence_identifier_unique: true
   validates :licence_transaction_industry, presence: true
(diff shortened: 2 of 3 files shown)


config/locales/en.yml
```diff
@@ -54,6 +54,7 @@ en:
               link_and_identifier_exists: Enter either the website where users can apply for the licence or the licence identifier
             licence_transaction_licence_identifier:
               taken: This licence identifier is already in use
+              invalid: Enter a licence identifier. This is a 5 or 6 digit number
```
KludgeKML, line 57: Should this show the actual format (like: enter a license identifier in the format XXX-X-X or XXXX-X-X), and maybe something about where you get the id from?

KludgeKML (approved):
Approved with caveat that maybe the warning message needs to say where licence identifiers come from?

config/locales/en.yml
1pretz1: I had similar concerns, but me and Katherine agreed that the shorter prompt is probably fine. TL;DR: 
> I think the proposed message is probably fine considering the point that new codes / or possibly existing ones will need to be sourced from us anyway, and thus us explaining the exact structure still doesn’t help the user if they don’t know the exact code they need beforehand. I’ll make the change!

You can read the conversation in the [Trello card](https://trello.com/c/uFIvraIX/2288-validate-licence-identifer-format#comment-65d76482aa851240ec89ccbd).

PR #2528, opened 2024-02-16
Make changes to farming grant finder

beccapearce:
- change the summary
- remove forestry commission 🤔
- remove facets open and closed and funding types
- move land types up in the form
- remove some grant scheme options

There are two grants in the database which will need to be changed as they have the facets open and closed and funding types already submitted


ryanb-gds (approved):
Looks like some specialist finder config

PR #2539, opened 2024-02-27
Add new values to sub_categories facet for UTAAC

lauraghiorghisor-tw:
Adding sub-categories for facet `SAFEGUARDING VULNERABLE GROUPS`.

Had a look into `publishing-api` and `search-api`, in:
- `_specialist_document.jsonnet` [here](https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/shared/definitions/_specialist_document.jsonnet)
- `utaac_decision.json` [here](https://github.com/alphagov/search-api/blob/main/config/schema/elasticsearch_types/utaac_decision.json)
but UTAAC field values are not included, so there was no need for changes. 

Deployed the finder to integration and published a document and could see all the values rendering correctly in the finder.

[Trello card](https://trello.com/c/GPrTSvrK/2353-administrative-appeals-tribunal-decisions-update-to-finder-facets)

### DIFF
--- a/lib/documents/schemas/utaac_decisions.json
+++ b/lib/documents/schemas/utaac_decisions.json
@@ -1747,16 +1747,156 @@
           "value": "revisions-supersessions-and-reviews-supersession-incapacity"
         },
         {
+          "label": "Safeguarding vulnerable groups - Adequacy",
+          "value": "safeguarding-vulnerable-groups-adequacy"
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
I think this should do the trick. Thanks for picking this up 🙏

PR #2575, opened 2024-04-02
Replace Fog with aws-sdk-s3

theseanything:
This allows us to use short lived AWS credentials via authentication methods such as instance profiles or IRSA.

No functionality changes.

### DIFF
--- a/spec/controllers/document_list_export_request_controller_spec.rb
+++ b/spec/controllers/document_list_export_request_controller_spec.rb
@@ -1,32 +1,21 @@
 require "rails_helper"
 
 RSpec.describe DocumentListExportRequestController, type: :controller do
+  let(:stubbed_client) { Aws::S3::Client.new(stub_responses: true) }
   before do
     log_in_as_gds_editor
 
-    Fog.mock!
-    ENV["AWS_REGION"] = "eu-west-1"
-    ENV["AWS_ACCESS_KEY_ID"] = "test"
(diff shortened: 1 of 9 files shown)


spec/controllers/document_list_export_request_controller_spec.rb
```diff
@@ -4,36 +4,27 @@
   before do
     log_in_as_gds_editor
 
-    Fog.mock!
-    ENV["AWS_REGION"] = "eu-west-1"
-    ENV["AWS_ACCESS_KEY_ID"] = "test"
-    ENV["AWS_SECRET_ACCESS_KEY"] = "test"
-    ENV["AWS_S3_BUCKET_NAME"] = "test-bucket"
+    @stubbed_client = Aws::S3::Client.new(stub_responses: true)
```
sengi: optional: might be somewhat more robust to use rspec's [let](https://rspec.info/features/3-12/rspec-core/helper-methods/let/) rather than an instance var for this sort of thing. (I'm still new-ish to rspec though so feel free to disregard if you know better than I do!)

sengi (approved):
LGTM, thanks! This should let us finally get rid of the last of the long-lived IAM keys 🎉

ryanb-gds (approved):
Looks good to me

spec/controllers/document_list_export_request_controller_spec.rb
theseanything: No you're absolutely right - good catch!

PR #2588, opened 2024-04-12
Remove uk market conformity assessment bodies finder

lauraghiorghisor-tw:
The finder documents have all been unpublished. The finder itself was expected to be unpublished. That work was done in 2023, by using the specialist-publisher `unpublish` rake task. Nonetheless, the code was not removed from the `specialist-publisher`, `publlishing-api` or `search-api`, causing the finder to be accidentally published again (for example by running the rake task for publishing all finders).

Once the 3 PRs are reviewed, the finder will be unpublished from all  environments using the `unpublish` rake task, then the PRs will be merged in.

An associated feature for creating an internal note was developed for this finder. It has not been used for other finders and it is not a development priority or business request. It will be removed in [this PR](https://github.com/alphagov/specialist-publisher/pull/2590).

[Trello card](https://trello.com/c/cZLoKAE2/1506-close-finder-for-ukmcab)


baisa (commented):
I would probably not include removal of internal notes in this PR. This is document's functionality and even if this was used for UKMCAB finder only we should not treat is as a part of finder removal?

baisa (approved):

PR #2621, opened 2024-05-14
Remove pre-production flag

lauraghiorghisor-tw:
Removing the `pre-production` field. Upon merging, we will then be able to publish the finder using the `publish_finder` rake task. 
We want to do this in staging to get some confidence that the finder is working as expected, since integration is playing up.

We will not be publishing the finder in prod, and we will revert this change upon testing.

[Trello card](https://trello.com/c/e0KYS9mQ/2584-farming-grant-options-breadcrumb-not-showing-not-able-to-save-documents)


ryanb-gds (approved):
:shipit:

PR #2722, opened 2024-08-08
Consolidate specialist finder configuration into the schemas

ChrisBAshton:
There's no reason why we need to spread finder configs across both a schema JSON file _and_ an ActiveModel.
This PR migrates the following data out of the models and into the schemas, making it easier to add and edit finders, and reducing room for bugs/discrepancies:

- Taxons
- Primary publishing organisation
- Title
- Rename `slug` to `admin_slug`

Ideas for the future:

- Move `admin_slug` into the schemas. Easier said than done: attempted in 595eaca071e2528f1be9b92c966a2ee6d7313989 and 8cea1b547cfb8fbda209adfc54ef6e6be9c6ad46
- Format specific fields. Easier said than done: attempted in 52a62b87c222a9d36b66936c3ddd8569c981e227. There are a handful of other weird method definitions in the specialist document models which means we can't just blindly delete the initialise method.
- Move the specialist document models into their own directory. Easier said than done: attempted in https://github.com/alphagov/specialist-publisher/commit/dfe52f3f09a18ac2e22a016b859dcc5bdfc1073b

Trello: https://trello.com/c/Bw2sYhwE/2776-consolidate-finder-configs-in-specialist-publisher

---


app/lib/finder_schema.rb
```diff
@@ -6,15 +6,21 @@ def self.schema_names
     end
   end
 
-  attr_reader :base_path, :organisations, :format, :content_id, :editing_organisations
+  def self.load_schema_for(type)
+    JSON.parse(File.read(Rails.root.join("lib/documents/schemas/#{type}.json")))
+  end
+
+  attr_reader :base_path, :organisations, :editing_organisations, :taxons, :format, :content_id, :document_title
```
ryanb-gds: Might want to consolidate the schema attribute reader to here.

app/models/document.rb
```diff
@@ -185,6 +172,10 @@ def facet_options(facet)
 
   delegate :organisations, to: :finder_schema
 
+  def primary_publishing_organisation
+    organisations.first
```
ryanb-gds, line 176: Is this a safe assumption? Just double checking

ryanb-gds (approved):
Good work on this, definitely makes things more consistent. Pretty sure you'll have considered the primary publishing org selection so going to approve this on that assumption!

app/models/document.rb
ChrisBAshton: It's the truth for all the schema JSON files at the moment - I checked them 👍 only risk is if someone doesn't follow this approach in future.

But worth noting that #1230 introduces a feature for the "Statutory Instruments" finder to override the primary publishing organisation per-document. (Subsequently rolled out to FloodAndCoastalErosionRiskManagementResearchReport and LicenceTransaction too).

app/lib/finder_schema.rb
ChrisBAshton: Good idea.

app/models/document.rb
ChrisBAshton: Hmmm I'll take another look at that :/

ChrisBAshton: Just deployed the branch to integration and was able to change the primary publishing org on https://specialist-publisher.integration.publishing.service.gov.uk/flood-and-coastal-erosion-risk-management-research-reports/6993f8c2-8936-492c-a5c5-4ea4b7444f97:en - so I think we're ok.

PR #2733, opened 2024-08-23
Updating docs on editing field values.

minhngocd:
The docuemntation seems to only work for adding new values to the facet rather than amending existing ones, which I assume will also entail editing and republishing all existing documents. in adding new values to the facet, we should only need to update the finder, not the search API schema. This has been tested in integration.


ChrisBAshton (approved):

minhngocd:
Another PR is going to be updating the docs in the same area, lumping in the changes to that one
https://github.com/alphagov/specialist-publisher/pull/2743

PR #2736, opened 2024-08-27
Point users to GOV.UK docker documentation for running specialist finders locally

ryanb-gds:
Documentation is difficult enough to keep up to date without having duplicate copies around!


minhngocd (approved):
🙏

PR #2781, opened 2024-09-23
Rake task to allow bulk publishing that disables email sending

minhngocd:
Introduced a flag on the document model - this flag is used to override whether or not emails should be sent for a document upon publishing. 

We use this flag in the bulk publishing rake task so we can silence alerts when publishing in large amounts.

In the future could potentially it could be used within a self-serve UI so users can decide themselves whether or not to alert with an email during publishing. However this UI solution needs to consider the fact that the flag is not yet persisted within Publishing API, and documents are persisted within Publishing API, not Specialist Publisher.

https://trello.com/c/BRwLszUk/2841-investigate-email-alerts-for-ova-finder

### DIFF
--- a/lib/tasks/publish.rake
+++ b/lib/tasks/publish.rake
@@ -4,12 +4,22 @@ require "services"
 
 namespace :publish do
   desc "Publishes all documents of given types in a draft publication state.\n" \
-    "Usage: `rake publish:all[protected_food_drink_name,esi_fund]`"
+    "Usage: `rake publish:all[protected_food_drink_name esi_fund]`"
   task :all, [:document_types] => :environment do |_, args|
     types = args.document_types.presence&.split(" ")&.compact
 
-    raise ArgumentError, "No type given." if types.empty?
+    raise ArgumentError, "No type given." if types.blank?
 
     Publisher.publish_all(types:)
+  end
+
+  desc "Publishes all documents of given types in a draft publication state, but does not send any email alerts.\n" \
+         "Usage: `rake publish:all_silent[protected_food_drink_name,esi_fund]`"
+  task :all_silent, [:document_types] => :environment do |_, args|
+    types = args.document_types.presence&.split(" ")&.compact
+
+    raise ArgumentError, "No type given." if types.blank?
+
+    Publisher.publish_all(types:, disable_email_alert: true)
   end
 end
--- a/spec/models/document_spec.rb
+++ b/spec/models/document_spec.rb
@@ -833,4 +833,41 @@ RSpec.describe Document do
       expect(AaibReport.title).to eq("AAIB Report")
     end
   end
+
+  describe "#send_email_on_publish?" do
+    it "should send email if document is major update and no disable_email_alert override exists" do
+      doc = MyDocumentType.new(update_type: "major")
+      expect(doc.send_email_on_publish?).to be true
+    end
+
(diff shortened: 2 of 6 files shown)


spec/models/document_spec.rb
```diff
@@ -833,4 +833,34 @@ def initialize(params = {})
...
+      doc = MyDocumentType.new(update_type: "major")
+      doc.disable_email_alert = false
+      expect(doc.send_email_on_publish?).to be true
+    end
+
+    it "(Caution) will stop sending email if document is major update and disable_email_alert override is a random string" do
+      doc = MyDocumentType.new(update_type: "major")
+      doc.disable_email_alert = "invalid"
+      expect(doc.send_email_on_publish?).to be false
+    end
```
ChrisBAshton: Hmmm could we have `send_email_on_publish?` raise an exception, rather than semi-support using strings? Better for it to crash than to silently do something we're not expecting, IMO.

spec/models/document_spec.rb
```diff
@@ -833,4 +833,34 @@ def initialize(params = {})
...
+
+    it "should stop send email if document is major update and disable_email_alert override is true" do
+      doc = MyDocumentType.new(update_type: "major")
+      doc.disable_email_alert = true
+      expect(doc.send_email_on_publish?).to be false
+    end
+
+    it "should stop send email if document is minor update" do
```
ChrisBAshton: Is it worth adding a test to ensure that `update_type: minor` and `disable_email_alert: false` doesn't send an email?
(It wouldn't, but someone could probably swap the expressions in the `send_email_on_publish?` method and it might then send).

lib/tasks/publish.rake
```diff
@@ -4,12 +4,22 @@ require "services"
 
 namespace :publish do
   desc "Publishes all documents of given types in a draft publication state.\n" \
-    "Usage: `rake publish:all[protected_food_drink_name,esi_fund]`"
```
ChrisBAshton, line 7: Oof - good catch.

ChrisBAshton (approved):
Nice 👍 some non-blocking comments below.

spec/models/document_spec.rb
minhngocd: good point - i'll add that

spec/models/document_spec.rb
minhngocd: i was on the fence about this - thought it wasn't likely to happen since the rake task passes bools anyways, so should be enough to call out as a test - but yes it would be safer. Will add that in

spec/models/document_spec.rb
minhngocd: Have added a test for this

spec/models/document_spec.rb
minhngocd: Tried implementing an exception raising and it feels like it doesn't read so well. 
```
  def send_email_on_publish?
    raise "Invalid disable email alert flag. Please use booleans only." if disable_email_alert.present && ![true, false].include?(disable_email_alert)
    update_type == "major" && !disable_email_alert
  end
```

I'm not sure this is necessary as the usage is internal code and not exposed to any external API or users. I feel like as long as we document this behaviour in the test, it should be enough. Will leave that discussion for next time and amend as a separate PR if needed.

minhngocd: Discussion happened - putting it in

PR #2793, opened 2024-10-04
Order OVA finder alphabetically

ChrisBAshton:
Trello: https://trello.com/c/tB7aFZQB/2981-add-alphabetical-sorting-for-ova-finder

### DIFF
--- a/lib/documents/schemas/veterans_support_organisations.json
+++ b/lib/documents/schemas/veterans_support_organisations.json
@@ -18,6 +18,7 @@
   ],
   "document_noun": "result",
   "document_title": "Veterans Support Organisation",
+  "default_order": "title",
   "facets": [
     {
       "key": "veterans_support_organisation_health_and_social_care",


minhngocd (commented):
The default order property seems deprecated according to the finder.json schema

```
    "finder_default_order": {
      "description": "DEPRECATED: Use “sort” property",
      "type": "string"
    },
```

The sort property seems to be the preferred option: 

```
"finder_sort": {
      "description": "These are the options for sorting the finder",
      "type": "array",
      "items": {
        "type": "object",
        "required": [
          "key",
          "name"
        ],
        "additionalProperties": false,
        "properties": {
          "default": {
            "description": "Indicates if this should be the default sort option",
            "type": "boolean"
          },
          "key": {
            "description": "Rummager field name, with an optional preceding “-” to sort in descending order",
            "type": "string"
          },
          "name": {
            "description": "Label for the sort option",
            "type": "string"
          }
        }
```

so i guess something like 
```
"sort": [
      {
        "key": "title",
        "name": "A-Z",
        "default": true
      }
]
```

minhngocd (approved):

ChrisBAshton:
@minhngocd I tried deploying the non-deprecated version (https://github.com/alphagov/specialist-publisher/pull/2794) but it didn't work.

I think given none of the other finders use the `sort` key, and that there are five other live finders using the `default_order` key, let's just merge this one as-is and come back to it later.

PR #2810, opened 2024-10-16
Introduce design system

dnkrj:

- Mount publishing components component guide
- Configure the stylesheets with dartsass
- Prefix application JS, CSS and layout with legacy
- Introduce design system layout
- Add temporary page for design system with tests

https://trello.com/c/TFLJjnxx/2954-add-design-system-base-into-specialist-publisher

## Screenshot

specialist-publisher dev gov uk_design-system(iPad Pro) (1)


app/controllers/temp_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  layout "design_system"
+
+  before_action :check_authorisation
+
+  def index; end
+
+private
+
+  def check_authorisation
+    authorize AaibReport # ignore this
```
ChrisBAshton: Maybe tweak the comment here for something like:

```suggestion
    # A `Pundit::AuthorizationNotPerformedError` exception is raised
    # unless we call the `authorize` model against an existing model.
    # This will likely be swapped out for `authorize current_format` when
    # we remove the temporary page and add the admin form pages.
    authorize AaibReport
```

ChrisBAshton (approved):
Nice one 👏

PR #2816, opened 2024-10-18
Introduce form to edit finder metadata

dnkrj:

- Copy the multiselect autocomplete from Whitehall
- Add `node_modules` to the asset load path
- Add route and template for form to edit finder metadata
- Move metadata summary card to shared template
- Adjust metadata summary card remove email alerts and displaying related links
- Conditionally show metadata summary card fields
- Use metadata summary card for confirmation page
- Temporarily include the proposed JSON for metadata

### Form
specialist-publisher dev gov uk_admin_metadata_cma-cases(iPad Pro) (2)

### Confirmation page
specialist-publisher dev gov uk_admin_metadata_cma-cases(iPad Pro)

Trello: https://trello.com/c/IGg2kTG5/3015-create-edit-finder-form-basic-details


app/assets/javascripts/application.js
```diff
@@ -1,3 +1,5 @@
 //= require govuk_publishing_components/dependencies
 //= require govuk_publishing_components/lib
 //= require govuk_publishing_components/components/copy-to-clipboard
+
+//= require components/autocomplete
```
ChrisBAshton, line 5: Overall approach looks good so far - I'll hold off on a full review until the POST logic is implemented.

But while I'm here, there are two things I'm not keen on with this component import work:

1. That we're copying and pasting code rather than reusing a module
2. That the existing component is considered inaccessible and [ripe for replacement](https://trello.com/c/zLNLSKok/85-design-accessible-replacement-for-accessible-autocomplete-multiselect)

We could address 1 fairly easily, by publishing accessible-autocomplete-multiselect to NPM and pulling from there. I think we should at least do that IMO (we may decide to invest time in improving the component rather than replacing it, in which case we should have our improvements apply to both publishing apps).

But I think we should also discuss 2 and acknowledge in the commit message that it has issues but that it's our only option at the moment?

ChrisBAshton: UPDATE after dev catchup today:

1. Let's continue with this as-is (you're right that the main code for this is already NPM, this is just 'scaffold' around it. Would be nice to put this into Rubygems but you point out we don't have a precedent for individual components)
2. We'll look at prioritising bringing https://trello.com/c/zLNLSKok/85-design-accessible-replacement-for-accessible-autocomplete-multiselect into a sprint in the near future, to replace this component.

app/assets/javascripts/application.js
ChrisBAshton, line 1: Let's reference https://trello.com/c/zLNLSKok/85-design-accessible-replacement-for-accessible-autocomplete-multiselect in the commit message to make it clear this is somewhat temporary.

spec/features/viewing_the_admin_summary_for_cma_cases_spec.rb
```diff
@@ -21,6 +21,5 @@
     expect(page.find(".govuk-summary-list__row", text: "Should summary of each content show under the title in the finder list page?")).to have_selector("dt", text: "No")
     expect(page.find(".govuk-summary-list__row", text: "Organisations the finder should be attached to")).to have_selector("dt", text: "Competition and Markets Authority")
     expect(page.find(".govuk-summary-list__row", text: "The document noun (How the documents on the finder are referred to)")).to have_selector("dt", text: "Case")
-    expect(page.find(".govuk-summary-list__row", text: "Would you like to set up email alerts for the finder?")).to have_selector("dt", text: "Yes")
```
ChrisBAshton, line 24: Worth linking to the "email alerts form" Trello card in the commit message here?

app/views/admin/_metadata_summary_card.html.erb
```diff
@@ -2,45 +2,52 @@
...
 <%= render "govuk_publishing_components/components/summary_card", {
   title: "Finder details",
   rows: [
-    {
+    *([{
       key: "Title of the finder",
       value: schema["name"]
-    },
-    {
+    }] if !(defined? previous_schema) || schema["name"] != previous_schema["name"]),
```
ChrisBAshton: The `*([{` syntax really threw me 😅 Trying this out locally, I think we can achieve the same outcome by dropping the `*` and the `[`...

```ruby
(
  {
    key: "Title of the finder",
    value: schema["name"],
  } if !(defined? previous_schema) || schema["name"] != previous_schema["name"]
),
```

app/controllers/admin_controller.rb
```diff
@@ -7,6 +7,37 @@ def summary; end
...
+      organisations: [],
+      related: [],
+    )
+
+    @params[:organisations].reject!(&:empty?)
+    @params[:related].reject!(&:empty?)
+
+    @proposed_schema = @current_format.finder_schema.schema.merge(@params.to_unsafe_h)
+
+    @proposed_schema["signup_copy"] = "You'll get an email each time a #{@params[:document_noun]} is updated or a new #{@params[:document_noun]} is published."
```
ChrisBAshton, line 26: Is this old, and to be done as part of the email alerts card?

app/views/admin/_metadata_summary_card.html.erb
```diff
@@ -2,45 +2,52 @@
   summary_card_actions ||= []
 %>
 
+<% related_links_value = capture do %>
+  <% if schema["related"] %>
+    Yes
+    <% schema["related"].each_with_index do |content_id, index| %>
+      <p>Link <%= index + 1 %>: <%= content_id %></p>
```
ChrisBAshton, line 9: Could we have a screenshot of this (more complicated) outcome, rather than the "No" outcome you have in the PR description currently?

ChrisBAshton (approved):
No blocking comments - nice one ⭐

app/views/admin/_metadata_summary_card.html.erb
dnkrj: I've been trying locally and I don't think that you can construct an array with conditional elements like that.
But feel free to update the PR code if you have a working example.

app/controllers/admin_controller.rb
dnkrj: Since signup copy is is consistently to the `document_noun` field I felt like it should be updated here at the same time.

An alternative would be that we remove the field entirely, what do you think?

app/views/admin/_metadata_summary_card.html.erb
ChrisBAshton: It seems to work for me locally, unless I'm not checking the right screens? (FWIW I'm trying this on the AAIB finder, and have tried editing the title and slug):

https://github.com/alphagov/specialist-publisher/pull/2829

dnkrj: Ah, it was the `.compact` call that I was missing. Nice! That's much more readable.

app/controllers/admin_controller.rb
ChrisBAshton: Ooh interesting! Feels like we either:

1) keep it here, or
2) remove the field configurability altogether (automatically construct it instead), or
3) move it to the new 'email' form and duplicate the document noun inline (i.e. "You'll get an email each time a foo is updated or a new foo is published."

I think I'm leaning toward 2. I can imagine some publishers might accidentally abuse the signup copy otherwise?

dnkrj: Agreed, I've gone through and the only finder that doesn't quite match the proposed format is: `You’ll get an email each time a new UK Approved Recommendation is published.` and that's close enough I think.

PR #2830, opened 2024-10-29
Integrate the finder edit forms with support api

dnkrj:

- Update the confirm_metadata page to post the proposed schema to a new zendesk endpoint to open a ticket using the support API.
- Use the `success_alert` component to display `flash[notice]`s.
- Revert "Temporarily include the proposed JSON for metadata"

## Screenshot
specialist-publisher dev gov uk_admin_cma-cases(iPad Pro) (1)

Trello Ticket:
https://trello.com/c/YdGWsy0I/3035-integrate-the-specialist-finder-edit-forms-with-support-api

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -13,6 +13,7 @@ class AdminController < ApplicationController
       :base_path,
       :description,
       :summary,
+      :show_summaries,
       :document_noun,
       organisations: [],
       related: [],
@@ -29,7 +30,7 @@ class AdminController < ApplicationController
       @proposed_schema.delete("related")
     end
 
-    if @proposed_schema["show_summaries"] == "true"
+    if params["show_summaries"] == "true"
       @proposed_schema["show_summaries"] = true
     else
       @proposed_schema.delete("show_summaries")
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -38,6 +38,11 @@ class AdminController < ApplicationController
     render :confirm_metadata
   end
 
+  def zendesk
+    GdsApi.support_api.raise_support_ticket(support_payload)
+    redirect_to "/admin/#{current_format.admin_slug}", notice: "Your changes have been submitted and Zendesk ticket created."
+  end
+
 private
 
   def check_authorisation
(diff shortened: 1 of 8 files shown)


spec/features/editing_the_cma_case_finder_spec.rb
```diff
@@ -11,6 +14,7 @@
     log_in_as_editor(:cma_editor)
     stub_publishing_api_has_content([], hash_including(document_type: CmaCase.document_type))
     stub_publishing_api_has_content(organisations, hash_including(document_type: Organisation.document_type))
+    stub_any_support_api_call
```
ChrisBAshton, line 17: I think it's important to add a test here, that verifies what we're sending to Zendesk. I'd at least want to cover the requester email, the description, and the tag.

app/controllers/admin_controller.rb
```diff
@@ -48,4 +53,17 @@ def check_authorisation
       redirect_to root_path
     end
   end
+
+  def support_payload
+    {
+      subject: "Specialist Finder Edit Request: #{current_format.title}",
+      tags: %w[specialist_finder_edit_request],
```
ChrisBAshton, line 61: Awesome! For reference, I've just raised a [ticket with User Support](https://govuk.zendesk.com/agent/tickets/5957502) to set up the trigger rule that will route tickets with this tag, into the Publishing Service Support queue.

app/views/admin/confirm_metadata.html.erb
```diff
@@ -33,14 +33,6 @@
       previous_schema: @current_format.finder_schema.schema,
     } %>
 
-    <%= render "govuk_publishing_components/components/details", {
-      title: "Proposed JSON",
```
ChrisBAshton, line 37: Given the detail is hidden away, I wonder if it's at all useful to keep this in? Up to you.

ChrisBAshton (commented):
One blocking comment RE a test- but otherwise looks great ⭐

spec/features/editing_the_cma_case_finder_spec.rb
dnkrj: I've added a scenario to the admin controller spec to cover this since it was easier to verify that the api had actually be called correctly there.

app/views/admin/confirm_metadata.html.erb
dnkrj: I can see how it might be useful for us, but it also breaks the layout when it's opened.
I'm keen to remove it for now and if we find that we have a genuine need for it add a properly thought out solution.

ChrisBAshton (approved):
LGTM 👏

PR #2842, opened 2024-11-05
Point developer to the schema that needs updating

ChrisBAshton:
The Zendesk integration currently just outputs the schema contents that need copying and pasting. But the developer still needs to track down the file in specialist-publisher, which is an unnecessarily manual step - let's give them the link directly to the file edit screen, ready to copy and paste the schema.

Trello: https://trello.com/c/YdGWsy0I/3035-integrate-the-specialist-finder-edit-forms-with-support-api

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -58,7 +58,10 @@ private
       subject: "Specialist Finder Edit Request: #{current_format.title.pluralize}",
       tags: %w[specialist_finder_edit_request],
       priority: "normal",
-      description: "```\r\n#{params[:proposed_schema]}\r\n```",
+      description: "Developer - raise a PR replacing this schema with the schema below: " \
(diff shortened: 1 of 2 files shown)


app/controllers/admin_controller.rb
```diff
@@ -58,7 +58,10 @@ def support_payload
       subject: "Specialist Finder Edit Request: #{current_format.title.pluralize}",
       tags: %w[specialist_finder_edit_request],
       priority: "normal",
-      description: "```\r\n#{params[:proposed_schema]}\r\n```",
+      description: "Developer: copy and paste the following diff into " \
```
dnkrj:
```suggestion
      description: "Developer: Raise a PR replacing the following existing schema with the proposed schema " \
```

dnkrj (approved):
Maybe "diff" isn't the right word to use here

app/controllers/admin_controller.rb
ChrisBAshton: Agreed!

PR #2847, opened 2024-11-07
Add new facet to SFO finder

GDSNewt:
Adds an additional facet to the new to SFO Finder specialist finder. As per documentation found [here](https://docs.publishing.service.gov.uk/repos/specialist-publisher/creating-editing-and-removing-specialist-document-types-and-finders.html#editing-a-specialist-document-type)

[First PR](https://github.com/alphagov/specialist-publisher/pull/2835)
[Trello](https://trello.com/c/E7KMeGZo/3034-specialist-finder-sfo-cases)
[Zendesk](https://govuk.zendesk.com/agent/tickets/5958029)

### DIFF
--- a/app/models/sfo_case.rb
+++ b/app/models/sfo_case.rb
@@ -1,8 +1,10 @@
 class SfoCase < Document
   validates :sfo_case_state, presence: true
+  validates :sfo_case_opened_date, presence: true, date: true
 
   FORMAT_SPECIFIC_FIELDS = %i[
     sfo_case_state
+    sfo_case_opened_date
   ].freeze
 
   attr_accessor(*FORMAT_SPECIFIC_FIELDS)
--- a/lib/documents/schemas/sfo_cases.json
+++ b/lib/documents/schemas/sfo_cases.json
@@ -34,6 +34,15 @@
           "value": "closed"
         }
       ]
+    },
+    {
+      "key": "sfo_case_opened_date",
+      "name": "Opened date",
+      "short_name": "Opened date",
+      "type": "date",
+      "preposition": "opened",
+      "display_as_result_metadata": false,
+      "filterable": true
     }
   ]
 }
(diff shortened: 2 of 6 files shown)


app/models/sfo_case.rb
```diff
@@ -1,8 +1,10 @@
 class SfoCase < Document
   validates :sfo_case_state, presence: true
+  validates :sfo_case_opened_date, allow_blank: true, date: true
```
lauraghiorghisor-tw: Can you please check with the users if they want to have this as a required field or not? It was not clear to me that it's optional from the Zendesk ticket.

lib/documents/schemas/sfo_cases.json
```diff
@@ -34,6 +34,14 @@
           "value": "closed"
         }
       ]
+    },
+    {
+      "key": "sfo_case_opened_date",
```
lauraghiorghisor-tw, line 39: Similarly, include these checks in the question to the user, around the 
- display metadata
- filterable
- preposition (this was missing from the schema, not sure what it defaults to if you don't include it)

lauraghiorghisor-tw (changes requested):
LGTM 

Few bits to check with the user.

app/models/sfo_case.rb
GDSNewt: Have checked and they want it to be mandatory, have amended the validation

lib/documents/schemas/sfo_cases.json
GDSNewt: Have checked with the user on the filterable and metadata issue. The preposition defaults to 'Related to'. In this case "Related to after(date)" and "Related to before (date)". 

I will change it to make it clearer it relates to cases opened

lib/documents/schemas/sfo_cases.json
```diff
@@ -34,6 +34,15 @@
...
+      "name": "Opened date",
+      "short_name": "Opened date",
+      "type": "date",
+      "preposition": "opened",
+      "display_as_result_metadata": true,
```
lauraghiorghisor-tw: Reading the user's comment, I think this should be false?

lauraghiorghisor-tw (approved):

PR #2878, opened 2024-11-27
Email alerts configuration form - part 1

ryanb-gds:
This includes just a simple yes/no question for email signup opt-in. I will add the more complex functionality for external signup links and email facets in a separate PR.

This PR introduces a finder schema model, based on the existing finder schema object that merely decorated a hash object. Given the amount of changes, I decided not to do any further refactoring in this PR in order to keep it reviewable. Future PRs will move more logic from the admin controller on to the finder schema model. I think we will also want a finder facet model, as much of the logic around facets on the existing finder schema class is poorly factored.

## Screenshot

Screenshot 2024-11-27 at 14 46 02

### DIFF
--- a/app/models/finder_schema.rb
+++ b/app/models/finder_schema.rb
@@ -1,4 +1,8 @@
 class FinderSchema
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include ActiveModel::Serializers::JSON
+
   # Pluralized names of all document types
   def self.schema_names
     Dir.glob(Rails.root.join("lib/documents/schemas/*.json")).map do |f|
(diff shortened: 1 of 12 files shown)


app/views/admin/_metadata_summary_card.html.erb
```diff
@@ -18,36 +18,40 @@
   rows: [
     ({
       key: "Title of the finder",
-      value: schema["name"]
-    } if !(defined? previous_schema) || schema["name"] != previous_schema["name"]),
+      value: schema.name
+    } if !(defined? previous_schema) || schema.name != previous_schema.name),
```
ryanb-gds, line 22: Going to see if I can figure out a better way of doing this using "[dirty](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html)" attributes

app/models/finder_schema.rb
```diff
@@ -1,33 +1,69 @@
...
   end
 
-  def self.load_schema_for(type)
-    JSON.parse(File.read(Rails.root.join("lib/documents/schemas/#{type}.json")))
+  def self.load_from_schema(type)
+    new.from_json(File.read(Rails.root.join("lib/documents/schemas/#{type}.json")))
+  end
+
+  attr_writer :editing_organisations, :facets, :organisations, :show_summaries, :taxons
+  attr_accessor :base_path,
```
ryanb-gds, line 18: Using `ActiveModel::Model` requires us to specify any attributes we want to use, which is slightly annoying. I considered trying to override some methods to avoid this requirement, but in the end I thought it would be simpler just to list the attributes. There aren't so very many.

ryanb-gds (commented):
Couple of thoughts

dnkrj (approved):

PR #2889, opened 2024-12-02
Remove signup title schema attribute

ryanb-gds:
The signup title schema attribute was only being used for one finder. If a signup title is not provided, the title of the finder is used instead. For the finder in question, the finder title value was the same as the signup title value provided.

We are therefore removing the signup title schema attribute so that we do not have to support it within the finder configuration editing form that we are building.

### DIFF
--- a/app/models/finder_schema.rb
+++ b/app/models/finder_schema.rb
@@ -36,7 +36,6 @@ class FinderSchema
                 :signup_content_id,
                 :signup_copy,
                 :signup_link,
-                :signup_title,
                 :subscription_list_title_prefix,
                 :summary,
                 :target_stack,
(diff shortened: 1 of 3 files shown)


lauraghiorghisor-tw:
If, in the future, we make a manual commit to add in this field (in response to a _very_ specific user request), does that affect how the edit form behaves? I assume it will just ignore the existence of that field altogether?

What I'm trying to get to is: if we remove something from our "offering" on the finder (basically how custom it can be, even for obscure fields like this), it might perhaps make sense for that removal to extend all the way to finder frontend/ be driven from needs of the frontend. Otherwise you end up with out of sight fields in the code that may not be captured in the UI, which may be problematic regardless of who is using the forms, devs or end users. I am assuming we would be using the form moving forward (primarily) but that's not to say manual rogue commits can't be made.

lauraghiorghisor-tw (approved):
LGTM but I left a comment on removing existing fields.

PR #2905, opened 2024-12-06
Email alerts configuration form

ryanb-gds:
I'm not especially happy with how this has turned out, felt like I spent a lot of time fighting with Active Model to get the behaviour right. Sort of wonder if we should have just stuck with a plain Hash object to represent the finder schemas, but I've already spent far too long on this to start again now.

There's still a bit of work outstanding to make sure that external email signups are handled correctly by the metadata summary card but I want to make that a separate PR as I intend to refactor the summary card to a ViewComponent, as we will need more logic than I care to put in an erb template.

Worth reading the commit messages, they hopefully explain some of the more odd-looking code.

Trello: https://trello.com/c/l9XfLavV

### DIFF
--- a/app/components/email_alert_fieldset_component.html.erb
+++ b/app/components/email_alert_fieldset_component.html.erb
@@ -3,7 +3,7 @@
   heading_level: 2,
   heading_size: "m",
 } do %>
-  <%= render "govuk_publishing_components/components/radio", {
+  <% render "govuk_publishing_components/components/radio", {
     heading: "Would you like to set up email alerts for the finder?",
     heading_size: "s",
     hint: "Email alerts should only be added if you've tested a user need for these",
--- a/app/components/email_alert_fieldset_component.html.erb
+++ b/app/components/email_alert_fieldset_component.html.erb
@@ -11,8 +11,21 @@
     items: [
       {
         value: "all_content",
-        text: "Yes. Allow users to sign up to the entire finder (all content)",
-        checked: @schema.signup_content_id
+        text: "Yes. Allow users to sign up to email alerts for everything published on the finder.",
+        checked: @schema.signup_content_id,
+        conditional: render_all_content_condition_inputs
+      },
+      {
+        value: "filtered_content",
+        text: "Yes. Allow users to sign up by specific filter criteria",
+        checked: @schema.email_filter_by,
+        conditional: render_filtered_content_condition_inputs
+      },
+      {
+        value: "external",
+        text: "Yes. Allow users to sign up for updates using a signup link.",
+        checked: @schema.signup_link,
+        conditional: render_external_condition_inputs
       },
       {
         value: "no",
--- a/app/components/email_alert_fieldset_component.rb
+++ b/app/components/email_alert_fieldset_component.rb
@@ -2,4 +2,57 @@ class EmailAlertFieldsetComponent < ViewComponent::Base
   def initialize(schema:)
     @schema = schema
   end
+
+  def render_all_content_condition_inputs
+    signup_content_id_input = render("govuk_publishing_components/components/input", {
+      type: "hidden",
+      name: "signup_content_id",
+      value: @schema.signup_content_id || SecureRandom.uuid,
+    })
+
+    # Do not render the title prefix input if the value is a hash because we don't want to override the existing the existing
+    # values in such cases. We are going to revisit this later, but for now we only allow users to override this setting
+    # for finders with a single string value. Trello card for future work: https://trello.com/c/Qe8wOpaw
+    if @schema.subscription_list_title_prefix.is_a?(Hash)
+      signup_content_id_input
+    else
+      render("govuk_publishing_components/components/input", {
+        label: {
+          text: "Email subscription topic",
+        },
+        name: "subscription_list_title_prefix",
+        value: @schema.subscription_list_title_prefix,
+      }) + signup_content_id_input
+    end
+  end
+
+  # We are handling changes to email facet filters offline at present.
+  def render_filtered_content_condition_inputs
+    render("govuk_publishing_components/components/input", {
+      type: "hidden",
+      name: "signup_content_id",
+      value: @schema.signup_content_id || SecureRandom.uuid,
+    }) +
+      render("govuk_publishing_components/components/checkboxes", {
+        name: "email_filter_by",
+        heading: "Selected filter: #{@schema.email_filter_by&.humanize}",
+        items: [
+          {
+            label: "Make changes to this filter criteria (The development team will reach out to you to discuss the requirements to allow users to sign up by specific filter criteria)",
+            value: "CHANGE_REQUESTED",
+          },
+        ],
+      })
+  end
+
+  def render_external_condition_inputs
+    render("govuk_publishing_components/components/input", {
+      label: {
+        text: "Signup link",
+      },
+      hint: "A link to an email signup page",
+      name: "signup_link",
+      value: @schema.signup_link,
+    })
+  end
 end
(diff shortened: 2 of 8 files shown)


app/models/finder_schema.rb
```diff
@@ -14,74 +20,73 @@ def self.load_from_schema(type)
...
+      email_filter_by: nil,
+      email_filter_facets: nil,
+    )
   end
 
-  def related=(value)
-    @related = value.nil? ? nil : value.reject(&:empty?)
+  def override_signup_copy
+    if signup_copy.present?
+      _assign_attribute(:signup_copy, "You'll get an email each time a #{document_noun} is updated or a new #{document_noun} is published.")
```
ryanb-gds, line 80: If anyone can find a less ugly way of doing this, please holler

spec/models/finder_schema_spec.rb
```diff
@@ -25,16 +25,17 @@
     end
   end
 
-  describe "#base_path" do
```
ryanb-gds, line 28: I decided to drop a good number of these tests as now that we are using Active Model attributes it felt like we were testing Rails. I have retained tests for all of our own code.

app/components/email_alert_fieldset_component.html.erb
```diff
@@ -3,16 +3,29 @@
   heading_level: 2,
   heading_size: "m",
 } do %>
-  <%= render "govuk_publishing_components/components/radio", {
```
ChrisBAshton, line 6: Do we not need this character?

app/components/email_alert_fieldset_component.rb
```diff
@@ -2,4 +2,57 @@ class EmailAlertFieldsetComponent < ViewComponent::Base
...
+  def render_all_content_condition_inputs
+    signup_content_id_input = render("govuk_publishing_components/components/input", {
+      type: "hidden",
+      name: "signup_content_id",
+      value: @schema.signup_content_id || SecureRandom.uuid,
+    })
+
+    # Do not render the title prefix input if the value is a hash because we don't want to override the existing the existing
+    # values in such cases. We are going to revisit this later, but for now we only allow users to override this setting
+    # for finders with a single string value.
```
ChrisBAshton: Would be nice to link to a Trello card in this comment (even if it's only a placeholder card for now).

app/components/email_alert_fieldset_component.rb
```diff
@@ -2,4 +2,57 @@ class EmailAlertFieldsetComponent < ViewComponent::Base
...
+      name: "signup_content_id",
+      value: @schema.signup_content_id || SecureRandom.uuid,
+    }) +
+      render("govuk_publishing_components/components/checkboxes", {
+        name: "email_filter_by",
+        heading: "Selected filter: #{@schema.email_filter_by&.humanize}",
+        items: [
+          {
+            label: "Make changes to this filter criteria (The development team will reach out to you to discuss the requirements to allow users to sign up by specific filter criteria)",
+            value: "CHANGE_REQUESTED",
```
ChrisBAshton, line 42: Nice.

spec/components/email_alert_fieldset_component_spec.rb
```diff
@@ -15,11 +15,69 @@
...
+    schema = FinderSchema.new
+    schema.signup_content_id = "123"
+    schema.subscription_list_title_prefix = "Finder email subscription"
+    render_inline(described_class.new(schema:))
+
+    expect(page).to have_text("Email subscription topic")
+    expect(page).to have_field("subscription_list_title_prefix", with: schema.subscription_list_title_prefix)
+  end
+
+  it "does not render the email subscription topic if the all_content value is checked but the current schema value is a hash" do
```
ChrisBAshton, line 55: Worth adding a similar comment to what you added earlier, to imply the test is temporary and can be deleted one day?

app/controllers/admin_controller.rb
ChrisBAshton, line 1: > The conditional field feature in the GOV.UK radio button component does not prevent the inclusion of the field values in the form submission, because it only hides the conditional DOM using CSS.

That's a shame - feels like a missing feature of the [checkbox](https://components.publishing.service.gov.uk/component-guide/checkboxes) component. Might be worth raising an issue on govuk-frontend? I [couldn't find an existing one](https://github.com/alphagov/govuk-frontend/issues?q=is%3Aissue+checkbox)

app/models/finder_schema.rb
```diff
@@ -2,6 +2,12 @@ class FinderSchema
   include ActiveModel::Model
   include ActiveModel::Attributes
   include ActiveModel::Serializers::JSON
+  extend ActiveModel::Callbacks
+
+  define_model_callbacks :update
+
+  before_update :reset_email_alerts
```
ChrisBAshton, line 9: Blocker: wouldn't this also reset the email alerts when we _don't_ want to reset them?
I.e. from the "edit metadata" or "edit filters/facets" forms?

app/models/finder_schema.rb
ChrisBAshton: Worth specifying in the commit message the various things you already tried (`self.<attribute_name> =`, etc). From dev huddle, sounds like you've already tried everything I would have, so no suggestions from me!

ChrisBAshton (commented):
Looks good - I have one blocking comment, probably just me misunderstanding though. 🤞

app/components/email_alert_fieldset_component.html.erb
ryanb-gds: Good point, that's actually a sneaky bug fix. I forgot to separate it into it's own commit

app/controllers/admin_controller.rb
ryanb-gds: I think this is more a feature of browser behaviour. The frontend team could remove the fields from the DOM via Javascript, but I suspect that might impact accessibility.

app/models/finder_schema.rb
ryanb-gds: Yes - those other forms don't exist yet do they?

If/when they do we'll have to find another approach. Two spring to mind:

1. exclude these attributes when copying the attributes from the old schema into the proposed one. I didn't do this at first because I didn't want to put more logic into the controller. I don't feel strongly about that though
2. create some kind of update context option along the lines of Rails validation contexts to control which callbacks are applied

Any thoughts? There are probably other options as well

app/components/email_alert_fieldset_component.html.erb
ryanb-gds: Separated it out

app/models/finder_schema.rb
ChrisBAshton: The first form already exists - see https://specialist-publisher.publishing.service.gov.uk/admin/metadata/aaib-reports

Of the two proposals, the second sounds best IMO.

ryanb-gds: That's the same form I'm working on though (you'll notice the Email Alerts section with the legend in the wrong place at the bottom 🤦 )

Appreciate we'll have to do the work anyway when we come to filters and facets - happy to do it now if you'd prefer

ChrisBAshton: Oh! I could have sworn we'd decided to make the email-signup-editing its own form. I guess not (or the designs have since been updated!).

Less of a reason to do it now if it's not going to break any existing forms. Just so long as we remember to bear it in mind for the facets form 👍 or you could do it now to reduce the risk of forgetting it.

ryanb-gds: I suggest we make it a later if that's ok - no guarantees that we're going to get to filters and facets this quarter and I think it's going to require quite a lot of change to the finder schema model anyway, so this may need a rethink

ChrisBAshton (approved):

PR #2906, opened 2024-12-06
Derive field label from its 'name' property in the finder schema

ChrisBAshton:
Whilst developing a finder recently, I was surprised that my 'name' wasn't having any effect on what was showing in the admin UI. I believe until now, the 'name' is only used in finder-frontend itself.

The label was derived from the key instead of the name, but this often wasn't suitable, meaning we had a lot of `label:` overrides (as keys had to be namespaced, but labels can afford to use more generic language).

I've tweaked the templates to derive the label from the 'name' instead, which is already a curated human-readable name for the thing - so we no longer need to try to guess from the key. We still have the freedom to use explicit label overrides where needed (indeed, I've left several untouched in the _licences.html.erb file, as the name was quite different to what was configured in that view).

Trello: https://trello.com/c/5RZhjdWe/3211-specialist-finder-for-the-data-ethics-guidance

---


ChrisBAshton:
Superseded by 2930.

PR #2923, opened 2024-12-19
IPO trademark decisions finder

kashifatcha:
This PR implements a new feature - nested facets - and uses the IPO trademark decision finder's config to test the implementation. The current version includes all finder requirements, and is ready to be released as draft, alongside its other PRs:  [publishing-api](https://github.com/alphagov/publishing-api/pull/3054) and [search-api](https://github.com/alphagov/search-api/pull/3105).

At the core of the implementation is the decision to tag a document with both facet and subfacet, in order to minimise the changes required in publishing-api and search-api. Details can be found in the [associated ADR](https://github.com/alphagov/specialist-publisher/pull/2923/files#diff-c27b1375e88de653fd2f7419af66ef74f1736157af812758a5c4f15ae4d76a9e). The tagging has been tested in integration, including that search-api returns the correct tags. This [query](https://www.staging.publishing.service.gov.uk/api/search.json?filter_format=trademark_decision&fields=trademark_decision_grounds_sub_section&fields=trademark_decision_grounds_section) returns the subfacet, as expected: 
```
{
results: [
{
trademark_decision_grounds_section: [
"section-3-1-graphical-representation",
"section-3-4-and-section-3-5-other-issues"
],
trademark_decision_grounds_sub_section: [
"section-3-1-graphical-representation-is-it-graphically-represented",
"section-3-4-and-section-3-5-other-issues-protected-emblems-etc"
],
index: "govuk",
es_score: null,
_id: "/trademark-decisions/new-td1",
elasticsearch_type: "trademark_decision",
document_type: "trademark_decision"
}
]...}
```

[Upcoming work ](https://trello.com/c/skH2B2fH/3373-nested-facets-finder-frontend)will address Finder Frontend filtering. 

We're also using a feature that is not very commonly used, found elsewhere on [this finder](https://www.gov.uk/find-licences)  - to help search long lists of facet options. This will be used in other upcoming finders, and we've proven that it works successfully here (required setting `show_option_select_filter` in specialist publisher, and the field type as `searchable_multivalued_text` in search api).

Specialist Publisher Trademark Decisions view:
> Screenshot 2025-02-17 at 09 53 54

Formed filled out:
!Note: we only allow selecting a child if the parent has any children. Whilst this is not the case for this finder config, if a nested facet value did not have any children, the parent value would appear as a select option.
> Screenshot 2025-02-20 at 09 54 21

Summary page:
> Screenshot 2025-02-20 at 09 54 30

Document preview
!Note: This is quite verbose on the nested values - we will discuss a different approach.
> Screenshot 2025-02-20 at 09 55 01

Finder summary
> Screenshot 2025-02-20 at 09 56 04

Finder
> Screenshot 2025-02-20 at 09 55 56

Finder Date filter
Screenshot 2025-02-17 at 09 59 20

Searchable filter
> Screenshot 2025-02-17 at 11 35 20

Free text search
- Mark
> Screenshot 2025-02-20 at 10 55 19
- Company
> Screenshot 2025-02-20 at 10 55 24

Nested facets filter
!Note: nested facets are not currently rendering on the Front End, so we're only rendering parent options
> Screenshot 2025-02-20 at 09 57 26

[Trello nested facets](https://trello.com/c/4RJusHld/3374-nested-facets-add-new-document)
[Trello IPO finder](https://trello.com/c/OA4zE8m0/3122-ipo-finder-request)


app/controllers/documents_controller.rb
```diff
@@ -94,6 +91,22 @@ def discard
 
 private
 
+  def merge_nested_facet_fields
+    return unless @document
+
+    @document.finder_schema.nested_facets.pluck("key", "sub_facet_key").each do |facet_key, sub_facet_key|
+      next unless @document.send(sub_facet_key)
```
ryanb-gds, line 98: Would `respond_to?` be more semantically appropriate here?

app/models/document.rb
```diff
@@ -81,8 +81,10 @@ def finder_schema
   def set_attributes(attrs, keys = nil)
     keys ||= attrs.keys
     keys.each do |key|
-      public_send(:"#{clean_key(key.to_s)}=", param_value(attrs, key))
+      assign_attributes(clean_key(key) => param_value(attrs, key))
```
ryanb-gds, line 84: Nice tidy up

app/models/document.rb
```diff
@@ -386,6 +395,35 @@ def self.apply_validations
...
+          nested_facet_value << value
+        else
+          parent_facet_value << value
+        end
+      end
+
+      assign_attributes(clean_key(nested_facet_from_schema["key"]) => parent_facet_value.uniq) if parent_facet_value.any?
+      assign_attributes(clean_key(nested_facet_from_schema["sub_facet_key"]) => nested_facet_value.uniq) if nested_facet_value.any?
+    end
+  end
```
ryanb-gds, line 425: I think at this point it's probably worth creating a facet model to abstract some of this away from the document [like we did for email alerts](https://github.com/alphagov/specialist-publisher/blob/main/app/models/email_alert.rb), but we can maybe pair on that in a separate PR if anyone is interested? Might also save us from all the merging in the controller, not sure about that though

spec/controllers/documents_controller_spec.rb
```diff
@@ -41,4 +41,105 @@
...
+  describe "POST create" do
+    let(:stub_content_id) { "test-content-id" }
+
+    before do
+      stub_any_publishing_api_put_content
+      stub_any_publishing_api_patch_links
+      expect(SecureRandom).to receive(:uuid).and_return(stub_content_id)
+    end
+
+    it "should create a document and send draft to Publishing API" do
```
ryanb-gds, line 54: Bit worrying that this test was missing!

spec/controllers/documents_controller_spec.rb
```diff
@@ -41,4 +41,105 @@
...
+          "trademark_decision_person_or_company_involved" => "",
+          "trademark_decision_grounds_section" => [
+            "",
+            "section-3-1-graphical-representation-is-it-graphically-represented",
+            "section-3-1-graphical-representation-is-it-capable-of-distinguishing",
+            "section-3-3-immoral-and-deceptive-marks-deceptive-as-to-nature-quality-etc",
+          ],
+        },
+        "save" => "",
+        "document_type_slug" => "trademark-decisions",
```
ryanb-gds, line 115: I wonder if we should create ourselves a test specialist finder that never gets published. Would be nice not to have to use actual finders in our generic tests.

app/components/document_metadata_component.rb
```diff
@@ -0,0 +1,31 @@
+class DocumentMetadataComponent < ViewComponent::Base
```
ryanb-gds, line 1: I like that this is a view component now 🙌

ryanb-gds (approved):
Good stuff, couple of minor suggestions but nothing that should stop merging if you're wanting to 🚢 

Edit: well, except the failing example 😅

spec/controllers/documents_controller_spec.rb
lauraghiorghisor-tw: We do cover this in feature tests too, but it's handy to have a controller test, closer to the code.

lauraghiorghisor-tw: Maybe it makes sense for the feature tests to stay more UI-focused, and bring the publishing API expectations in controller tests.

spec/controllers/documents_controller_spec.rb
lauraghiorghisor-tw: I like this, for the next person 😅

app/controllers/documents_controller.rb
minhngocd: unfortunately no, because respond_to? would return true, but the actual value of it may be blank or nil

minhngocd: alternatively, i've been changing the `public_send`s to `assign_attributes` so i'm sure we can change it to `get_attribute` of some sort

app/models/document.rb
minhngocd: sounds good! yes, maybe we can do that as a separate thing

PR #2929, opened 2024-12-27
Create 'edit finder' form: "Filters and options"

ChrisBAshton:
Supersedes #2840. This PR:

- Add summary card for facets, displaying them on the admin page
- Introduces route and form to modify the facets for a finder
- Introduces page to confirm changes to the facets of a finder
- Introduces new "Facet" model to keep logic out of the controller

Trello: https://trello.com/c/KNf2zNFI/3037-create-edit-finder-form-filters-and-options

## Screenshot

Summary page:

specialist-publisher dev gov uk_admin_cma-cases(iPad Pro)

Edit facets page:

Screenshot 2025-01-03 at 16 20 32

Confirmation page:

Screenshot 2025-01-03 at 16 21 15

With expanded diff:

Screenshot 2025-01-03 at 16 21 39

With expanded schema:

Screenshot 2025-01-03 at 16 22 02

## Caveats

1. The form doesn't work without JS.

2. The question "Do publishers need to select an option...?" in the design is blocked on [moving validation logic into the schema](https://trello.com/c/PArCm42y/3292-dry-up-specialist-publisher-validation), so has been omitted for the time being.

3. All properties [in the finder.jsonnet](https://github.com/alphagov/publishing-api/blob/fb5200249a7e44ac9e6ca84be87ba96586889b16/content_schemas/formats/shared/definitions/finder.jsonnet#L146-L254) are supported, with the exception of the following barely-used filter properties:

- `show_option_select_filter`  ([used only in licence_transactions.json](https://github.com/alphagov/specialist-publisher/blob/fd1c76d40820b3311bf420634cb9615092cb1263/lib/documents/schemas/licence_transactions.json#L56))
- `large` ([used only in licence_transactions.json](https://github.com/alphagov/specialist-publisher/blob/fd1c76d40820b3311bf420634cb9615092cb1263/lib/documents/schemas/licence_transactions.json#L59))
- `open_on_load` ([used only in licence_transactions.json](https://github.com/alphagov/specialist-publisher/blob/fd1c76d40820b3311bf420634cb9615092cb1263/lib/documents/schemas/licence_transactions.json#L60))

---

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -5,6 +5,8 @@ class AdminController < ApplicationController
 
   def summary; end
 
+  def edit_facets; end
+
   def edit_metadata; end
 
   def confirm_metadata
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -9,6 +9,21 @@ class AdminController < ApplicationController
 
   def edit_metadata; end
 
+  def confirm_facets
+    @params = facets_params
+    @params["facets"] = @params["facets"].values.map { |facet_params|
+      next if facet_params["_destroy"] == "1"
+
+      Facet.from_finder_admin_form_params(facet_params)
+           .to_finder_schema_attributes
+    }.compact
+
+    @proposed_schema = FinderSchema.new(@current_format.finder_schema.attributes)
+    @proposed_schema.update(@params)
+
+    render :confirm_facets
+  end
+
   def confirm_metadata
     @params = params.permit(
       :name,
@@ -75,6 +90,23 @@ private
       :filtered_content_list_title_prefix,
       :email_filter_by,
       :signup_link,
+    )
+  end
+
+  def facets_params
+    allowed_facet_params = %i[
+      key
+      name
+      short_name
+      type
+      preposition
+      display_as_result_metadata
+      filterable
+      allowed_values
+      _destroy
+    ]
+    params.permit(
+      facets: allowed_facet_params,
     )
   end
 end
--- a/app/models/facet.rb
+++ b/app/models/facet.rb
@@ -1,17 +1,16 @@
 class Facet
   include ActiveModel::Model
+  include ActiveModel::Attributes
 
-  attr_accessor(
-    :key,
-    :name,
-    :short_name,
-    :type,
-    :preposition,
-    :display_as_result_metadata,
-    :filterable,
-    :allowed_values,
-    :specialist_publisher_properties,
-  )
+  attribute :key
+  attribute :name
+  attribute :short_name
+  attribute :type
+  attribute :preposition
+  attribute :display_as_result_metadata, :boolean
+  attribute :filterable, :boolean
+  attribute :allowed_values
+  attribute :specialist_publisher_properties
 
   def to_finder_schema_attributes
     {
@@ -35,8 +34,8 @@ class Facet
       facet.short_name = nil_if_blank(params["short_name"])
       facet.type = facet_type(params["type"])
       facet.preposition = nil_if_blank(params["preposition"])
-      facet.display_as_result_metadata = str_to_bool(params["display_as_result_metadata"])
-      facet.filterable = str_to_bool(params["filterable"])
+      facet.display_as_result_metadata = params["display_as_result_metadata"]
+      facet.filterable = params["filterable"]
       facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"])
       facet.specialist_publisher_properties = facet_specialist_publisher_properties(params["type"])
       facet
@@ -50,14 +49,6 @@ class Facet
 
     def nil_if_blank(str)
       str.presence
-    end
-
-    def str_to_bool(str)
-      if str == "true"
-        true
-      elsif str == "false"
-        false
-      end
     end
 
     def facet_type(type)
--- a/app/models/facet.rb
+++ b/app/models/facet.rb
@@ -1,0 +1,92 @@
+class Facet
+  include ActiveModel::Model
+
+  attr_accessor(
+    :key,
+    :name,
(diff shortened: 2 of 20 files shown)


app/controllers/admin_controller.rb
```diff
@@ -75,4 +92,21 @@ def email_alert_params
       :signup_link,
     )
   end
+
+  def facets_params
+    allowed_facets = %i[
```
ryanb-gds: Suggesting slightly more specificity - `allowed_facets` to me suggests we're specifying what facets are allowed rather than what attributes a facet can have:
```suggestion
    allowed_facet_params = %i[
```

app/models/facet.rb
```diff
@@ -0,0 +1,92 @@
...
+      str.presence
+    end
+
+    def str_to_bool(str)
+      if str == "true"
+        true
+      elsif str == "false"
+        false
+      end
+    end
```
ryanb-gds: You could use Active Model for casting ([docs](https://guides.rubyonrails.org/active_model_basics.html#attributes)) if you were up for changing your attribute accessors to full Active Model attribute definitions. I did do this for the finder schema and was a bit equivocal about the outcome, appreciate you may decide/have decided already it's not worth the overhead.

app/models/facet.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    def facet_type(type)
+      facet_types_that_allow_enum_values.include?(type) ? "text" : type
+    end
+
+    def facet_allowed_values(values, type)
+      return nil if values.nil? || facet_types_that_allow_enum_values.exclude?(type)
```
ryanb-gds, line 59: I do wonder if we should be using [validation](https://guides.rubyonrails.org/active_model_basics.html#validations) here rather than ignoring invalid values, but can understand why we might choose not to. I can't remember what I did for the email facets in the end - I think I might have been lazy and assumed valid input because of the conditional behaviour making it difficult to provide invalid data, but of course that's not really very good because never trust client input/user could have JS disabled etc.

app/views/admin/_facet_fields.html.erb
```diff
@@ -0,0 +1,110 @@
...
+
+<%= render "govuk_publishing_components/components/textarea", {
+  label: {
+    text: "Filter options ('Multiple select' or 'One option' only)",
+    heading_size: "s",
+  },
+  hint: sanitize("Put each option on a new line. The underlying name for existing (live) options will appear in curly braces: please don't edit these, and don't add them for new options (they'll be created automatically later on in the process). Example:<br><strong>Pre-existing value {pre-existing-value}</strong><br><strong>New value</strong>"),
+  name: "facets[#{index}][allowed_values]",
+  rows: 5,
+  value: facet["allowed_values"]&.map { |opt| "#{opt["label"]} {#{opt["value"]}}" }&.join("\n"),
```
ryanb-gds, line 60: What's the value in showing requestors the machine value here? Seems like it would simplify our facet model code to omit it, and they're presumably going to be generated anyway using some kind of snake case/kebab case conversion.

app/views/admin/_facets_summary_card.html.erb
```diff
@@ -0,0 +1,51 @@
...
+    # We're on the edit page, so want to summarise the changes. So overriding `summary_rows`.
+    removed_facets = previous_schema.facets.map { |facet| facet["name"] } - schema.facets.map { |facet| facet["name"] }
+    previous_schema_facet_names = previous_schema.facets.map { |facet| facet["name"] }
+    proposed_schema_facet_names = schema.facets.map { |facet| facet["name"] }
+
+    summary_rows = proposed_schema_facet_names.map do |facet_name|
+      if (previous_schema_facet_names.include?(facet_name))
+        old_facet_config = previous_schema.facets.find { |f| f["name"] == facet_name }
+        new_facet_config = schema.facets.find { |f| f["name"] == facet_name }
+        updated = !facet_name.in?(removed_facets) && !old_facet_config.nil? && old_facet_config != new_facet_config
```
ryanb-gds, line 24: Feels like there ought to be a less verbose way to do this comparison stuff, or maybe we just create a schema diff model and move it elsewhere. Outside the scope of this PR though

ryanb-gds (commented):
Looking good. A few thoughts, but certainly nothing that would stop this merging, just observations really

app/views/admin/_facet_fields.html.erb
ChrisBAshton: The value is in retaining the _existing_ machine values. There are lots of [examples](https://github.com/alphagov/specialist-publisher/blob/535c7a402b072681a3d3c6838486a9b90dd0901b/lib/documents/schemas/aaib_reports.json#L83-L86) where we can't just kebab-case the name. I go into more detail on the "[Add page to edit facets](https://github.com/alphagov/specialist-publisher/pull/2929/commits/51d2fcd2bc3fc972f8627d8aba1f4930982e0ef0)" commit message.

ryanb-gds: Ah of course, thanks for explaining. Feels a _little_ fragile but I haven't got any better ideas

app/models/facet.rb
ChrisBAshton: It's a nice idea. I've now explored it in https://github.com/alphagov/specialist-publisher/commit/d5e35cd275f292739583f7e25d704a5683e943db, but as that commit message explains, it opens a bit of a can of worms - I'm keen that we leave that for a future PR.

ryanb-gds (approved):
Lovely stuff. Would be good to have one of those fancy JS enabled feature tests for the email subscription options as well.

ChrisBAshton:
> Lovely stuff. Would be good to have one of those fancy JS enabled feature tests for the email subscription options as well.

Working on it (in next Trello card)! :)

PR #2930, opened 2024-12-30
DRY up the views for creating/editing specialist documents

ChrisBAshton:
Best reviewed commit by commit. This PR:

- Adds a new `specialist_publisher_properties` hash to facets in the `lib/documents/schemas` folder, where they have an `allowed_values` property. This is so that we can describe, in the schema, whether said facet should allow multiple choices, or only one.
- Also adds a `omit_from_finder_content_item` property, which indicates whether or not the facet should be stripped from the finder configuration when presented to publishing API. This allows us to strip away a load of hardcoded logic related to "hidden indexable content".
- (The `specialist_publisher_properties` property is stripped out when publishing the finder schema itself, so no changes are required downstream to Publishing API / Search API.)
- With the schema now describing whether a dropdown list should allow single or multiple selections, we have all we need to start making the "new/edit specialist document" forms themselves schema-driven! Instead of each form being its own bespoke thing, we now have one shared template that looks at the finder schema to decide what to render.
- This has been rolled out to 31 of the 36 forms 🎉  The remaining five are a little too bespoke, but have "TODO" comments in them, signposting what needs to be done in order to get them to use the shared template.
- Note that some of the forms look a little different as a result (think slightly different labels, or a slightly different ordering of fields), but nothing that should be detrimental to the publisher (and at any rate, nothing worth complicating our nicely simplified solution for!). The forms are now consistent with what's in the finder schema (so if for example we need to reorder fields, we can easily do so from within the JSON file).

## Result

With this merged, it should make the [work to request edits to facets/options](https://trello.com/c/KNf2zNFI/3037-create-edit-finder-form-filters-and-options) much more effective, as, without this PR, adding or deleting facets would still require a developer to edit the corresponding view. With this PR, that should no longer be needed - they should be able to copy and paste the generated schema and voila, that's the Specialist Publisher part done±! (Leave aside Publishing API and Search API for the minute...)

± I mean, there's always something else to sort out in Specialist Publisher. In this case, FORMAT_SPECIFIC_FIELDS, validation, etc, might mean dev work is still required. There's a card up next  to [make the validation logic schema-driven](https://trello.com/c/PArCm42y/3292-make-specialist-publisher-validation-schema-driven) which should tackle a lot of that.

Once we have every form using the same shared view, there is a real opportunity to massively DRY up our test coverage too. We have a lot of tests doing much the same thing: testing that a particular type of specialist document can be created. This may have been necessary in a world where every part of that specialist document workflow was bespoke, but once they're using the same form, we can reasonably just test the creation of one specialist document and assume the rest are also fine.

## Screenshots

| Before | After |
|---|---|
|||

Trello: https://trello.com/c/KijYVpyo/3290-dry-up-specialist-publisher-views

---


app/presenters/finder_content_item_presenter.rb
```diff
@@ -27,6 +27,14 @@ def content_id
     file.fetch("content_id")
   end
 
+  def self.facets_without_specialist_publisher_properties(facets)
+    facets.reject { |facet| facet["specialist_publisher_properties"]&.fetch("omit_from_finder_content_item", false) }
+          .map do |facet|
+      facet.delete("specialist_publisher_properties")
+      facet
+    end
+  end
```
ryanb-gds, line 36: At some point I'm planning to pass this presenter our schema model instead of the raw content from the config file, at which point we will be able to move this logic there

app/views/shared/_specialist_document_form.html.erb
```diff
@@ -0,0 +1,48 @@
+<% f.object.class.finder_schema.facets.each do |field_config| %>
+  <% field = field_config["key"].to_sym %>
+  <% if field_config.nil? %>
+    <%= render layout: "shared/specialist_document_form_error", locals: { field: field } do %>
+    <% end %>
+  <% elsif field_config["type"] == "date" %>
+    <% format_name = f.object.class.finder_schema.filter["format"].to_sym %>
+    <%= render layout: "shared/date_fields", locals: { f: f, field: field, format: format_name, label: field_config["name"] } do %>
+    <% end %>
+  <% elsif !field_config["allowed_values"] %>
```
ryanb-gds, line 10: I do wonder if we should consider giving every facet a type rather than relying on this sort of indirect way of knowing that we have a select input, but that would be one for another day

ryanb-gds (approved):
Brilliant stuff Chris, this is a huge win 👏 👏 👏

app/views/shared/_specialist_document_form.html.erb
ChrisBAshton: Agreed! It felt like too big of a refactor to include here, but that is the direction I'd like to see us go in 👍

ChrisBAshton:
(Rebased against main)

PR #2933, opened 2024-12-31
Remove unused finder schema facet properties

ChrisBAshton:
I noticed that there were several facet fields defined in the
tribunal related finders, which aren't actually displayed in
the publishing UI, nor are they referenced in Publishing API.
(Presumably they were superseded by something else but never
removed).

I did this by cross-referencing every facet specified in the
finder schema with a corresponding property on
https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/shared/definitions/_specialist_document.jsonnet

Trello: https://trello.com/c/KijYVpyo/3290-dry-up-specialist-publisher-views

---

### DIFF
--- a/lib/documents/schemas/asylum_support_decisions.json
+++ b/lib/documents/schemas/asylum_support_decisions.json
@@ -50,13 +50,6 @@
       ]
     },
     {
-      "key": "tribunal_decision_category_name",
-      "name": "Category name",
-      "type": "text",
-      "display_as_result_metadata": true,
-      "filterable": false
-    },
-    {
       "key": "tribunal_decision_sub_categories",
       "name": "Sub-category",
       "type": "text",
@@ -141,13 +134,6 @@
           "value": "section-95-other"
         }
       ]
-    },
-    {
-      "key": "tribunal_decision_sub_category_name",
-      "name": "Sub-category name",
-      "type": "text",
-      "display_as_result_metadata": true,
-      "filterable": false
     },
     {
       "key": "tribunal_decision_judges",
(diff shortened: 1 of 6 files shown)


spec/lib/documents/schemas/finder_schemas_validation_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "spec_helper"
+
+RSpec.describe "Sense-checks for every finder schema" do
+  Dir["lib/documents/schemas/*.json"].map { |file| JSON.parse(File.read(file)) }.each do |finder_schema|
+    it "has each of its facets corresponding with a property in Publishing API" do
+      # TODO: append "_metadata" to the flood_and_coastal_... finder in Publishing API so that this workaround isn't required
```
ryanb-gds: This is unfortunate 😅

ryanb-gds (commented):
Are we 100% sure we want to do this? We have a presenter layer which decouples the finder config from the content item schema, and theoretically gives us the flexibility of a layer of abstraction between the two. We don't leverage that flexibility very heavily, except in the case of hidden indexable content, but we might want to in future. If we add this test we couple the finder config extremely tightly to the Publishing API schema even though we have a translation layer in between (so the test doesn't really prove anything IMO, as the presenter could change the keys to whatever it wants)

Would it make more sense to build the presenter using the finder config in this test and check that the keys from the presenter match the key in the content item schema?

ChrisBAshton:
Fair point, @ryanb-gds - I think I'll just make this PR about removing redundant config (i.e. commit c8c888f527a6045c9e074873f56c8cb63a544521) and not worry too much about how we concluded the config is safe to delete. 😀

ryanb-gds (approved):
I had wondered about these 😅 glad to see they can go

PR #2940, opened 2025-01-07
Full support for "email alert" configuration on the edit metadata form

ChrisBAshton:
This PR adds full support for editing "email alert" configuration on finders, insofar as publishers can request:

1. Subscribe to all selected facets by default
2. Subscribe to one particular facet
3. Subscribe via an external service
4. Do not allow subscriptions

Note that the 'hidden' properties inside `email_filter_options` (configuring the 'topic name' associated with each facet option, and also configuring which should be pre-checked, which is only used by one finder) is not configurable. It is, however, preserved, so the publisher can make tweaks to email alert configuration and not break things. (It is also automatically removed if the publisher turns off email alert subscriptions altogether). This should mean existing/legacy subscriber lists are unaffected by these changes, even in cases where they're configured slightly off-standard.

NB, it took a fair bit of investigation to figure out what each property does, in order to simplify the behaviour in the form. I've documented this in our [Specialist Publisher proposed improvements](https://docs.google.com/document/d/1hZD3eJjX3qWL6ntWV_aUq0U43m5LIEblwAeIRoUY7x0/edit?tab=t.0#heading=h.2f6xql52kyeg) doc.

## Screenshots

| Form | Results page | Expanded |
|---|---|---|
|1 form|2 results|3 expanded|

---

### DIFF
--- a/app/components/email_alert_fieldset_component.rb
+++ b/app/components/email_alert_fieldset_component.rb
@@ -11,23 +11,12 @@ class EmailAlertFieldsetComponent < ViewComponent::Base
     ].compact.join.html_safe
   end
 
-  # We are handling changes to email facet filters offline at present.
   def render_filtered_content_condition_inputs
     [
       render_hidden_signup_content_id_input("filtered_content_signup_id"),
       render_email_topic_list_title_prefix("filtered_content_list_title_prefix"),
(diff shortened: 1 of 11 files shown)


ryanb-gds (approved):
Looks good!

PR #2943, opened 2025-01-08
Massively simplify email-signup-related finder schema config

ChrisBAshton:
In short, this PR removes the old `email_filter_facets` config (which mostly duplicated the `facets` config further below it) in favour of generating it on the fly (with some override functionality added in as necessary). This PR also fixes up some inconsistencies between the email signup pages and the corresponding finder page.

See commits for details. Note that I temporarily add hundreds of lines of generated config, make my changes, and then remove the generated files, so that we could have confidence that we're still generating nearly identical content items for the email signup page presenter. The net result is a >2k line reduction 🎉 📉 

NB, this refactor took a fair bit of investigation into what each option does. I've documented this in our [Specialist Publisher proposed improvements](https://docs.google.com/document/d/1hZD3eJjX3qWL6ntWV_aUq0U43m5LIEblwAeIRoUY7x0/edit?tab=t.0#heading=h.2f6xql52kyeg) doc.

Ideas for next steps:

- Move _all_ email-alert-related config into the new `email_filter_options` hash (e.g. `signup_content_id`, `signup_link`), so that it's all nicely self contained. I had a quick go at this but it does involve updating a lot of tests, so the blast radius is pretty high. Probably one for a future PR.

---


lib/documents/schemas/cma_cases.json
```diff
@@ -30,88 +30,13 @@
...
-          "radio_button_name": "SAU referral",
-          "topic_name": "sau referral",
-          "prechecked": false
-        }
-      ]
+  "email_filter_options": {
+    "email_filter_by": "case_type",
+    "downcase_email_alert_topic_names": true,
+    "email_alert_topic_name_overrides": {
+      "ca98-and-civil-cartels": "CA98 and civil cartels"
```
ryanb-gds: I'm not sure using the facet key as a JSON config key feel quite right, as normally JSON config keys are snake cased (plus I have worries about some facet keys maybe having invalid characters or something, though that's probably not realistic). I guess in general I would say data and schema shouldn't mix, so it would be better to do something like
```
"email_alert_topic_name_overrides": [
    {
         facet_key: "ca98-and-civil-cartels",
         topic_name_override: "CA98 and civil cartels"
    }
]

ryanb-gds (approved):
This does couple the email subscription page extremely tightly to the finder page, but I think in this case that's the correct decision given the amount of config it will save us from having to create for new finders. Awesome 😄

lib/documents/schemas/cma_cases.json
ChrisBAshton: I think your proposal is more self-documenting 👍 though FWIW, I think having hyphens and special characters in the JSON key is still valid JSON.

Another concern I had about this bit of code, now that I remember - it will apply a topic name override irrespective of which facet the topic appears in. So if we happen to have the same facet key in two different facets, they'll both be given the override. We don't have any examples of that yet, and it seems quite unlikely to happen in real life (vs the overhead of nesting these overrides in a hash of facet names) but thought I'd better mention it while we're here!

ChrisBAshton: I'll sort that out while I'm here 👍

PR #2946, opened 2025-01-13
Remove pluralization of email subscription topic prefixes

ChrisBAshton:

Support for a hash value for `subscription_list_title_prefix` has been around for ages (at least [9 years](b2e3d30#diff-793b8db55b8b5c75a5bf7007ed820f5038b1e7f6160a7f43a12477372fc3c3ceR16)) but it is only used by a handful of finders.

A [whole bunch of logic](https://github.com/alphagov/finder-frontend/blob/484396b660715811500af3e3b226ecf0dd69c675/app/lib/email_alert_title_builder.rb#L19-L47) has to be maintained in finder-frontend to support plurality. Until now, we haven't felt the pain of it in Specialist Publisher, but we're now building a form whereby publishers can request edits to the configuration of their finder. That means we either need to add support for pluralised subscription topics in the form itself, or skip over any existing hash values, or, as we're doing in this commit, omit the hash values altogether. This commit argues that the feature is poorly utilised, not very valuable, and is actively blocking development of the "request edits to email alerts" form.

Only 6 finders made use of the feature, and for five of these, we were able to trivially replace references to "...the following {nouns|noun}" to "...the following noun(s)". There is one finder that had "category" vs "categories", which is harder to substitute, but [taking inspiration from other finders](https://github.com/alphagov/specialist-publisher/blob/b638d1326b570988d63aaec9de2bd6c2c877ea22/lib/documents/schemas/flood_and_coastal_erosion_risk_management_research_reports.json#L17) we can simply follow the form "{Finder name} {document noun}", with no need to overcomplicate things by trying to form a partial sentence.

With this commit, we can delete the corresponding code from finder-frontend.

NB, I wasn't sure what would happen to existing subscriber lists if we were to go and change the subscriber list title (but keeping all other params the same) - is a new one created, and if so, is that even a problem? So I tried it out. Rails console on Integration (Email Alert API), getting a subscriber list:

``` email-alert-api(prod)> SubscriberList.find_by(slug: "upper-tribunal-tax-and-chancery-chamber-decisions-with-the-following-category-charity") => id: 8330, title: "Upper Tribunal (Tax and Chancery Chamber) decision...", created_at: "2017-04-04 11:25:58.190324000 +0100", updated_at: "2019-09-16 13:26:39.390801000 +0100", document_type: "", tags: {"format"=>{"any"=>["tax_tribunal_decision"]}, "tribunal_decision_category"=>{"any"=>["charity"]}}, links: {}, email_document_supertype: "", government_document_supertype: "", signon_user_uid: nil, slug: "upper-tribunal-tax-and-chancery-chamber-decisions-...", url: nil, tags_digest: "56ebfeb705da5037ebe24ed1617068a34810753649c60eb80b...", links_digest: nil, content_id: nil, description: nil, last_audited_at: nil, last_alerted_at: "2024-11-12 08:45:04.474051000 +0000"> ```

I then tried calling `find_or_create_subscriber_list` with the same links, but a new title:

``` finder-frontend(prod)> GdsApi.email_alert_api.find_or_create_subscriber_list({ "title" => "Upper Tribunal NEW", "tags" => { "format"=>{"any"=>["tax_tribunal_decision"]}, "tribunal_decision_category"=>{"any"=>["charity"]} } }) ```

We can see that `title` got updated on the existing subscriber list:

``` email-alert-api(prod)> existing.reload => id: 8330, title: "Upper Tribunal NEW", created_at: "2017-04-04 11:25:58.190324000 +0100", updated_at: "2025-01-14 10:45:26.063566000 +0000", document_type: "", tags: {"format"=>{"any"=>["tax_tribunal_decision"]}, "tribunal_decision_category"=>{"any"=>["charity"]}}, links: {}, ... ```

Therefore we can safely conclude that tweaking the subscriber list title generation logic won't have a detrimental effect on data fragmentation. But just to be sure, the team in charge of Email Alert API will monitor the effects of this change, in https://trello.com/c/5HrbIlhQ/287-confirm-status-of-email-lists-affected-by-pluralisation-simplification

## Next steps

- Remove the corresponding code from Finder Frontend: https://github.com/alphagov/finder-frontend/pull/3626
- Remove the hash option from the schema in Publishing API: TODO

---

### DIFF
--- a/lib/documents/schemas/tax_tribunal_decisions.json
+++ b/lib/documents/schemas/tax_tribunal_decisions.json
@@ -22,10 +22,7 @@
     "357110bb-cbc5-4708-9711-1b26e6c63e86"
   ],
   "signup_content_id": "ae5afec1-30d6-4997-bdf8-7de94d2dd910",
-  "subscription_list_title_prefix": {
-    "singular": "Upper Tribunal (Tax and Chancery Chamber) decisions with the following category: ",
-    "plural": "Upper Tribunal (Tax and Chancery Chamber) decisions with the following categories: "
-  },
+  "subscription_list_title_prefix": "Upper Tribunal (Tax and Chancery Chamber) decisions",
   "email_filter_options": {
     "email_filter_by": "tribunal_decision_category",
     "downcase_email_alert_topic_names": true
--- a/lib/documents/schemas/algorithmic_transparency_records.json
+++ b/lib/documents/schemas/algorithmic_transparency_records.json
@@ -21,10 +21,7 @@
   ],
   "show_summaries": true,
   "signup_content_id": "99f5bacb-d1ef-4b93-87ef-bae129fed396",
-  "subscription_list_title_prefix": {
-    "singular": "Algorithmic transparency records with the following function: ",
-    "plural": "Algorithmic transparency records with the following functions: "
-  },
+  "subscription_list_title_prefix": "Algorithmic transparency records with the following function(s): ",
   "email_filter_options": {
     "email_filter_by": "algorithmic_transparency_record_function"
   },
--- a/lib/documents/schemas/algorithmic_transparency_records.json
+++ b/lib/documents/schemas/algorithmic_transparency_records.json
@@ -21,7 +21,7 @@
   ],
   "show_summaries": true,
   "signup_content_id": "99f5bacb-d1ef-4b93-87ef-bae129fed396",
-  "subscription_list_title_prefix": "Algorithmic transparency records with the following function(s): ",
+  "subscription_list_title_prefix": "Algorithmic transparency records",
   "email_filter_options": {
     "email_filter_by": "algorithmic_transparency_record_function"
   },
--- a/lib/documents/schemas/cma_cases.json
+++ b/lib/documents/schemas/cma_cases.json
@@ -26,10 +26,7 @@
     "competition/mergers",
     "competition/regulatory-appeals-references"
   ],
-  "subscription_list_title_prefix": {
-    "singular": "CMA cases with the following case type: ",
-    "plural": "CMA cases with the following case types: "
-  },
+  "subscription_list_title_prefix": "CMA cases with the following case type(s): ",
   "email_filter_options": {
     "email_filter_by": "case_type",
     "downcase_email_alert_topic_names": true,
--- a/lib/documents/schemas/cma_cases.json
+++ b/lib/documents/schemas/cma_cases.json
@@ -26,7 +26,7 @@
     "competition/mergers",
     "competition/regulatory-appeals-references"
   ],
-  "subscription_list_title_prefix": "CMA cases with the following case type(s): ",
+  "subscription_list_title_prefix": "CMA cases",
   "email_filter_options": {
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: lib/documents/schemas/tax_tribunal_decisions.json [hunk 1]
```diff
@@ -22,10 +22,7 @@
     "357110bb-cbc5-4708-9711-1b26e6c63e86"
   ],
   "signup_content_id": "ae5afec1-30d6-4997-bdf8-7de94d2dd910",
-  "subscription_list_title_prefix": {
-    "singular": "Upper Tribunal (Tax and Chancery Chamber) decisions with the following category: ",
-    "plural": "Upper Tribunal (Tax and Chancery Chamber) decisions with the following categories: "
-  },
+  "subscription_list_title_prefix": "Upper Tribunal (Tax and Chancery Chamber) decisions",
```
KludgeKML, line 25: Would "**Upper Tribunal (Tax and Chancery Chamber) decisions categorised by:**" make sense? (probably best to ask a content designer)

ChrisBAshton: Good prompt! I've now dug into the [corresponding logic in finder-frontend](https://github.com/alphagov/finder-frontend/blob/484396b660715811500af3e3b226ecf0dd69c675/app/lib/email_alert_title_builder.rb#L26). For strings, rather than hashes, the prefix is `"#{subscription_list_title_prefix.strip} with"`: so alongside the [suffix](https://github.com/alphagov/finder-frontend/blob/484396b660715811500af3e3b226ecf0dd69c675/app/lib/email_alert_title_builder.rb#L55) we'd have a value of "Upper Tribunal (Tax and Chancery Chamber) decisions with category Charity" (with what I currently have committed). I think that's ok.

(I.e. your suggestion would render as "Upper Tribunal (Tax and Chancery Chamber) decisions categorised by: with category Charity")

But that does mean I need to fix up my other updated values (e.g. we'd currently have "Algorithmic transparency records with the following function(s): with function General public services", which doesn't read great. I'll push up another commit.

ryanb-gds (approved):
Nice to see this gone

lib/documents/schemas/tax_tribunal_decisions.json
KludgeKML: Hmm, so has it always been awful in Finder-Frontend? Because the titles inside email-alert-api look okay, and I'm worried with this change (which includes removing a space) we'll go from (eg):

`CMA cases with the following case type: oim project`

..to..

`CMA casesoim project`

ChrisBAshton: It shouldn't do. I've written some tests for finder-frontend in https://github.com/alphagov/finder-frontend/pull/3626/files and it looks as though the subject would come out as `CMA cases with case type oim project`

KludgeKML: Ah, right - sorry, I was thinking this prefix was used by both FF and email alert api, but it's just used by FF and then the results are passed to email alert api. Cool.

KludgeKML (approved):
LGTM!

PR #2964, opened 2025-01-20
Update documentation

lauraghiorghisor-tw:
- update links
- remove references to changes to search-api, where they're not needed anymore
- include guidance on zendesk process with diff PRs
- include removal of facets and values

[Trello](https://trello.com/c/gLwEO6ib/3357-update-documentation-for-specialist-publisher)

### DIFF
--- a/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
+++ b/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
@@ -4,8 +4,13 @@ To create or edit a new specialist document you will have to make changes to thi
 application, [publishing-api][publishing-api] and [search-api][search-api]. You will not have to make any changes to frontend
 applications.
 
+**IMPORTANT**
+
+Recent changes to Specialist Publisher now allow users to self-serve. These requests will be coming through Zendesk, as a code diff for the [specialist-publisher][specialist-publisher] repo. Follow the steps below to validate the thoroughness of the changes and open the remaining PRs against [publishing-api][publishing-api] and [search-api][search-api], if needed.
+
 [publishing-api]: https://github.com/alphagov/publishing-api
 [search-api]: https://github.com/alphagov/search-api
+[specialist-publisher]: https://github.com/alphagov/specialist-publisher
 
 # __Creating__ a specialist document type
 
--- a/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
+++ b/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
@@ -149,11 +149,54 @@ To republish the finder:
 1. Deploy Specialist Publisher, deploy after the Search API schema update / reindex to avoid users publishing new documents with the new field.
 1. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
 
+## Removing a field from an existing specialist document
+
+NB: In order to remove a field, ensure no document are tagged with that field, or that the finder's owners are aware of the data loss implications.
+
+1. In `publishing-api`:
+    - Remove the field from the [specialist_document schema](https://github.com/alphagov/publishing-api/blob/6d5595470bd0e7f3072e06f0113e3ca5514b6e98/content_schemas/formats/shared/definitions/_specialist_document.jsonnet). 
+    - Run `bundle exec rake build_schemas` to regenerate schemas after removing the new value(s).
+   
+   See removal of `key_reference` in [example commit](https://github.com/alphagov/publishing-api/pull/3075/commits/03197cc43f11b762314a17ecbc37ec9601c0ede9).
+
+2. In `specialist publisher`:
+    - Remove field from [model](https://github.com/alphagov/specialist-publisher/tree/main/app/models).
+    - Remove field usage from [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields) (if this is a legacy finder that isn't referring to the shared view).
+    - Remove field from [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) files.
+    - Remove any other usages, such as from tests and factories.
+   
+   See removal of `key_reference` in [example commit](https://github.com/alphagov/specialist-publisher/pull/2942/commits/6b215cb8d02fdcee6d6e05d108f7e2cffca091c4).
+
+3. In `search-api`, remove the field from:
+    - The relevant schema in the [elasticsearch_types ](https://github.com/alphagov/search-api/tree/main/config/schema/elasticsearch_types)directory.
+    - The [elasticsearch_presenter](https://github.com/alphagov/search-api/blob/main/lib/govuk_index/presenters/elasticsearch_presenter.rb).
+    - The [specialist_presenter](https://github.com/alphagov/search-api/blob/main/lib/govuk_index/presenters/specialist_presenter.rb).
+    - The [field_definitions](https://github.com/alphagov/search-api/blob/main/config/schema/field_definitions.json) file.
+   
+   See removal of `key_reference` in [example commit](https://github.com/alphagov/search-api/pull/3120/commits/b18f25ce86f496e46294ebce6e37e42bf035c105).
+
+To republish the finder:
+1. Deploy Publishing API, Search API.
+2. Run `rake SEARCH_INDEX=govuk 'search:update_schema'` on Search API. If this errors, you may have to do a [full reindex](https://docs.publishing.service.gov.uk/manual/reindex-elasticsearch.html#how-to-reindex-an-elasticsearch-index).
+3. Deploy Specialist Publisher, deploy after the Search API schema update / reindex to avoid users publishing new documents with the new field.
+4. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
+
+
 ## Adding values for existing fields on a specialist document
 Specific values for fields of type array are now defined only in the `specialist_publisher` app. To add a value:
 1. In `specialist publisher`, add the new values to the relevant file in the [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) directory. See [this](https://github.com/alphagov/specialist-publisher/pull/2958/commits/930b4c82928a616cc848d1e759cf31b521771b15) commit for an example.
 2. Deploy Specialist Publisher.
 3. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
+
+## Removing values for existing fields on a specialist document
+
+NB: In order to remove a field value, ensure no document are tagged with that value, or that the finder's owners are aware of the data loss implications.
+
+Specific values for fields of type array are now defined only in the `specialist_publisher` app. To remove a value:
+1. In `specialist publisher`, remove the value entry from the `allowed_values` array of the relevant field, from the corresponding json file in the [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) directory.
+2. Deploy Specialist Publisher.
+3. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
+
 
 # __Editing__ a specialist finder
 
--- a/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
+++ b/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
@@ -129,9 +129,8 @@ To release the finder to the live stack:
 1. Merge and deploy Publishing API, Specialist Publisher and Search API, if changed.
    - Ensure you deploy Publishing API first, to avoid schema validation errors.
 2. Also deploy Email Alert API if you have made changes to it.
-3. Run `rake SEARCH_INDEX=govuk 'search:update_schema'` on Search API if you have made changes.
-   - This command will likely succeed for a newly created finder with no published documents, or when adding new fields and values.
-   - When making structural changes to finders, such as changing a field's type, this command might error and you will have to do a [full reindex](https://docs.publishing.service.gov.uk/manual/reindex-elasticsearch.html#how-to-reindex-an-elasticsearch-index). A full reindex takes around 30-45 minutes on Production, or 3-4 hours on Integration.
+3. Depending on the changes you've made, you might need to update mappings in Search API.
+   - You will likely need to run `rake SEARCH_INDEX=govuk 'search:update_schema'`. For further details, see this [section on reindexing](#reindexing-breakdown).
 4. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
 
 ## 8. Permissions
@@ -240,3 +239,44 @@ The following steps are required to remove a finder:
 4. Remove usages from `search-api`. See [example commit](https://github.com/alphagov/search-api/pull/2881/files).
 5. Remove any usages from `finder-frontend`, if applicable.
 6. Deploy all changes.
+
+# Reindexing breakdown
+
+Regardless on the changes you're trying to make, you can implement them in such a way that a full reindex is not necessary. 
+
+<!-- TODO Old mappings will be cleaned up by a monthly cronjob. -->
+
+Here's a breakdown of the different scenarios:
+
+1. Add new, change or delete facet options
+
+No mapping update or reindex necessary, because Elasticsearch does not have an allowlist of options. Any value works, as long as it is the correct type.
+
+2. Add new facet
+
+Mapping update (via `update_schema` rake task) necessary, but full reindex not required. Once the mapping has been updated, you can start adding data to both existing and new documents for the new field.
+
+Run: `rake SEARCH_INDEX=govuk 'search:update_schema'` in a Search API console, to update the mapping.
+
+3. Change name of facet
+
+Just change the facet label in Specialist Publisher. 
+
+There's no actual need to change the field key. If you nonetheless want to change the key as well:
+- declare a new field by following steps similar to [adding a new field](#adding-a-new-field-to-an-existing-specialist-document)
+- remove the old field config from Specialist Publisher
+- run `update_schema` as in 2) above
+- republish the finder
+- retag and republish all the content from Specialist Publisher with the new field.
+
+The Elasticsearch mapping for the old field would still exist, but it wouldn't be doing any harm.
+
+4. Change type of facet
+
+To avoid a reindex, you can follow the same steps as for a field key change at 3) above.
+
+5. Delete facet
+
+Remove config from Specialist Publisher and republish all content. The Elasticsearch mapping for the field would still exist, but it wouldn't be doing any harm.
+
+If you still want to reindex, follow the instructions [here](https://docs.publishing.service.gov.uk/manual/reindex-elasticsearch.html#how-to-reindex-an-elasticsearch-index). A full reindex takes around 30-45 minutes on Production, or 3-4 hours on Integration.
--- a/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
+++ b/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
@@ -40,6 +40,7 @@ You'll need to generate your own UUIDs for the `content_id` (of the finder), and
    irb(main):002:0> SecureRandom.uuid
    => "5087e8b6-ee54-40f9-b592-8c2813c7037d"
    ```
+For a breakdown of email subscription options see [Configure the email sign up page](#4-configure-the-email-sign-up-page).
 
 ### Create the model
 
@@ -75,9 +76,34 @@ Finally, you'll need to add your custom fields to:
 
 ## 4. Configure the email sign up page
 
-The email sign up page is rendered by [Finder Frontend](https://github.com/alphagov/finder-frontend) using the configuration in the new schema added to specialist publisher. The schema should specify `email_filter_by` and `email_filter_facets` (e.g. [cma-cases](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/documents/schemas/cma_cases.json#L29)).
+The email sign up page is rendered by [Finder Frontend](https://github.com/alphagov/finder-frontend) using the schema configuration in specialist publisher.
 
-If your email sign up page should have checkboxes (e.g. [cma-cases](https://www.gov.uk/cma-cases/email-signup)), you will need to edit email-alert-api by adding the new tags to [valid_tags.rb](https://github.com/alphagov/email-alert-api/blob/3e0018510ea85f5d561e2865ad149832b94688a1/lib/valid_tags.rb#L2).
+The finder default is to have no email subscription. Email subscriptions can be set as:
+
+### 1. Subscribe to all fields
+
+- Configure `signup_content_id` - a new `UUID` for the email signup page. Whilst this is enough to configure email subscription, it does not offer the user any filtering options.
+- You can allow the user to preserve their facet selection when navigating to the email subscription page. In the `email_filter_options` hash, set `email_filter_by` to `all_selected_facets`. This will pick up all the facets that have `allowed_values` and `filterable: true`. See [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/marine_equipment_approved_recommendations.json#L12).
+- You may exclude some of the facets by additionally setting `all_selected_facets_except_for` - see [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/export_health_certificates.json#L8).
+- Set `subscription_list_title_prefix` (optional).
+
+### 2. Subscribe to specific fields(set)
+
+- Configure the `signup_content_id` and, optionally, the `subscription_list_title_prefix`, as in the previous step.
+- Set `email_filter_by` to a specific facet - see [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/tax_tribunal_decisions.json#L13).
+- Edit [email-alert-api](https://github.com/alphagov/email-alert-api/tree/main/lib) by adding the new filters to [valid_tags.rb](https://github.com/alphagov/email-alert-api/blob/main/lib/valid_tags.rb).
+
+#### Breakdown of the email options:
+- `subscription_list_title_prefix` - typically set in most finders. It defines the beginning of the email title. For example, CMA cases have this value set to `CMA cases`, meaning an email title would read as "_CMA cases_ with digital markets unit". When omitted, the email title will only reference subscribed facets.
+- `email_alert_topic_name_overrides` - changes the display label of individual facets options in the email title the user receives upon subscribing.
+- `downcase_email_alert_topic_names` - downcases the display label of individual facet options in the email title the user receives upon subscribing.
+- `pre_checked_email_alert_checkboxes` - takes an array of facets, which will appear pre-checked on the email signup page.
+
+
+### 3. Signup link
+
+If the subscription is managed by an external service, it should be set via a `signup_link`. See [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/drug_safety_updates.json#L148).
+
 
 ## 5. Deploy a finder to the draft stack (for previewing)
 
--- a/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
+++ b/docs/creating-editing-and-removing-specialist-document-types-and-finders.md
@@ -9,15 +9,11 @@ applications.
 
 # __Creating__ a specialist document type
 
-<!-- TODO: Update example links in the documentation, to reflect latest documentation. Some of the steps are now no longer needed, but example PRs are outdated since no new requests for changes have been made. Also, remove the notes about not adding enums, once all examples are updated. -->
-
 ## 1. Add a schema to Publishing API
-See [example PR here](https://github.com/alphagov/publishing-api/pull/2589/files)
+See [example PR here](https://github.com/alphagov/publishing-api/pull/3026/files).
 
 1. Add the format to [allowed document types list](https://github.com/alphagov/publishing-api/blob/main/content_schemas/allowed_document_types.yml).
 2. Add any new field definitions to [this file](https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/shared/definitions/_specialist_document.jsonnet).
-
-    **Note**: Do not specify the field values as an enum, as we're moving towards a more relaxed schema definition.
 3. Run `bundle exec rake build_schemas` to regenerate schemas.
 
 When the PR is reviewed and its tests passing, it can be merged and deployed at this point.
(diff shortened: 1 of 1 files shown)


docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -4,20 +4,20 @@ To create or edit a new specialist document you will have to make changes to thi
 application, [publishing-api][publishing-api] and [search-api][search-api]. You will not have to make any changes to frontend
 applications.
 
+### IMPORTANT
+Recent changes to Specialist Publisher now allow users to self-serve. These requests will be coming through Zendesk, as a code diff for the [specialist-publisher][specialist-publisher] repo. Follow the steps below to validate the thoroughness of the changes and open the remaining PRs against [publishing-api][publishing-api] and [search-api][search-api], if needed.
```
lauraghiorghisor-tw, line 9: @ChrisBAshton - does this make sense? We'll probably change this as we go, but I think most of this documentation should still stay, despite the automation. Anything else that we should include about the Zendesk PRs at this time?

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -42,11 +42,14 @@ You'll need to generate your own UUIDs for the `content_id` (of the finder), and
 
 ### Create the model
 
-See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/models/cma_case.rb)
+See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/models/cma_case.rb).
 
 ### Create the view template
 
-See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/views/metadata_fields/_cma_cases.html.erb)
+- As per the latest changes, the view configuration is now being autogenerated from the schema.
```
lauraghiorghisor-tw: @ChrisBAshton Does this make sense for the view changes? Are you planning to add back in the possibility to customize placeholders and labels? (or whatever else we had custom).

ChrisBAshton: Thanks Laura. Not planning to add back in the option to customise placeholders and labels, no - I'm aiming for 100% standardised, schema-driven views/models etc 😅 

I think this section can be simplified btw, to something like:

```md
Create a `app/views/metadata_fields/_{format_name}.html.erb` file ([example](https://github.com/alphagov/specialist-publisher/blob/main/app/views/metadata_fields/_cma_cases.html.erb)).

Whilst a handful of legacy finders still use a custom view, all new finders should be referencing the [shared template](https://github.com/alphagov/specialist-publisher/blob/main/app/views/shared/_specialist_document_form.html.erb), which figures out which form fields to display based on what's in the finder schema.
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -9,15 +9,11 @@ applications.
 
 # __Creating__ a specialist document type
 
-<!-- TODO: Update example links in the documentation, to reflect latest documentation. Some of the steps are now no longer needed, but example PRs are outdated since no new requests for changes have been made. Also, remove the notes about not adding enums, once all examples are updated. -->
-
 ## 1. Add a schema to Publishing API
-See [example PR here](https://github.com/alphagov/publishing-api/pull/2589/files)
```
ChrisBAshton, line 15: I think https://github.com/alphagov/publishing-api/pull/3026 would be a better choice here, as it shows custom metadata fields that PR 2943 does not.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -42,11 +38,14 @@ You'll need to generate your own UUIDs for the `content_id` (of the finder), and
...
-See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/models/cma_case.rb)
+See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/models/cma_case.rb).
 
 ### Create the view template
 
-See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/views/metadata_fields/_cma_cases.html.erb)
+- As per the latest changes, the view configuration is now being autogenerated from the schema.
+- The configuration for the select type (one/multiple) now lives [in the schema](https://github.com/alphagov/specialist-publisher/blob/36170eb7841fc4acaad77603f3e55e0a80122a74/lib/documents/schemas/algorithmic_transparency_records.json#L151).
+
+See [CMA cases](https://github.com/alphagov/specialist-publisher/blob/main/app/views/metadata_fields/_cma_cases.html.erb).
```
ChrisBAshton: On line 55 below (GitHub won't let me comment, unfortunately, `https://github.com/alphagov/search-api/blob/1700c85e1484d1d9b2c1d46f276326bc06b51a14/config/schema/field_definitions.json` should be swapped out for `https://github.com/alphagov/search-api/blob/main/config/schema/field_definitions.json`. The referenced commit doesn't exist in the repo history, and this change would bring the link in line with the link on line 54.

ChrisBAshton: On line 70, we reference `email_filter_by` and `email_filter_facets`, which needs updating since #2943 (sorry!).

ChrisBAshton: Line 72 caught me by surprise!

> If your email sign up page should have checkboxes (e.g. [cma-cases](https://www.gov.uk/cma-cases/email-signup)), you will need to edit email-alert-api by adding the new tags to [valid_tags.rb](https://github.com/alphagov/email-alert-api/blob/3e0018510ea85f5d561e2865ad149832b94688a1/lib/valid_tags.rb#L2).

We should think about how best to call that out when making changes to email alert config 🤔

ChrisBAshton: Line 112 also caught me out:

> You'll need to [create the new permission manually](https://docs.publishing.service.gov.uk/repos/signon/usage.html#creating-new-permissions).

This sounds like a recommended step but it isn't really clear (it reads more as guidance than a thing to do). I'm not sure I did it for the last finder or two I wrote 🤔

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -119,19 +118,18 @@ We often receive requests to add new fields to a specialist document or to add n
...
-
-  **NOTE**: You will need to run `bundle exec rake build_schemas` to regenerate schemas after adding the new value(s). Do **not** specify the field values as an enum, as we're moving towards a more relaxed schema definition.
+   - Add the new field in the [specialist_document schema](https://github.com/alphagov/publishing-api/blob/6d5595470bd0e7f3072e06f0113e3ca5514b6e98/content_schemas/formats/shared/definitions/_specialist_document.jsonnet). See [example commit](https://github.com/alphagov/publishing-api/pull/2968/commits/b7d9cd1f6bb5d8d08fda7b6e219b2467134406c4).
+   - Run `bundle exec rake build_schemas` to regenerate schemas after adding the new value(s).
 
 2. In `specialist publisher`:
-   - Add the new field to the relevant [model](https://github.com/alphagov/specialist-publisher/tree/main/app/models).
-   - Add field to relevant [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields).
+   - Add the new field to the relevant [model](https://github.com/alphagov/specialist-publisher/tree/main/app/models), with any necessary validations and tests.
```
ChrisBAshton: We don't mention adding validations and tests in the "Create a model" step at the top of the doc.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -119,19 +118,18 @@ We often receive requests to add new fields to a specialist document or to add n
...
    - Add fields to relevant [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) files. 
-   
-<!-- TODO: Update these with up-to-date examples -->
-   See [this](https://github.com/alphagov/specialist-publisher/pull/1899/commits/cc9e8fe482dbca2ef678bb8219252e7bd4f4d154) commit for an example.
+   - Update the [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields), if needed. NB: As per the latest changes, you should not need to change the view file. The configuration for the select type (one/muliple) now lives [in the schema](https://github.com/alphagov/specialist-publisher/blob/36170eb7841fc4acaad77603f3e55e0a80122a74/lib/documents/schemas/algorithmic_transparency_records.json#L151).
```
ChrisBAshton:
```suggestion
   - For legacy finders that don't use the shared view, you may need to update the [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields).
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -119,19 +118,18 @@ We often receive requests to add new fields to a specialist document or to add n
...
 
-3. In `search-api`, add the new field in the following places (see [this](https://github.com/alphagov/search-api/pull/2320/commits/ca6d0142e29b9755aad2e6bd59a3f576b727bd24) commit for an example):
+<!-- TODO: Update these with up-to-date examples that don't include any view changes-->
```
ChrisBAshton:
```suggestion
<!-- TODO: Update these with up-to-date examples that don't include any view or 'expanded field' changes-->
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
ChrisBAshton: Yeah that sounds about right 👍

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -145,12 +145,54 @@ To republish the finder:
 1. Deploy Specialist Publisher, deploy after the Search API schema update / reindex to avoid users publishing new documents with the new field.
 1. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
 
+## Removing a field from an existing specialist document
+
+NB: In order to remove a field, ensure no document are tagged with that field, or that the finder's owners are aware of the data loss implications.
+
+1. In `publishing-api`:
+    - Remove the field from the [specialist_document schema](https://github.com/alphagov/publishing-api/blob/6d5595470bd0e7f3072e06f0113e3ca5514b6e98/content_schemas/formats/shared/definitions/_specialist_document.jsonnet). 
+    - Run `bundle exec rake build_schemas` to regenerate schemas after adding the new value(s).
```
ChrisBAshton:
```suggestion
    - Run `bundle exec rake build_schemas` to regenerate schemas after removing the value(s).
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -145,12 +145,54 @@ To republish the finder:
...
+   
+   See removal of `key_reference` in [example commit](https://github.com/alphagov/publishing-api/pull/3075/commits/03197cc43f11b762314a17ecbc37ec9601c0ede9).
+
+2. In `specialist publisher`:
+    - Remove field from [model](https://github.com/alphagov/specialist-publisher/tree/main/app/models).
+    - Remove field usage from [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields). NB: As per the latest changes, there should be no field usage in the views.
```
ChrisBAshton:
```suggestion
    - Remove field usage from [view](https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields) (if this is a legacy finder that isn't referring to the shared view).
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -145,12 +145,54 @@ To republish the finder:
...
+
+
 ## Adding values for existing fields on a specialist document
 Specific values for fields of type array are now defined only in the `specialist_publisher` app. To add a value:
 1. In `specialist publisher`, add the new values to the relevant file in the [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) directory. See [this](https://github.com/alphagov/specialist-publisher/pull/2958/commits/930b4c82928a616cc848d1e759cf31b521771b15) commit for an example.
 2. Deploy Specialist Publisher.
 3. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
 
+## Removing values for existing fields on a specialist document
+NB: In order to remove a field value, ensure no document are tagged with that value, or that the finder's owners are aware of the data loss implications.
```
ChrisBAshton, line 208:
```suggestion

NB: In order to remove a field value, ensure no documents are tagged with that value, or that the finder's owners are aware of the data loss implications.
```

ChrisBAshton (commented):
Thanks for putting this together, Laura. ⭐ 

There's a lot of repeated instructions in this file (e.g. how to republish a finder), and a fair bit of inconsistency as a result (e.g. in how we make edits to the model: sometimes we refer to tests and validation, sometimes we don't).

I think this file may benefit from a restructure with a lot more anchor links to itself. Something like:

```
## Creating a specialist finder

1. [Add a new format to Publishing API]()
2. [Update the field definitions in Publishing API](#update-the-field-definitions-in-publishing-api)
3. ...
4. ...
?. [Publish the finder](#publish-the-finder)

## Editing a specialist finder

### Adding, editing or removing a field definition

See [Update the field definitions in Publishing API](#update-the-field-definitions-in-publishing-api)

## Common tasks

### Update the field definitions in Publishing API
<how to update field definition>

### Publish the finder
<rake task stuff>
```

(i.e. aiming to define each thing only once, and write instructions that can be applicable whether creating something from scratch or editing something that exists already)

What do you think? I appreciate that's quite a big change 😬

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: Thanks Chris, I think I will keep a little NB for the select type, whilst the new approach is still a bit fresh.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: We have a recent one but it's just values, no facet PRs yet. Thanks for the pointer ;)

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: Added a mention in the create step as well.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -143,14 +148,54 @@ To republish the finder:
...
+    - The [specialist_presenter](https://github.com/alphagov/search-api/blob/main/lib/govuk_index/presenters/specialist_presenter.rb).
+    - The [field_definitions](https://github.com/alphagov/search-api/blob/main/config/schema/field_definitions.json) file.
+   
+   See removal of `key_reference` in [example commit](https://github.com/alphagov/search-api/pull/3120/commits/b18f25ce86f496e46294ebce6e37e42bf035c105).
 
 To republish the finder:
-1. Deploy Publishing API, Search API, Deploy Specialist Publisher.
-1. Publish the finder by running the rake task `publishing_api:publish_finders` or `publishing_api:publish_finder[your_format_name_based_on_the_schema_file]` against the specialist publisher app (rake tasks [here](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/tasks/publishing_api.rake)).
+1. Deploy Publishing API, Search API.
+2. Run `rake SEARCH_INDEX=govuk 'search:update_schema'` on Search API. If this errors, you may have to do a [full reindex](https://docs.publishing.service.gov.uk/manual/reindex-elasticsearch.html#how-to-reindex-an-elasticsearch-index).
```
ryanb-gds: I'm pretty sure this will error. I don't think you can remove a field from a mapping, you'd have to do a full reindex, even if there is no data in Elasticsearch for that field. I would argue that we shouldn't reindex every time that we are making field changes and should just make sure it gets done every few months to keep the mappings fresh. Would be good to get @ChrisBAshton perspective, or maybe even the Search API owners' opinions

ryanb-gds (changes requested):
Probably not much point suggesting to people they might not need to do a full reindex.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -74,9 +75,35 @@ Finally, you'll need to add your custom fields to:
...
+- Configure `signup_content_id` - a new `UUID` for the email signup page.
+- In the `email_filter_options` hash, set `email_filter_by` to `all_selected_facets`.
+- Set `subscription_list_title_prefix` (optional). <TODO what does this do?>
+
+See [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/marine_equipment_approved_recommendations.json#L12).
+
+### Subscribe to specific fields(set)
+
+- Set same options as for [all fields](#subscribe-to-all-fields).
+- Set the `email_filter_options`
```
ChrisBAshton:
```suggestion
- Set same options as for [all fields](#subscribe-to-all-fields), _except_ for `email_filter_options`, which needs setting slightly differently:
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -74,9 +75,35 @@ Finally, you'll need to add your custom fields to:
...
+  - Set `email_filter_by` to a specific facet - see [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/tax_tribunal_decisions.json#L13).
+  - Alternatively, set `all_selected_facets_except_for` to exclude specific facets - see [example](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/export_health_certificates.json#L8).
```
ChrisBAshton: Actually this is to be paired with `all_selected_facets` - so should be moved to the section above. From the [doc](https://docs.google.com/document/d/1hZD3eJjX3qWL6ntWV_aUq0U43m5LIEblwAeIRoUY7x0/edit?tab=t.0):

> `"email_filter_by": "all_selected_facets"`
> This special setting, introduced in [finder-frontend#1757](https://github.com/alphagov/finder-frontend/pull/1757), is used to describe the email subscriber list options open to the user. Sometimes paired with `all_selected_facets_except_for`, which is a way of using `all_selected_facets` but omitting some facets that would otherwise be included (i.e. filterable: true ones with an allowed_values array).

May also be worth calling that out explicitly in the section above (i.e. `all_selected_facets` == all facets that have `allowed_values` and have `filterable: true`, but no other facets).

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -74,9 +75,35 @@ Finally, you'll need to add your custom fields to:
...
+- Edit [email-alert-api](https://github.com/alphagov/email-alert-api/tree/main/lib) by adding the new filters to [valid_tags.rb](https://github.com/alphagov/email-alert-api/blob/main/lib/valid_tags.rb).
```
ChrisBAshton, line 96: I have a question about this one:

In the [example provided](https://github.com/alphagov/specialist-publisher/blob/91ee849549c5e5478126d06842513a516cacceb2/lib/documents/schemas/tax_tribunal_decisions.json#L13), (`tribunal_decision_category`), I can see that has been [added to valid_tags.rb](https://github.com/alphagov/email-alert-api/blob/3a0da7545fd88d79263600d1b40451beb82d9300/lib/valid_tags.rb#L69) - but so have all the other `tribunal_decision_*` fields (`tribunal_decision_judges`, etc).

So is the developer expected to add _only the facet referenced by `email_filter_by`, or should they be adding all of the possible facets that can be displayed on the email signup page (i.e. everything `filterable: true` and with `allowed_values`, or should it even be that they add _every_ facet key? I don't know enough about this bit of the system 🤔

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -74,9 +75,35 @@ Finally, you'll need to add your custom fields to:
...
+   
+Other `email_filter_options` are:
+   - `email_alert_topic_name_overrides` - changes the display label of individual facets <TODO in the email?>
+   - `downcase_email_alert_topic_names` - downcases the display label of individual facets <TODO in the email?>
```
ChrisBAshton: These values are used by Specialist Publisher to determine the `topic_name` that is included in the presenter:
https://github.com/alphagov/specialist-publisher/blob/4f52d2397f26fdccde7d5a0165c7aba80831398c/app/presenters/finder_signup_content_item_presenter.rb#L68-L79

The topic_name is [used in finder-frontend's EmailAlertTitleBuilder](https://github.com/alphagov/finder-frontend/blob/08ef89ddd92bd4e1d37256e15b48396bacdb7754/app/lib/email_alert_title_builder.rb#L41-L61) to construct the suffix of the "subscriber list title", which is the identifier given to the subscriber list in Email Alert API.

I _think_ that's all it does. So - [according to the tests](https://github.com/alphagov/finder-frontend/pull/3636/files#diff-c00e3a7c3d858e465d0629e0a5b3d8e59f446e991ba6db1293036835ca7208fbR82) - this setting should only result in a change in how facet options are referred to in the email title.

I've just checked and it doesn't affects the display of the facet options on the email signup pages themselves. For example, CMA cases shows [Digital Markets Unit](https://github.com/alphagov/specialist-publisher/blob/60ed75ec42c1c32d481f7edb9f4ce4fc3a1b643a/lib/documents/schemas/cma_cases.json#L34-L37) in the config, paired with [downcase_email_alert_topic_names: true](https://github.com/alphagov/specialist-publisher/blob/60ed75ec42c1c32d481f7edb9f4ce4fc3a1b643a/lib/documents/schemas/cma_cases.json#L8). The [email signup page](https://www.gov.uk/cma-cases/email-signup) shows "Digital Markets Unit", and only on the actual signup confirmation is it then referred to as "CMA cases with digital markets unit"

> Screenshot 2025-01-23 at 17 33 17

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -74,9 +75,35 @@ Finally, you'll need to add your custom fields to:
...
-The email sign up page is rendered by [Finder Frontend](https://github.com/alphagov/finder-frontend) using the configuration in the new schema added to specialist publisher. The schema should specify `email_filter_by` and `email_filter_facets` (e.g. [cma-cases](https://github.com/alphagov/specialist-publisher/blob/ce68fdb008cab05225e0493e19decba5365e1e20/lib/documents/schemas/cma_cases.json#L29)).
+The email sign up page is rendered by [Finder Frontend](https://github.com/alphagov/finder-frontend) using the schema configuration in specialist publisher.
+
+The finder default is to have no email subscription. Email subscriptions can be set as:
+
+### Subscribe to all fields
+
+- Configure `signup_content_id` - a new `UUID` for the email signup page.
+- In the `email_filter_options` hash, set `email_filter_by` to `all_selected_facets`.
+- Set `subscription_list_title_prefix` (optional). <TODO what does this do?>
```
ChrisBAshton: See my other comment about the topic names. Unlike topic names which determine the suffix, this bit determines the prefix - the beginning bit of the subscriber list title:
https://github.com/alphagov/finder-frontend/blob/08ef89ddd92bd4e1d37256e15b48396bacdb7754/app/lib/email_alert_title_builder.rb#L6-L26

[CMA cases](https://github.com/alphagov/specialist-publisher/blob/60ed75ec42c1c32d481f7edb9f4ce4fc3a1b643a/lib/documents/schemas/cma_cases.json#L406) has it defined as "CMA cases", forming the prefix of "CMA cases with digital markets unit".

If omitted, as in [medical safety alerts finder](https://github.com/alphagov/specialist-publisher/blob/60ed75ec42c1c32d481f7edb9f4ce4fc3a1b643a/lib/documents/schemas/medical_safety_alerts.json#L174), it defaults to just listing the facet choices you picked, without any prefix, because I guess the [prefix method in finder-frontend returns `nil` in those situations](https://github.com/alphagov/finder-frontend/blob/08ef89ddd92bd4e1d37256e15b48396bacdb7754/app/lib/email_alert_title_builder.rb#L20). So in that example:

> Field safety notice, national patient safety alert, device safety information, and medicines recall/notification

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
ChrisBAshton: Definitely up for getting an Search API owner opinion 👍 Can you explain the pros and cons of just running the reindex every once in a while?

My several-steps-removed-from-the-coalface understanding is that without the reindex, the finder wouldn't work. E.g. new facet option added, documents published that that option selected, and then user tries to search for that document, but no results. Or is that _not_ the case - some sort of automatic mapping when new facets defined? - and so we only have to worry about _renamed_ facet keys? (In which case, same situation: user tries to search and gets no results because the mappings are out of date).
^ for that situation, it feels like a "reindex every few months" approach isn't going to work.

NB, (still speaking a mile away from the coalface!), removing an option feels like less of an issue, since it would just disappear as an option from the facets on the finder, so users won't be trying and failing to search for something.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: Added more of a breakdown of email options and also improved permissions bit in latest commits.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: OH I see 🤔. The `except_` is not something we provide through the UI, correct?

lauraghiorghisor-tw: I think I got it now. Pushed changes.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: It appears that ALL of the facets of a finder that has any type of email subscription, including even the `drug_safety_updates` which has a set `signup_link` and the `export_health_certificates` `certificate_status` which is excluded - are in the valid tags.  

I think we need to test this. There is a good chance these were added due to lack of knowledge of what's required.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -138,12 +138,13 @@ To release the finder to the live stack:
...
 ## 7. Permissions
 
-Specialist Publisher grants access to the publishing interface for your new document type to the following Signon users:
+Specialist Publisher grants access to the publishing interface for a document type to the following Signon users:
 
-- Users that belong to the owner organisation AND have `Editor` permissions
-- Users that have the permission `your_new_document_type_editor`, e.g. `oim_project_editor`
+1. Users that belong to the owner organisation - have default "write" permissions, allowing them to view and draft new documents.
+2. Users that belong to the owner organisation AND have `editor` permission in Signon - are considered "departmental editors" and can publish, unpublish and discard documents.
+3. Users that have the permission `<your_new_document_type>_editor` in Signon - are granted "departmental editor" access regardless of their organisation. This is sometimes required for cross-departmental documents.
```
ChrisBAshton: Ah, so _that's_ the reason behind it - good to know. I wondered if perhaps it was also meant for users in an organisation who should be allowed to publish documents for Finder A, but not for Finder B (where both finders are owned by that organisation).

ChrisBAshton: I'd move the permission creation instruction onto this line, and also a clarification that it's not always needed.

```suggestion
3. Users that have the permission `<your_new_document_type>_editor` in Signon are granted "departmental editor" access regardless of their organisation. This is sometimes required for cross-departmental documents. These special permissions need to be [created manually](https://docs.publishing.service.gov.uk/repos/signon/usage.html#creating-editing-and-deleting-permissions) in Signon. You do not need to create this permission unless cross-departmental access has been explicitly requested.
```

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -138,12 +138,13 @@ To release the finder to the live stack:
...
 
-You'll need to [create the new permission manually](https://docs.publishing.service.gov.uk/repos/signon/usage.html#creating-new-permissions).
+You'll need to manually grant users access to the Specialist Publisher app in Signon, if they do not have it, as well as granting more advanced permissions such as `editor`. Custom cross-organisation permissions also need to be [created manually](https://docs.publishing.service.gov.uk/repos/signon/usage.html#creating-editing-and-deleting-permissions) in Signon.
```
ChrisBAshton:
```suggestion
You'll need to manually grant users access to the Specialist Publisher app in Signon, and the `editor` permission (or custom cross-organisation permissions from step 3 above) if appropriate.
```

ChrisBAshton (commented):
Added a quick review of the permissions section as requested.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: In terms of the documentation itself, I think it's safe to say we need to run `update_schema` on new finders and new facets/values but reindex when making "structural" changes such as changing field(facet)'s types. How does that sound?

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: `all_selected_facets` definitely don't have their facets listed, so I think we can keep this bit of guidance on valid tags under point 2. Subscribe to specific fields(set)

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: I figured it was not visible anywhere in the app, and couldn't remember how to test the email 👀. Updated now.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: Updated now.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
lauraghiorghisor-tw: > publish documents for Finder A, but not for Finder B 

Woah, that's a possibility. But then again it depends if the people are inside or outside of the organisation. Because if they belong to the organisation they will still have access to both document types, regardless of the custom permission. It's compounded, not exclusive.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
ryanb-gds: So, here's how it breaks down:

1. New, changed or deleted facet option

No mapping update or reindex necessary, because Elasticsearch does not have an allowlist of options. We can put any values we want in as long as they are the correct type.

2. New facet

Mapping update (via `update_schema` rake task) necessary, but full reindex not required. Once the mapping has been updated we can start adding data to both existing and new documents for the new field

3. Changed name of facet
We can probably just change the label in Specialist Publisher, there's probably not actually a need to change the field key. If we really want to change the key as well, then declaring a new field and republishing all the content from Specialist Publisher would be sufficient. We could then remove the old field config from Specialist Publisher. The Elasticsearch mapping for the old field would still exist, but it wouldn't be doing any harm.

4. Changed type of facet
The easiest way to do this is just to declare a new field and republish all the content from Specialist Publisher, as for case 3. Again, old Specialist Publisher content could be removed. The Elasticsearch mapping for the old field would still exist, but it wouldn't be doing any harm.

5. Deleted facet
Remove config from Specialist Publisher and republish all content. The Elasticsearch mapping for the field would still exist, but it wouldn't be doing any harm.

Obviously we would end up with a bunch of old mappings hanging around, which over time could be problematic. Therefore we'd probably want to run a reindex on a regular basis, maybe once a month. But it could be just a cronjob that runs overnight rather than a big scary developer task.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -68,49 +75,75 @@ Finally, you'll need to add your custom fields to:
...
-
-- Users that belong to the owner organisation AND have `Editor` permissions
-- Users that have the permission `your_new_document_type_editor`, e.g. `oim_project_editor`
-
-You'll need to [create the new permission manually](https://docs.publishing.service.gov.uk/repos/signon/usage.html#creating-new-permissions).
+   1. Open a new PR changing the `target_stack` of the finder from `draft` to `live` in Specialist Publisher json schema config
+   2. [Deploy the finder](#7-deploy-a-finder) to the live stack
+
+
+## 7. Deploy a finder
```
lauraghiorghisor-tw, line 127: @ChrisBAshton 

> There's a lot of repeated instructions in this file (e.g. how to republish a finder), and a fair bit of inconsistency as a result (e.g. in how we make edits to the model: sometimes we refer to tests and validation, sometimes we don't).

We had that style before and it gets quite difficult to navigate as you're trying to keep your place in the file. I settled for reusing the "republish a finder" section. Which I now have as its own "7. Deploy a finder". Which sorts of mainly works as a main paragraph/title. And I already find myself having to say things like "if needed" because sometimes you come to this deploy section with 3 PRs sometimes 2, or even 1. Same for the reindex bit, we can be more definitive about when the `update_schema` works or doesn't if we keep this inline. 

What do you think, Chris? To reuse or to repeat? That is the question 😅

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
ChrisBAshton: Investigated: 
https://trello.com/c/gLwEO6ib/3357-update-documentation-for-specialist-publisher#comment-67975901e987e676e348c151

lauraghiorghisor-tw: Changes added.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -178,3 +239,44 @@ The following steps are required to remove a finder:
 4. Remove usages from `search-api`. See [example commit](https://github.com/alphagov/search-api/pull/2881/files).
 5. Remove any usages from `finder-frontend`, if applicable.
 6. Deploy all changes.
+
+# Reindexing breakdown
```
lauraghiorghisor-tw, line 243: @ryanb-gds I am thinking that the best way to include these instructions is to revert the "Deploy a finder" section and just have the Republish step outlined in each top-level section. Then we can be explicit about if it's an update schema or not.

And maybe rename a facet and type change can have their own entries. 🤔

ryanb-gds: I'm starting to think this document could be broken down into several documents that are structured more around specific tasks. Not about to request you do that as part of this PR though, don't panic! 😅

ryanb-gds (approved):
I think this will do for now. There's probably room for improvement, but I think the best thing we can do is use the document for a bit and update it where we find it useful

PR #2976, opened 2025-01-24
Add form for developers: "Request a new finder"

ChrisBAshton:
This adds a new form, aimed at developers (and protected behind `gds_admin` permission), for generating a schema for a new finder. It is accessible at `/admin/new`.

The form reuses the view partials, and the controller logic, from the existing 'edit' forms for metadata and facets, with minimal tweaks. The form can be submitted multiple times, and the POSTed values are retained, so the schema can be gradually built up over several submissions rather than requiring everything all in one go.

While we're here, we've also tidied up some calls to Design System margin classnames, and opened a can of worms RE the permissions / policy implementation.

Trello: https://trello.com/c/UpRa6hBs/3359-build-request-a-new-finder-form

## Screenshots

<table>
<tr><th>Landing page form</th><th>Generated schema page (auto-expanded)</th></tr>
<tr><td valign="top"></td><td valign="top"></td></tr>
</table>

---


app/controllers/admin_controller.rb
```diff
@@ -3,48 +3,37 @@ class AdminController < ApplicationController
 
   before_action :check_authorisation, if: :document_type_slug
 
+  def new
+    authorize current_user, :can_request_new_finder?, policy_class: FinderAdministrationPolicy
+
+    @proposed_schema = FinderSchema.new
+    @proposed_schema.facets = []
+
+    if request.method == "POST"
```
ryanb-gds: Is there a reason we're not going for a standard `create` controller method?

app/policies/finder_administration_policy.rb
```diff
@@ -0,0 +1,16 @@
+class FinderAdministrationPolicy
+  attr_reader :user
+
+  # TODO: figure out why the exact same user object is passed in as
+  # two params from the controller. I must be doing something wrong,
+  # even if it's basically harmless!
+  def initialize(user, _user)
```
ryanb-gds, line 7: The Pundit gem's docs suggest you don't need to explicitly pass a user in when you call `authorize` in the controller, it passes the current_user in magically. That's why I'd guess you're getting double.

app/views/admin/_facets_form_fields.html.erb
```diff
@@ -0,0 +1,15 @@
+<p class="govuk-body">
+  Filters and its options are to help users find desired content on a specialist finder. These are selected by you for each specialist content item created in the Specialist Publisher. These are shown in the blue box on the documents and elsewhere.
```
ryanb-gds: The first sentence probably doesn't need "are to" in it?

ryanb-gds (approved):
This looks good, just a couple of observations

app/controllers/admin_controller.rb
ChrisBAshton: Because I don't know Rails as well as I'd like 😁 I'll fix this up now.

app/policies/finder_administration_policy.rb
ChrisBAshton: Awesome, thanks - will fix this up now.

ChrisBAshton: True... the [docs say](https://github.com/varvet/pundit#:~:text=The%20first%20argument,be%20anything%20really.):

> The first argument is a user. In your controller, Pundit will call the `current_user` method to retrieve what to send into this argument
> The second argument is some kind of model object, whose authorization you want to check. This does not need to be an ActiveRecord or even an ActiveModel object, it can be anything really.

I'm currently calling this in the controller:

```ruby
authorize current_user, :can_request_new_finder?, policy_class: FinderAdministrationPolicy
```

If I just drop `current_user` from that call, I get `undefined method new?' for an instance of FinderAdministrationPolicy` - so the `can_request_new_finder?` override seems to be ignored. If I go ahead and rename `def can_request_new_finder?` to `def new?`, it works, but then I'm needlessly passing a `:can_request_new_finder?` override (which is coming through as the second parameter - `_user`, in the code above). If I remove that override altogether (so now just `authorize policy_class: FinderAdministrationPolicy`) I get `ArgumentError in AdminController#new: wrong number of arguments (given 0, expected 1..2)`...

...it becomes a bit whackamole and I end up no further along - I still end up defining an unused parameter to the `initialize` method to get things working 🤔 I'll leave this for a future dev to sort out I think!

ryanb-gds: Weird 🤷

PR #2978, opened 2025-01-24
Make document validation schema-driven - presence only

lauraghiorghisor-tw:
Apply model validations based on schema configuration.

> Screenshot 2025-01-27 at 18 16 31

[Trello](https://trello.com/c/PArCm42y/3292-make-specialist-publisher-validation-schema-driven)

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -104,19 +104,19 @@ private
   end
 
   def facets_params
-    allowed_facet_params = %i[
-      key
-      name
-      short_name
-      type
-      preposition
-      display_as_result_metadata
-      filterable
-      allowed_values
-      _destroy
-    ]
     params.permit(
-      facets: allowed_facet_params,
+      facets: [
+        :key,
+        :name,
+        :short_name,
+        :type,
+        :preposition,
+        :display_as_result_metadata,
+        :filterable,
+        :allowed_values,
+        :_destroy,
+        { validations: [] },
+      ],
     )
   end
 
--- a/app/models/algorithmic_transparency_record.rb
+++ b/app/models/algorithmic_transparency_record.rb
@@ -1,9 +1,5 @@
 class AlgorithmicTransparencyRecord < Document
-  validates :algorithmic_transparency_record_organisation, presence: true
-  validates :algorithmic_transparency_record_organisation_type, presence: true
-  validates :algorithmic_transparency_record_phase, presence: true
-  validates :algorithmic_transparency_record_date_published, presence: true
-  validates :algorithmic_transparency_record_atrs_version, presence: true
+  apply_validations
 
   FORMAT_SPECIFIC_FIELDS = %i[
     algorithmic_transparency_record_organisation
--- a/app/models/facet.rb
+++ b/app/models/facet.rb
@@ -37,7 +37,7 @@ class Facet
       facet.display_as_result_metadata = params["display_as_result_metadata"]
       facet.filterable = params["filterable"]
       facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"])
-      facet.specialist_publisher_properties = facet_specialist_publisher_properties(params["type"])
+      facet.specialist_publisher_properties = facet_specialist_publisher_properties(params["type"], params["validations"])
       facet
     end
 
@@ -67,13 +67,32 @@ class Facet
       end
     end
 
-    def facet_specialist_publisher_properties(type)
+    def facet_specialist_publisher_properties(type, validations)
+      properties = facet_specialist_publisher_properties_select(type).merge(facet_specialist_publisher_properties_validations(validations))
+
+      properties.presence
+    end
+
+    def facet_specialist_publisher_properties_select(type)
       case type
       when "enum_text_multiple"
         { select: "multiple" }
       when "enum_text_single"
         { select: "one" }
+      else
+        {}
       end
+    end
+
+    def facet_specialist_publisher_properties_validations(validations)
+      return {} unless validations
+
+      rules = {}
+      validations.each do |key|
+        rules[key.to_sym] = {}
+      end
+
+      { validations: rules }
     end
 
     def facet_types_that_allow_enum_values
--- a/app/models/protected_food_drink_name.rb
+++ b/app/models/protected_food_drink_name.rb
@@ -1,18 +1,9 @@
 class ProtectedFoodDrinkName < Document
-  validates :registered_name, presence: true
-  validates :register, presence: true
-  validates :status, presence: true
-  validates :class_category, presence: true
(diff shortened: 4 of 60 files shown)


app/controllers/admin_controller.rb
```diff
@@ -100,19 +100,19 @@ def email_alert_params
...
+        :key,
+        :name,
+        :short_name,
+        :type,
+        :preposition,
+        :display_as_result_metadata,
+        :filterable,
+        :allowed_values,
+        :_destroy,
+        { validations: [] },
```
lauraghiorghisor-tw, line 118: All I wanted was the `validations: []`, which I couldn't write in a %i, then rubocop got a bit crazy and added the additional {}. Suggestions welcome.

lauraghiorghisor-tw:
Things we did not do, presence-wise:
- AaibReport - has an `unless` block for the presence validation - https://github.com/alphagov/specialist-publisher/blob/378d2c5851970575fb70cd17241cd2991d62080e/app/models/aaib_report.rb#L5
- StatutoryInstrument - has an `if` - https://github.com/alphagov/specialist-publisher/blob/378d2c5851970575fb70cd17241cd2991d62080e/app/models/statutory_instrument.rb#L8


The current implementation allows the presence validation to take a message, but not a block.

app/views/admin/_facet_fields.html.erb
```diff
@@ -49,6 +49,21 @@
     ]
 } %>
 
+<%= render "govuk_publishing_components/components/checkboxes", {
+  id: "facets-#{index}-validations",
+  name: "facets[#{index}][validations][]",
+  heading: "Validation",
```
lauraghiorghisor-tw, line 55: This can be changed to something else, I guess.

app/models/protected_food_drink_name.rb
```diff
@@ -1,18 +1,9 @@
 class ProtectedFoodDrinkName < Document
-  validates :registered_name, presence: true
-  validates :register, presence: true
-  validates :status, presence: true
-  validates :class_category, presence: true
-  validates :protection_type, presence: true
-  validates :country_of_origin, presence: true
-  validates :traditional_term_grapevine_product_category, presence: true, allow_blank: true
```
lauraghiorghisor-tw, line 8: Removed this and below because no validations were happening due to using both `presence: true` and `allow_blank`. As far as we can tell, this is correct. We tried it in the UI and also a little unit test that we discarded.

app/models/utaac_decision.rb
```diff
@@ -1,8 +1,6 @@
 class UtaacDecision < Document
-  validates :tribunal_decision_categories, presence: true
-  validates :tribunal_decision_sub_categories, presence: false
```
lauraghiorghisor-tw, line 3: Again, redundant. Removed.

app/controllers/admin_controller.rb
GDSNewt: I'm assuming it's a hard requirement - that is we can't get CI to pass without it? 

My google searches on whether this is strictly needed are inconclusive, but if it works then i'm sure it's fine.

app/models/facet.rb
```diff
@@ -67,13 +67,32 @@ def facet_allowed_values(values, type)
...
+    end
+
+    def facet_specialist_publisher_properties_select(type)
       case type
       when "enum_text_multiple"
         { select: "multiple" }
       when "enum_text_single"
         { select: "one" }
+      else
+        {}
```
GDSNewt, line 83: Is this handling for the introduction of properties in future? Will things blow up if a facet doesn't have specialist properties? I seem to recall date doesn't have them yet, at least in the generic view Chris made.

GDSNewt (approved):
Looks good, getting rid of the presence stuff should save lots of time. 👍 

Eager to get the dates soon 🙂 

:shipit:

app/controllers/admin_controller.rb
lauraghiorghisor-tw: If you don't set this (and set it right, I should add), then the fields you're sending from the form get dropped, and nothing happens 💨

app/models/facet.rb
lauraghiorghisor-tw: In the `specialist_publisher_properties` method that calls this select bit and the new validations bit, we're doing a smth.merge(smth), where both smths could be nil (before) and thus make the merge rather unhappy. So we changed both bits to return {} instead. Checking for nil otherwise ended up being uglier than this ☠️ but I can appreciate that this is now unclear.
 
The `specialist_publisher_properties` method that calls the two bits also eventually needs to be converted from {} to nil so we can compact the whole facets field 😅 - we don't want to include any `specialist_publisher_properties` that are blank, nor any blank `validations` in the generated schema. This code would be used by the diffing logic and cause issues otherwise.

app/views/admin/_facet_fields.html.erb
ChrisBAshton: Looks good 👍

app/models/algorithmic_transparency_record.rb
```diff
@@ -1,9 +1,5 @@
 class AlgorithmicTransparencyRecord < Document
-  validates :algorithmic_transparency_record_organisation, presence: true
-  validates :algorithmic_transparency_record_organisation_type, presence: true
-  validates :algorithmic_transparency_record_phase, presence: true
-  validates :algorithmic_transparency_record_date_published, presence: true
-  validates :algorithmic_transparency_record_atrs_version, presence: true
+  apply_validations
```
ChrisBAshton, line 2: Similar to my comment in the other PR - can we call `apply_validations` in the Document model itself, and therefore omit it from all these child classes?

PS: really nice cleanup of the unused validation for `traditional_term_grapevine_product_category` etc ⭐ 

PPS:

> In cases where we had `presence: true, date: true, allow_blank: true`, we removed both presence and allow_blank, because allow_blank has no effect on date. Our DateValidator will always allow blank. We want to keep that feature, as it makes it easier to keep the date and presence validations separate. The date will be dealt with in a future commit.

Is there a card to capture that work? Just hope it doesn't get forgotten!

lauraghiorghisor-tw: I meant the date PR that was opened as part of this card.

Our DateValidator will always allow blank. -> I mean DOES now already always allow blank, rather.

lauraghiorghisor-tw: >  can we call apply_validations in the Document model itself

We tried that but it wasn't as straightforward. I genuinely expected to just make the one change (call the method at the top of `Document`) and have all tests passing with basically no additional changes. I couldn't quite understand why, but the error we were getting was smth like not being able to find `apply_validations` as a class method on the Document class, when it was clearly defined as a class method on the Document class.

PR #2982, opened 2025-01-27
Refactor format specific fields

lauraghiorghisor-tw:
Extract format specific fields from schema

### DIFF
--- a/app/models/business_finance_support_scheme.rb
+++ b/app/models/business_finance_support_scheme.rb
@@ -2,15 +2,7 @@ class BusinessFinanceSupportScheme < Document
   apply_validations
   validates :continuation_link, presence: true
 
-  FORMAT_SPECIFIC_FIELDS = %i[
-    business_sizes
-    business_stages
-    continuation_link
-    industries
-    regions
-    types_of_support
-    will_continue_on
-  ].freeze
+  FORMAT_SPECIFIC_FIELDS = format_specific_fields + %i[continuation_link will_continue_on]
 
   attr_accessor(*FORMAT_SPECIFIC_FIELDS)
 
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -370,6 +370,10 @@ class Document
     []
   end
 
+  def self.format_specific_fields
+    finder_schema.facets.map { |facet| facet["key"].to_sym }.freeze
+  end
+
   def self.apply_validations
     finder_schema.facets.each do |facet|
       key = facet["key"]
--- a/app/models/research_for_development_output.rb
+++ b/app/models/research_for_development_output.rb
@@ -1,14 +1,7 @@
 class ResearchForDevelopmentOutput < Document
   apply_validations
 
-  FORMAT_SPECIFIC_FIELDS = %i[
-    research_document_type
-    country
-    first_published_at
-    authors
-    theme
-    review_status
-  ].freeze
+  FORMAT_SPECIFIC_FIELDS = format_specific_fields + [:review_status]
 
   attr_accessor(*FORMAT_SPECIFIC_FIELDS)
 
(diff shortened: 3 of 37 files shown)


app/models/business_finance_support_scheme.rb
```diff
@@ -2,15 +2,7 @@ class BusinessFinanceSupportScheme < Document
...
-  FORMAT_SPECIFIC_FIELDS = %i[
-    business_sizes
-    business_stages
-    continuation_link
-    industries
-    regions
-    types_of_support
-    will_continue_on
-  ].freeze
+  FORMAT_SPECIFIC_FIELDS = [:continuation_link, :will_continue_on] + format_specific_fields
```
lauraghiorghisor-tw: Additional fields that are not configured as facets, so we had to add them separately.

app/models/statutory_instrument.rb
```diff
@@ -7,13 +7,7 @@ class StatutoryInstrument < Document
...
   validates :withdrawn_date, absence: { message: "must be blank if not withdrawn" }, unless: :withdrawn_sifting_status?
 
-  FORMAT_SPECIFIC_FIELDS = %i[
-    laid_date
-    sift_end_date
-    sifting_status
-    subject
-    withdrawn_date
-  ].freeze
+  FORMAT_SPECIFIC_FIELDS = [:withdrawn_date] + format_specific_fields
```
lauraghiorghisor-tw: Additional fields that are not configured as facets, so we had to add them separately.

app/models/research_for_development_output.rb
```diff
@@ -2,14 +2,7 @@ class ResearchForDevelopmentOutput < Document
...
 
-  FORMAT_SPECIFIC_FIELDS = %i[
-    research_document_type
-    country
-    first_published_at
-    authors
-    theme
-    review_status
-  ].freeze
+  FORMAT_SPECIFIC_FIELDS = [:review_status] + format_specific_fields
```
lauraghiorghisor-tw: Additional fields that are not configured as facets, so we had to add them separately.

GDSNewt: Be intrigued to hear more about this in a dev huddle / in future tickets

GDSNewt (approved):

app/models/research_for_development_output.rb
lauraghiorghisor-tw: There were around 20 failing tests that pointed to`ResearchForDevelopmentOutput`, `StatutoryInstrument` and `BusinessFinanceSupportScheme` - these models  do something with these fields specifically. 

You can see they were a blocker for @ChrisBAshton as well when doing the views ->  https://github.com/alphagov/specialist-publisher/tree/main/app/views/metadata_fields there are TODOs in all three views corresponding to these models.

app/models/document.rb
```diff
@@ -370,6 +370,10 @@ def custom_error_message_fields
     []
   end
 
+  def self.format_specific_fields
```
ChrisBAshton, line 373: Is it possible to delete `FORMAT_SPECIFIC_FIELDS = format_specific_fields` from the models altogether, and have the `Document.initialize` method call this internally?

ChrisBAshton (approved):
Much better than before 👍 But I do still wonder if we can streamline the model bloat further - see below

app/models/document.rb
lauraghiorghisor-tw: Tried and failed, but leaving this here: 

```
  def initialize(params = {}, format_specific_fields = [])
    @content_id = params.fetch(:content_id, SecureRandom.uuid)
    @format_specific_fields = self.class.format_specific_fields

    set_attributes(params, COMMON_FIELDS + self.class.format_specific_fields)
  end

  def self.inherited(subclass)
    super
    format_specific_fields.each { |field| subclass.attr_accessor(field) }
  end

  def self.format_specific_fields
    finder_schema.facets.map { |facet| facet["key"].to_sym }.freeze
  end
```

Progressed to getting the error `No such file or directory @ rb_sysopen - /govuk/specialist-publisher/lib/documents/schemas/documents.json` - meaning that it is now trying to load schema for the Document model...

Another thing to note is that we'd probably still have to pass the `format_specific_fields` to the initializer for those fancy models that have additional `attr_accessors`, at least for now.

GDSNewt: This is the exact issue I came across when trying to move everything into Document.rb 😢

PR #2984, opened 2025-01-28
Make validation schema driven -  date

lauraghiorghisor-tw:
[Trello](https://trello.com/c/PArCm42y/3292-make-specialist-publisher-validation-schema-driven)

### DIFF
--- a/app/models/aaib_report.rb
+++ b/app/models/aaib_report.rb
@@ -1,7 +1,6 @@
 class AaibReport < Document
   validates :date_of_occurrence,
             presence: true,
-            date: true,
             unless: lambda { |report|
                       report.report_type == "safety-study" && report.date_of_occurrence.blank?
                     }
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -376,6 +376,7 @@ class Document
       validations = facet.dig("specialist_publisher_properties", "validations") || {}
 
       validates key.to_sym, presence: validations.fetch("required").deep_symbolize_keys if validations.key?("required")
+      validates key.to_sym, date: true if facet["type"] == "date"
     end
   end
 
--- a/app/models/statutory_instrument.rb
+++ b/app/models/statutory_instrument.rb
@@ -1,10 +1,8 @@
 class StatutoryInstrument < Document
   apply_validations
-  validates :laid_date, date: true
-  validates :sift_end_date, date: true
   validates :sift_end_date, absence: { message: "must be blank when withdrawn" }, if: :withdrawn_sifting_status?
   validates :primary_publishing_organisation, presence: true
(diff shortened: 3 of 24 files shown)


app/models/statutory_instrument.rb
```diff
@@ -1,10 +1,8 @@
 class StatutoryInstrument < Document
   apply_validations
-  validates :laid_date, date: true
-  validates :sift_end_date, date: true
   validates :sift_end_date, absence: { message: "must be blank when withdrawn" }, if: :withdrawn_sifting_status?
   validates :primary_publishing_organisation, presence: true
-  validates :withdrawn_date, presence: true, date: true, if: :withdrawn_sifting_status?
+  validates :withdrawn_date, presence: true, if: :withdrawn_sifting_status?
```
lauraghiorghisor-tw, line 5: I believe removing date here is ok, since it allows blank. Then the if: block would only apply to presence.

app/models/aaib_report.rb
```diff
@@ -1,7 +1,6 @@
 class AaibReport < Document
   validates :date_of_occurrence,
             presence: true,
-            date: true,
```
lauraghiorghisor-tw, line 4: Deleted date from here - again, think it's fine, as it allows blank, and the unless block would apply to the presence validation.

GDSNewt (approved):

app/models/document.rb
```diff
@@ -376,6 +376,7 @@ def self.apply_validations
       validations = facet.dig("specialist_publisher_properties", "validations") || {}
 
       validates key.to_sym, presence: validations.fetch("required").deep_symbolize_keys if validations.key?("required")
+      validates key.to_sym, date: true if facet["type"] == "date"
```
ChrisBAshton, line 379: Nice - and presumably this doesn't need adding to the "request edits" form because it's just something we'll want to impose on _all_ date types.

lauraghiorghisor-tw: Yep, no need to add it to the schema, nor to any form (edit/add).

The only iffy bit here was the fact that in those few cases where we have `presence: true, date: true` + conditional blocks (`if:` or `unless:`) -> the block applies to BOTH the presence and date.

Removing the date from there and applying it universally, based on type, means that we now validate date when the conditional is also NOT met. Nonetheless, those conditionals are written in such a way that basically you validate date+presence under those specific circumstances but then you still validate date if provided, even when the field is not presence-required. Which is still what happens now, because our custom [DateValidator](https://github.com/alphagov/specialist-publisher/blob/d540540c0565b47fda2505bd6dab7f774aea58e5/app/validators/date_validator.rb#L3) allows blank. I pointed out in comments above those instances where we made this call.

PR #2987, opened 2025-01-28
Fix misnamed date inputs

ryanb-gds:
I don't really know how this has been working thus far, but moving to Rack 3 in https://github.com/alphagov/specialist-publisher/pull/2871 seems to stop it from working. Anyway, this PR makes sure we're using the Rails input naming conventions properly.

https://trello.com/c/gQ5F9ysj


ChrisBAshton (approved):
LGTM!

PR #2992, opened 2025-01-29
Upgrade to Rails 8

ryanb-gds:
As usual Rails are fiddling about with the error pages. A few new parameters get filtered from logs too.

Trello: https://trello.com/c/lYQQqSvP


ChrisBAshton (approved):
LGTM 👍

PR #3020, opened 2025-02-19
Added ADR around nested facet logic implementation

minhngocd:
https://trello.com/c/OA4zE8m0/3430-ipo-finder-request
https://trello.com/c/4RJusHld/3374-nested-facets-add-new-document

PR where implementation took place: https://github.com/alphagov/specialist-publisher/pull/2923

### DIFF
--- a/.adr-dir
+++ b/.adr-dir
@@ -1,0 +1,1 @@
+docs/adr
--- a/docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
+++ b/docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
@@ -1,0 +1,73 @@
+# 1. Implement nested facets tagging for specialist documents as a single dropdown
+
+Date: 2025-02-14
+
+## Status
+
+Proposed
+
+## Related PRs
+
+- [Implemented as part of this PR](https://github.com/alphagov/specialist-publisher/pull/2923)
+
+## Context
+
+Nested facets are required by a department as part of their new finder. Their existing finder contains: 
+
+- For filtering, a "Facet" dropdown and a "Sub facet" dropdown to filter for documents. Users are able to filter through a single facet value lookup at a time. 
+- The documents are tagged with multiple Facet and Sub facet values
+
+We need to be able to replicate this behaviour with minimal efforts to meet a deadline. 
+
+## Decision
+
+What we have implemented:
+
+- On the publisher side for tagging, a single dropdown that merges the facets' values (main facet and sub facets, essentially leaf nodes).
+  - we allow main facets to be defined without subfacets. It would show up in the dropdown if that is the case. For the initial finder we are creating, they don't have any main facet value with no sub-facets. 
+  - if the main facet has sub-facets, then you can only choose the sub-facets (leaf nodes). We don't allow tagging of main facet if the value has sub-facets available. Reason being we still have things to clarify about that behaviour (is it equivalent to tagging all subs, or no subs). Since the current finder doesn't require it, we de-scope this further down the line.
+- When we send the document to Publishing API, we write logic into the document model to figure out whether the leaf node is a parent facet or sub-facet value in the dropdown, and split the metadata we send to Publishing API into its own facet and a separate sub-facet in `specialist-publisher/app/models/document.rb` 
+  - Reasons for this being if we send them as separate facets, we can keep Publishing API and Search API functionality unchanged. 
+- When we load the edit page or new page again, the document is loaded from Publishing API, with the facet and sub-facet being separate facets, but the dropdown element expects a combined leaf node dropdown as mentioned in the first point, so we have to merge them back together again into the main facet field in `specialist-publisher/app/controllers/documents_controller.rb`.
+
+## Consequences
+
+Pros: 
+
+- Restrict user input to only valid options with a single dropdown. 
+- This supports both multiple select entries and single select entries. 
+- The workflow discussed as part of analysis document works.
+
+Cons:
+
+- Logic to merge the parent facet values, sub-facet values at form phase, split them at send phase, merge them again at load phase might be a little convoluted.
+
+## Alternatives proposed
+
+### Have tagging done as 2 separate facets with validation 
+
+Even on publisher side, surface all of a facet's parents facets as one dropdown, and all of the sub-facets as a separate dropdown. As outlined on [this commit](https://github.com/alphagov/specialist-publisher/commit/c7346db771df5ccafd5fff8f7f687a8211436f6a).
+
+Pros:
+
+- Consistency of 2 separate facets being displayed to users upon tagging and viewing metadata. 
+- No change logic required to split or merge facets
+
+Cons:
+
+- Repeat of information. The sub-facets are also prefixed with parent facet in any case (e.g. to differentiate between multiple "Not Applicable" sub-facet entries). Users will have to tag both parent facet and the sub-facet with the prefix. 
+- There is no structural tie between the parent facet and the sub-facet when tagging. Nothing stopping the user from tagging mismatching pairs or not tagging a sub-facet. 
+  - In order to get around this, we would have to implement some validation on the backend to ensure users are adding correct pairs.
+
+### Remove support for "mixed" leaf nodes and only support sub-facet entry only
+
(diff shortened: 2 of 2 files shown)


docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Alternatives proposed
+
+### Have tagging done as 2 separate facets with validation 
+
+Even on publisher side, surface all of a facet's parents facets as one dropdown, and all of the sub-facets as a separate dropdown. As outlined on [this branch](https://github.com/alphagov/specialist-publisher/compare/intellectual-property-office-finder_2?expand=1).
+
+Pros:
+
+- Consistency of 2 separate facets being displayed to users upon tagging and viewing metadata.
```
ryanb-gds, line 53: I'm not sure this consistency is really a pro, is it? These are two separate user groups so there isn't really any added value in having a similar UX

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
```diff
@@ -0,0 +1,71 @@
...
+- [Implemented as part of this PR](https://github.com/alphagov/specialist-publisher/pull/2923)
+
+## Context
+
+Nested facets are required by a department as part of their new finder. Their existing finder contains: 
+
+- For filtering, a "Facet" dropdown and a "Sub facet" dropdown to filter for documents. Users are able to filter through a single facet value lookup at a time. 
+- The documents are tagged with multiple Facet and Sub facet values
+
+We need to be able to replicate this behaviour with minimal efforts. Decisions and analysis have been outlined in [this document](https://docs.google.com/document/d/1GaSnboJ8_0KnCd3uaCDxuJg0JtECh45hOfLnuoDkaAI/edit?tab=t.0).
```
ryanb-gds: I don't particularly like referring to Google Docs in an ADR, I feel the ADR should stand alone.

ryanb-gds (changes requested):
Generally looks good, I think we should take the private Google Doc link out of the public ADR though 😉

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
minhngocd: yea agreed, i'm also not keen on having validation to cross check across the two fields - but we'd note it here in any case as part of the alternatives we'd considered.

minhngocd:
@ryanb-gds updated

ryanb-gds (approved):
Good ADR, thanks Minno and co!

.adr-dir
```diff
@@ -0,0 +1 @@
+docs/adr
```
ChrisBAshton, line 1: I've not seen this `.adr-dir` file convention before. What's it for? Should it be included in this PR?

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
```diff
@@ -0,0 +1,71 @@
...
+
+Cons:
+
+- Logic to merge the parent facet values, sub-facet values at form phase, split them at send phase, merge them again at load phase might be a little convoluted.
+
+## Alternatives proposed
+
+### Have tagging done as 2 separate facets with validation 
+
+Even on publisher side, surface all of a facet's parents facets as one dropdown, and all of the sub-facets as a separate dropdown. As outlined on [this branch](https://github.com/alphagov/specialist-publisher/compare/intellectual-property-office-finder_2?expand=1).
```
ChrisBAshton: Suggest linking to the commit instead, as the branch can easily be deleted in future, but the commit should live forever (citation needed!)

```suggestion
Even on publisher side, surface all of a facet's parents facets as one dropdown, and all of the sub-facets as a separate dropdown. As outlined in [this commit](https://github.com/alphagov/specialist-publisher/commit/c7346db771df5ccafd5fff8f7f687a8211436f6a).
```

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
```diff
@@ -0,0 +1,71 @@
...
+
+Pros:
+
+- Consistency of 2 separate facets being displayed to users upon tagging and viewing metadata. 
+- No change logic required to split or merge facets
+
+Cons:
+
+- Repeat of information. The sub-facets are also prefixed with parent facet in any case (e.g. to differentiate between multiple "Not Applicable" sub-facet entries). Users will have to tag both parent facet and the sub-facet with the prefix. 
+- There is no structural tie between the parent facet and the sub-facet when tagging. Nothing stopping the user from tagging mismatching pairs or not tagging a sub-facet.
```
ChrisBAshton, line 59: "Not tagging a sub-facet" is allowed though, isn't it? As per our [assumptions](https://docs.google.com/document/d/1GaSnboJ8_0KnCd3uaCDxuJg0JtECh45hOfLnuoDkaAI/edit?tab=t.0#heading=h.cpq4r1qiln6j).

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
```diff
@@ -0,0 +1,71 @@
...
+- The documents are tagged with multiple Facet and Sub facet values
+
+We need to be able to replicate this behaviour with minimal efforts to meet a deadline. 
+
+## Decision
+
+What we have implemented:
+
+- On the publisher side for tagging, a single dropdown that merges the facets' values (main facet and sub facets, essentially leaf nodes)
+- When we send the document to Publishing API, we write logic into the document model to figure out whether the leaf node is a parent facet or sub-facet value in the dropdown, and split the metadata we send to Publishing API into its own facet and a separate sub-facet in `specialist-publisher/app/models/document.rb`
```
ChrisBAshton, line 29: Just checking whether what's written here is what's been implemented. The ADR implies (to me) that publishers can choose both "sub facets" and parent "facets" from the dropdown - but I'm only seeing subfacets in the dropdown on https://specialist-publisher.integration.publishing.service.gov.uk/trademark-decisions/new.

ChrisBAshton (commented):
Thanks for splitting this out into a separate PR. Some comments below :)

.adr-dir
minhngocd: i'm using adr tools https://github.com/npryce/adr-tools which auto generates the templates for the ADRs and it uses this file to figure out which folder the ADRs live in

minhngocd: will append to the commit message

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
minhngocd: hmm I'm not sure we clarified between 
- Having a mixture of leaf nodes - hence tagging a main-facet because it doesn't have sub-facets
- Tagging a main-facet only despite it also having sub-facets

I think in the future we do want to support tagging main-facet only despite having sub-facets, but we felt like we could clarify that later down the line, as the current MVP doesn't require it.

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
minhngocd: @ChrisBAshton similar to point above - in our implementation: 
- we allow main facets to be defined _without_ subfacets. It would show up in the dropdown if that is the case. Unfortunately for this finder, it doesn't have any main facets without sub facets. 
- what we don't allow, is if the main facet has sub-facets, then you can only choose the sub-facets (leaf nodes). Reason being we still have things to clarify about that behaviour (is it equivalent to tagging _all_ subs, or _no_ subs). Since the current finder doesn't require it, i thought de-scoping would be fine.

happy to discuss further

docs/adr/0001-implement-nested-facets-tagging-for-specialist-documents-as-a-single-dropdown.md
ChrisBAshton: Ahh fair enough.

ChrisBAshton (approved):
LGTM!

PR #3028, opened 2025-02-25
Change finder presenter logic for nested facets

lauraghiorghisor-tw:
Changing nested facets implementation. We now send two distinct facets in the publishing-api load, corresponding to the main and sub facet, respectively, in order to minimise how much we need to change Finder Frontend down the line.

[Trello](https://trello.com/c/skH2B2fH/3373-nested-facets-finder-frontend)

### DIFF
--- a/app/presenters/finder_content_item_presenter.rb
+++ b/app/presenters/finder_content_item_presenter.rb
@@ -27,14 +27,6 @@ class FinderContentItemPresenter
     file.fetch("content_id")
   end
 
-  def self.facets_without_specialist_publisher_properties(facets)
-    facets.reject { |facet| facet["specialist_publisher_properties"]&.fetch("omit_from_finder_content_item", false) }
-          .map do |facet|
-      facet.delete("specialist_publisher_properties")
-      facet
-    end
-  end
-
 private
 
   def title
@@ -60,7 +52,7 @@ private
       signup_link: file.fetch("signup_link", nil),
       summary: file.fetch("summary", nil),
       label_text: file.fetch("label_text", nil),
-      facets: FinderContentItemPresenter.facets_without_specialist_publisher_properties(file.fetch("facets", nil)),
+      facets: FinderFacetPresenter.new(file.fetch("facets", nil)).to_json,
       default_order: file.fetch("default_order", nil),
       default_documents_per_page: 50,
     }.reject { |_, value| value.nil? }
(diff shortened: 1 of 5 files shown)


app/presenters/finder_content_item_presenter.rb
```diff
@@ -60,12 +60,57 @@ def details
       signup_link: file.fetch("signup_link", nil),
       summary: file.fetch("summary", nil),
       label_text: file.fetch("label_text", nil),
-      facets: FinderContentItemPresenter.facets_without_specialist_publisher_properties(file.fetch("facets", nil)),
```
minhngocd, line 63: should we pull this out as a `FinderFacetPresenter` ?

app/presenters/finder_content_item_presenter.rb
```diff
@@ -60,12 +60,57 @@ def details
...
+        facets_including_nested << main_facet_hash_without_sub_facets(facet) << sub_facet_hash(facet)
+      else
+        facets_including_nested << facet
+      end
+    end
+  end
+
+  def main_facet_hash_without_sub_facets(facet_hash)
+    hash_dup = facet_hash.dup
+    hash_dup["allowed_values"] = hash_dup["allowed_values"].map { |v| v.except("sub_facets") }
```
minhngocd: can we not use hash.except(key)?

app/presenters/finder_content_item_presenter.rb
```diff
@@ -60,12 +60,57 @@ def details
...
+      "sub_facet_name" => nil,
+      "short_name" => facet_hash["sub_facet_name"],
+      "preposition" => facet_hash["sub_facet_name"],
+    ).compact
+  end
+
+  def sub_facet_allowed_values(facet_hash)
+    sub_facet_values_with_main_facet_reference = []
+
+    facet_hash["allowed_values"].each do |allowed_value|
```
minhngocd: instance of inject to replace the `[]` instantiation and return

app/presenters/finder_content_item_presenter.rb
```diff
@@ -27,14 +27,6 @@ def content_id
     file.fetch("content_id")
   end
 
-  def self.facets_without_specialist_publisher_properties(facets)
```
ChrisBAshton, line 30: It would be good to do the moving of this into FinderFacetPresenter in a separate commit to then adding the sub_facet_hash logic.

ChrisBAshton (approved):

app/presenters/finder_content_item_presenter.rb
minhngocd: Updated now. The refactoring is in a separate commit

PR #3029, opened 2025-02-26
Dry specialist publisher tests

GDSNewt:
Adding a separate test suite for each new finder is inefficient, especially when testing the same underlying code.

This PR consolidates multiple test files into a few test suites that iterate across all schema formats. This helps standardise finder tests, improve coverage, and automate more of the process for developers.

**Things to note**

1. Not a full refactor – Some document types don’t easily fit into a general test suite. These have been left unchanged and added to an exceptions list.
2. No new tests added – This PR focuses solely on consolidating existing tests, not filling coverage gaps.
3. Inconsistent lifecycle testing – While this improves coverage of the creation stage, post-creation actions are still primarily covered by the CMA cases tests.
4. May increase complacency - there is some reassurance in writing tests manually. There are scenarios where we may erroneously add, or forget to add, validation to our schemas. The automated tests will not be aware of such errors. Having said that, those issues should be picked up when testing the finder in integration. 

[Trello](https://trello.com/c/VfGijIuK/3412-dry-up-the-specialist-publisher-tests)


lib/documents/schemas/esi_funds.json
```diff
@@ -180,7 +180,7 @@
     }
   ],
   "filter": {
-    "format": "european_structural_investment_fund"
+    "format": "esi_fund"
```
ChrisBAshton: We can't (easily) change this as this value is used in Search API too and is associated with all the ESI specialist documents.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+          fill_in facet["name"], with: "Example #{facet['name']}"
+        end
+      end
+
+      case document_type.to_s
+      when "marine_equipment_approved_recommendation"
+        fill_in "Year adopted", with: "2025"
+      when "protected_food_drink_name"
+        fill_in "Time of registration (UK scheme)", with: "12:00"
+      end
```
ChrisBAshton: I think we need to have a red line on having any finder-specific conditionals in this general test (even if it means pulling these out into the `EXCEPTIONS_TO_GENERAL_TESTING` array)

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    scenario "getting to the new document page" do
+      visit document_path
+      click_link "Add another #{schema.document_title}"
+
+      expect(page.status_code).to eq(200)
+      expect(page.current_path).to eq(new_document_path)
+    end
+
+    scenario "with valid data" do
```
ChrisBAshton: This reads like a context block (`context "with valid data"`) - I'd then expect an assertion `it` block (or several) to be made further down (e.g. `it "saves the draft"`) to make it clear what we're testing.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    scenario "with no data" do
+      visit new_document_path
+      click_button "Save as draft"
+
+      expect(page.status_code).to eq(422)
+
+      schema.facets.each do |facet|
+        facet["key"]
```
ChrisBAshton: Can delete this line

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+        expect(page.status_code).to eq(422)
+        expect(page).to have_css(".elements-error-summary")
+        expect(page).to have_css(".elements-error-message")
+        expect(page).to have_content("not a valid date")
+      end
+    end
+  end
+
+  Dir["lib/documents/schemas/*.json"].each do |file|
```
ChrisBAshton, line 125: Nice job! 👏

spec/features/open_before_closed_validation_spec.rb
ChrisBAshton, line 1: I'd rename this file to something like `cma_cases_date_validation_spec.rb` to make it clear it's finder-specific.

ChrisBAshton: This could also be in its own separate commit.

spec/fixtures/factories.rb
```diff
@@ -842,4 +842,14 @@
...
+
+  factory :export_health_certificate, parent: :document do
+    base_path { "/export-health-certificates/example-document" }
+    document_type { "export_health_certificate" }
+  end
+
+  factory :traffic_commissioner_regulatory_decision, parent: :document do
+    base_path { "/traffic_commissioner_regulatory_decisions/example-document" }
+    document_type { "traffic_commissioner_regulatory_decision" }
+  end
```
ChrisBAshton: Adding these factories could be in their own separate commit (prior to adding the general tests), highlighting that these should have been added when they were first defined but that that step got missed).

spec/features/creating_a_cma_case_spec.rb
ChrisBAshton, line 1: Commit message `Delete unnecessarily duplicated "creating a ..." tests` needs updating, as it still references keeping some old tests, but the commit is entirely deletions.

spec/features/visiting_the_app_spec.rb
```diff
@@ -1,21 +0,0 @@
...
-    visit "/"
-    expect(page).to have_selector("h1", text: "CMA Cases")
-    expect(page).to have_selector("nav a", text: "CMA Cases")
-    expect(page).not_to have_content("AAIB Reports")
-    expect(page).not_to have_content("Manuals")
-  end
-
-  scenario "visiting any path should set an authenticated user header" do
-    visit "/"
-    expect(/uid-\d+/).to match(GdsApi::GovukHeaders.headers[:x_govuk_authenticated_user])
```
ChrisBAshton, line 19: This header test isn't covered elsewhere - so may still be worth keeping?

spec/fixtures/factories.rb
```diff
@@ -761,7 +761,7 @@
           "digital_market_research_publisher" => %w[gambling-commission],
           "digital_market_research_area" => %w[media-and-entertainment],
           "digital_market_research_topic" => %w[future-connectivity],
-          "digital_market_research_publish_date" => "2021-02-18T10:12:26+00:00",
```
ChrisBAshton, line 777: Has the behaviour changed or was this fixture data always wrong?

spec/models/business_finance_support_scheme_spec.rb
```diff
@@ -1,11 +0,0 @@
-require "spec_helper"
-require "models/valid_against_schema"
-
-RSpec.describe BusinessFinanceSupportScheme do
-  let(:payload) { FactoryBot.create(:business_finance_support_scheme) }
-  include_examples "it saves payloads that are valid against the 'specialist_document' schema"
-
-  it "is exportable" do
```
ChrisBAshton: I think we need to keep these tests, as we're not testing `self.exportable?` anywhere else.

(We could remove these tests if we write a test that an `exportable` boolean in the schema can be surfaced by `<Finder>.exportable?` but it doesn't look like that property exists in the schema yet  - it's hardcoded in the model).

spec/models/document_type_spec.rb
ChrisBAshton, line 1: I'd move this into its own commit - it'll make the PR easier to review.

I had drafted the following comment further up in this commit, because it looked as though you were deleting the schema validation tests without adding a new equivalent.

When you move this into its own commit, you can also use that as an opportunity to explain (in the message) why `ai_assurance_portfolio_technique` needs to be excluded.

```
I think we'll want to copy this sort of general schema validation check over to `spec/features/creating_a_new_document_spec.rb` (or some other generic test file). We want to ensure that all finder schemas are valid.
```

spec/models/protected_food_drink_name_spec.rb
```diff
@@ -1,27 +0,0 @@
...
-    it "has a valid base_path" do
-      expect(document.base_path).to eq("/protected-food-drink-names/sterea-ellada")
-    end
-  end
-
-  context "with Cyrillic characters" do
-    subject(:document) { ProtectedFoodDrinkName.new(title: "Черноморски район") }
-
-    it "has a valid base_path" do
-      expect(document.base_path).to eq("/protected-food-drink-names/chiernomorski-raion")
```
ChrisBAshton: Hmmm we don't have an equivalent for that sort of test, do we? May need to keep.

And if it isn't specific to this finder (from a quick look, doesn't seem to be) then we should also consider moving this into a generic test somewhere.

spec/models/residential_property_tribunal_decision_spec.rb
```diff
@@ -1,39 +0,0 @@
...
-      tribunal_decision.tribunal_decision_sub_category = nil
-      expect(tribunal_decision).not_to be_valid
-    end
-
-    it "is invalid if the tribunal decision sub category does not match up to the tribunal decision category" do
-      tribunal_decision.tribunal_decision_category = "rents"
-      tribunal_decision.tribunal_decision_sub_category = "tenant-associations---request-for-information"
-      expect(tribunal_decision).not_to be_valid
-    end
-  end
```
ChrisBAshton: Does this have an equivalent in the general tests? From a quick look, it seems this should be kept.

spec/models/trademark_decision_spec.rb
```diff
@@ -1,23 +0,0 @@
...
-  it "is not exportable" do
-    expect(described_class).not_to be_exportable
-  end
-
-  describe "validations" do
-    it "is valid from the payload" do
-      expect(trademark_decision).to be_valid
-    end
-
-    it "is invalid if the organisation is missing" do
```
ChrisBAshton, line 18: Agree this sort of thing should go - but we should probably have a generic test that a facet marked as 'required' must have a value specified.

ChrisBAshton (commented):
This is looking great, Alex - exactly the sort of thing I had in mind! Some comments below.

It would also be good to have an updated idea of what needs doing in the test space when creating a new finder. Are you drafting an update to <https://docs.publishing.service.gov.uk/repos/specialist-publisher/creating-editing-and-removing-specialist-document-types-and-finders.html>?
I've just raised PR 3011 to add the HMRC Finder and from a quick look at this PR, I think all I'll need to do is add a factory for it - is that right?

spec/features/creating_a_new_document_spec.rb
GDSNewt: Fair enough, as I note in the commit message it did feel weird doing it but I thought it was 'small enough' not to be too offensive! 😅

lib/documents/schemas/esi_funds.json
GDSNewt: Yeah was worried that might be the case, format was a convenient value to use from the schema to set the document type in the test. I'm trying to add a conditional, in the test, to do this but having some issues accessing DOM elements atm 😢

spec/features/creating_a_new_document_spec.rb
GDSNewt: I'll change the scenario descriptions to read a bit clearer

spec/features/creating_a_new_document_spec.rb
GDSNewt: Thanks, remnant of old code

spec/features/open_before_closed_validation_spec.rb
GDSNewt: I chose to make it generic as AnimalDiseaseCase also uses the OpenBeforeClosedValidator

spec/fixtures/factories.rb
GDSNewt: I believe it must have always been wrong, when I ran this through the 'valid_against_schema' shared example it complained about the time element. 

I don't think this was caught previously because the drcf finder didn't have it's own model spec file to check it

GDSNewt: Same goes for animal disease case

spec/models/business_finance_support_scheme_spec.rb
GDSNewt: Ah yep, you're right. Think I removed by mistake!

spec/features/open_before_closed_validation_spec.rb
ChrisBAshton: Ah, didn't spot that - fair enough.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        fill_in "Time of registration (UK scheme)", with: "12:00"
+      end
+
+      expect(page).to have_css("div.govspeak-help")
+      expect(page).to have_content("To add an attachment, please save the draft first.")
+      expect(save_button_disable_with_message).to eq("Saving...")
+      click_button "Save as draft"
+
+      expect(page.status_code).to eq(200)
+      expect(page).to have_content("Created Example #{document_type.to_s.humanize}")
```
lauraghiorghisor-tw, line 75: Since we're looping all through all the facets and filling them in, it feels a bit like we've done that for nothing if we don't also include any expectations on that data. 

Nonetheless, expecting on ALL of that might slow down things a little bit - perhaps it's enough to only fill out the required fields at this point and thus expect that we have created the document only - or also expect on the filled in fields.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+    scenario "with valid data" do
+      visit new_document_path
+      fill_in "Title", with: "Example #{document_type.to_s.humanize}"
+      fill_in "Summary", with: "Example Summary"
+      fill_in "Body", with: "Example Body"
+
+      schema.facets.each do |facet|
+        key = facet["key"]
+        properties = facet["specialist_publisher_properties"] || {}
+        properties["validations"] || {}
```
lauraghiorghisor-tw, line 49: Redundant line.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      click_button "Save as draft"
+
+      expect(page.status_code).to eq(422)
+
+      schema.facets.each do |facet|
+        facet["key"]
+        properties = facet["specialist_publisher_properties"] || {}
+        validations = properties["validations"] || {}
+
+        if validations["presence"]
```
lauraghiorghisor-tw, line 89: This test doesn't actually test anything, since this line will never evaluate to true as there is no `presence` key in the schema. Apologies, I have called it `required`, just to be fancy 🙈

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+  end
+
+  Dir["lib/documents/schemas/*.json"].each do |file|
+    schema = JSON.parse(File.read(file))
+    format = schema["filter"]["format"]
+
+    next if EXCEPTIONS_TO_GENERAL_TESTING.include?(format)
+
+    describe "Creating a #{format.humanize}" do
+      base_path = format == "product_safety_alert_report_recall" ? "/product-safety-alerts-reports-recalls" : "/#{format.pluralize.dasherize}"
```
lauraghiorghisor-tw, line 139: The one convention we do uphold across the codebase, though I'd be hard pressed to tell you why it matters that we do so...,  is that our models are called, for example `ProductSafetyAlertReportRecall` which must (as per Rails) live in a file called `product_safety_alert_report_recall.rb` whose corresponding json file is called `product_safety_alert_report_recalls.json` - PLURALIZED. 

You'd need to do a bit of string manipulation here, but it might be a good bet to use the `file` variable here rather than the format from the schema 🤔

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      include_context "common setup", :gds_editor, format.to_sym, base_path, "#{base_path}/new"
```
lauraghiorghisor-tw, line 140: Is `base_path/new` required as its own parameter?

app/models/ai_assurance_portfolio_technique.rb
```diff
@@ -1,4 +1,5 @@
 class AiAssurancePortfolioTechnique < Document
+  apply_validations
```
lauraghiorghisor-tw, line 2: We did not add `apply_validations` to models that simply did not have any validation. They did not have any presence checks, nor any date fields. I believe it is still redundant to add this.

spec/features/creating_a_new_document_spec.rb
lauraghiorghisor-tw: This test doesn't actually test anything, since this line will never evaluate to true as there is no `presence` key in the schema. Apologies, I have called it `required`, just to be fancy 🙈

spec/features/creating_a_new_document_spec.rb
lauraghiorghisor-tw, line 69: Since we're looping through all the facets and filling them in, it feels a bit like we've done that for nothing if we don't also include any expectations on that data.

Nonetheless, expecting on ALL of that might slow down things a little bit - perhaps it's enough to only fill out **only** the required fields at this point and thus expect that we have created the document only - or also expect on the filled in fields.

spec/features/creating_a_new_document_spec.rb
lauraghiorghisor-tw: Redundant line?

app/models/ai_assurance_portfolio_technique.rb
lauraghiorghisor-tw, line 2: We did not add `apply_validations` to models that simply did not have any validation. They did not have any presence checks, nor any date fields (the only schema validations we added).

I believe it is still redundant to add this. I understand the "completeness" aspect, but it might also read as, "ah this thing should have validations", but then you look in the schema and there are none 🕵🏻‍♀️ Up to you though.

spec/features/creating_a_new_document_spec.rb
lauraghiorghisor-tw: The one convention we do uphold across the codebase, though I'd be hard pressed to tell you why it matters that we do so..., is that our models are called, for example `ProductSafetyAlertReportRecall` which must (as per Rails) live in a file called `product_safety_alert_report_recall.rb` whose corresponding json file is called `product_safety_alert_report_recalls.json` - PLURALIZED.

You'd need to do a bit of string manipulation here, but it might be a good bet to use the file variable here rather than the format from the schema 🤔

That should also fix your `esi_funds`.

Alternatively we can maintain a manual map of formats for the test.

spec/features/creating_a_new_document_spec.rb
lauraghiorghisor-tw, line 139: Is `base_path/new` required as its own parameter?

spec/models/document_type_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  ai_assurance_portfolio_technique
+].freeze
+
+Dir["lib/documents/schemas/*.json"].each do |file|
+  schema = JSON.parse(File.read(file))
+  format = schema["filter"]["format"]
+
+  next if EXCEPTIONS_TO_GENERAL_TESTING.include?(format)
+
+  RSpec.describe format.classify.constantize do
```
lauraghiorghisor-tw, line 15: I would actually quite like the validation tests to live here, rather than in the feature specs.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
+require "spec_helper"
+
+EXCEPTIONS_TO_GENERAL_TESTING = %w[
+  ai_assurance_portfolio_technique
```
lauraghiorghisor-tw: I did not quite follow the reasoning behind this. Why would we have a non-live finder? Seems live https://www.gov.uk/ai-assurance-techniques.

spec/features/creating_a_cma_case_spec.rb
```diff
@@ -1,403 +0,0 @@
...
-    select "Energy", from: "Market sector"
-
-    click_button "Save as draft"
-
-    expect(page.status_code).to eq(200)
-
-    expect(page).to have_content("Created Example CMA Case")
-  end
-
-  scenario "with a very long title" do
```
lauraghiorghisor-tw, line 304: Not carried over.

spec/features/creating_a_cma_case_spec.rb
```diff
@@ -1,403 +0,0 @@
...
-        "finder" => %w[fef4ac7c-024a-4943-9f19-e85a8369a1f3],
-      },
-    }
-    assert_publishing_api_put_content(content_id, expected_sent_payload)
-
-    expect(page.status_code).to eq(200)
-    expect(page).to have_content("At veroeos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpa qui officia deserunt mollitia animi")
-  end
-
-  scenario "previewing GovSpeak", js: true do
```
lauraghiorghisor-tw, line 362: Not carried over.

spec/features/creating_a_cma_case_spec.rb
```diff
@@ -1,403 +0,0 @@
...
-    click_link "Preview"
-
-    within(".preview_container") do
-      expect(page).to have_content("link text")
-      expect(page).not_to have_content("http://www.example.com")
-      expect(page).not_to have_content("some text")
-    end
-  end
-
-  scenario "a draft with the same path as an existing draft" do
```
lauraghiorghisor-tw, line 385: Not carried over.

spec/features/access_control_spec.rb
```diff
@@ -1,126 +1,60 @@
 require "spec_helper"
 
 RSpec.feature "Access control", type: :feature do
```
lauraghiorghisor-tw, line 3: There are a few specs in the now deleted `creating_an_employment_appeal_tribunal_decisions_spec.rb` that rather nicely covered editor and writer access - see `context "when logged in as an editor"` and `context "when logged in as a writer" `. Made me think it would be good to have a complete account of these permissions in this test.

spec/features/creating_a_residential_property_tribunal_decision_spec.rb
```diff
@@ -1,86 +0,0 @@
...
-    fill_in "Body", with: "<script>alert('hello')</script>"
-    select "Rents", from: "Category"
-    select "Park homes - Site licence - payment of annual fee", from: "Sub-category"
-
-    click_button "Save as draft"
-
-    expect(page.status_code).to eq(422)
-
-    expect(page).to have_content("Body cannot include invalid Govspeak")
-    expect(page).to have_content("Tribunal decision sub category must belong to the selected tribunal decision category")
```
lauraghiorghisor-tw, line 84: Noting the `ResidentialPropertyTribunalDecisionSubCategoryValidator` test here - but this validator does have unit tests so probably ok.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+  business_finance_support_scheme
+  flood_and_coastal_erosion_risk_management_research_report
+  licence_transaction
+  research_for_development_output
+  statutory_instrument
+].freeze
+
+RSpec.feature "Creating a document", type: :feature do
+  shared_context "common setup" do |editor, document_type, document_path, new_document_path|
```
lauraghiorghisor-tw, line 15: From what I've seen around, it is customary to have the shared context in its own file. Would also mean that it's a tad easier to get to the important part of this file, which uses the shared context some 100 lines down.

spec/features/creating_a_new_document_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+      click_button "Save as draft"
+
+      expect(page.status_code).to eq(422)
+      expect(page).to have_css(".elements-error-summary")
+      expect(page).to have_css(".elements-error-message")
+
+      expect(page).to have_content("Body cannot include invalid Govspeak")
+    end
+
+    scenario "with invalid date" do
```
lauraghiorghisor-tw: Date validation is applied universally based on the type of the facet in the schema. Perhaps it is not that much value iterating the check per document type.

Maybe it suffices to extract this into its own feature spec file, as you did with open before close validation. And it would make it more visible that that is where our date validation logic is checked.

lauraghiorghisor-tw (changes requested):
Left some comments, the one requiring change is around the presence validation tests which is currently not running anything. 

I second Chris's recommendation around documentation. I'd particularly like to know what the best approach is for adding anything custom. Or what a TDD flow can look like. BUT, maybe that's something we first figure out by using the repo for a while and then update the docs ;) 

Since these tests are now running a bit magically perhaps improving the visualisation of the test run could help. Maybe we can change the .rspec development environment settings to run with `--format doc` flag on. This would then render as below:
> Screenshot 2025-03-03 at 11 50 10.

spec/features/access_control_spec.rb
GDSNewt: Will update the commit message to say that this file will cover all editor permissions and that writer permission is covered by the actions presenter spec.

lib/documents/schemas/esi_funds.json
GDSNewt: I'm conceding defeat on esi_funds, at least as far as this PR goes. Wrangling this is holding up the rest of the PR, so i'll make a note of this in the PR description and add esi_funds to the exception list.

spec/models/protected_food_drink_name_spec.rb
GDSNewt: As far as I can tell it's effectively a test of the stringex gem

https://github.com/alphagov/specialist-publisher/pull/1773

I can't see anywhere else we use it, I suppose the question is do we trust the stringex gem?

spec/features/creating_a_new_document_spec.rb
GDSNewt: I'm adopting the 'format = file name from the schemas', as I take your point that 'in theory' it should be more consistent. 

I don't think we gain any wins with regards to the finders with weird base paths i.e 

product_safety_alert_report_recall
ai_assurance_portfolio_technique

As the issues relating to them are 'format and file name agnostic'.

Similarly it doesn't quite fix esi funds, as there is an issue with some of the labels in the DOM - so I will just have to exempt esi funds for now.

spec/models/residential_property_tribunal_decision_spec.rb
GDSNewt: Yep, i'm removed the 'nil' validations but kept the last one

spec/models/protected_food_drink_name_spec.rb
GDSNewt: I will keep it for the unlikely scenario in which Stringex's  transliteration code changes

app/models/ai_assurance_portfolio_technique.rb
GDSNewt: My audit of the schemas that have required fields are as such. 

1. algorithmic_transparency_record
2. animal_disease_case
3. asylum_support_decision
4. cma_case
5. drcf_digital_markets_research
6. employment_appeal_tribunal_decision
7. employment_tribunal_decision
8. export_health_certificate
9. farming_grant
10. life_saving_maritime_appliance_service_station
11. maib_report
12. marine_notice
13. medical_safety_alert
14. product_safety_alert_report_recall
15. raib_report
16. residential_property_tribunal_decision
17. sfo_case
18. tax_tribunal_decision
19. traffic_commissioner_regulatory_decision
20. trademark_decision
21. utaac_decision

My only hesitation in removing the 'apply validation' for models not in this list is the 'complacency' factor, i.e should they ever attain fields that do need validation we may forget to add 'apply validation to the model'.

I don't feel super strongly about this, I think this small bit of redundancy is permissible

spec/features/creating_a_cma_case_spec.rb
GDSNewt: I think this is covered in document_spec.rb 

  ```
context "saving a draft where another draft has the same base_path" do
    before do
      stub_any_publishing_api_put_content
        .to_raise(GdsApi::HTTPErrorResponse.new(422, "base path=/foo/bar conflicts with content_id=123"))
    end

    subject do
      MyDocumentType.new(
        title: "A document",
        summary: "An introduction",
        body: "Some text",
      )
    end

    it "cannot be saved" do
      expect(subject.save).to be false
    end

    it "populates an error on the document" do
      subject.save
      expect(subject.errors[:base]).to eq(["Warning: This document's URL is already used on GOV.UK. You can't publish it until you change the title."])
    end`

```

File: spec/features/creating_a_new_document_spec.rb (hunk 26 above)
Comment (GDSNewt) [reply spec/features/creating_a_new_document_spec.rb:15]: I tried to get this working but it proved very awkward and imo, made it less clear what was happening. I could see value in having a truly separate shared context were I to use it across multiple files (such as the proposed date validator), but i'm reluctant to do that - for reasons I lay out in the next comment.

File: spec/features/creating_a_new_document_spec.rb (hunk 27 above)
Comment (GDSNewt) [reply spec/features/creating_a_new_document_spec.rb]: You're right but I think that the same argument could be said for the 'required' validations, in which case we probably only need to test one document type - maybe that's what we should be doing?

In the spirit of testing every document type anyway, I don't think it's too egregious to check their date facet at the same time.

File: app/models/ai_assurance_portfolio_technique.rb (hunk 18 above)
Comment (lauraghiorghisor-tw) [reply app/models/ai_assurance_portfolio_technique.rb:2]: I suppose in that case you may never know that the validation tests are not being run, true.

File: spec/features/creating_a_new_document_spec.rb (hunk 15 above)
Comment (lauraghiorghisor-tw) [reply spec/features/creating_a_new_document_spec.rb]: What I mean is that 
```
        if validations["presence"]
          expect(page).to have_content("#{facet['name']} serfdgthyjukilo;pkjhgfdscan't be blank")
        end
```
still passes  - because the key is called `required`.

lauraghiorghisor-tw (approved):

app/models/ai_assurance_portfolio_technique.rb
ChrisBAshton: +1 we should just have `apply_validations` _everywhere_ and not worry that sometimes it doesn't really do anything 👍 consistency is best

PR #3035, opened 2025-03-05
Nested facets for admin form

lauraghiorghisor-tw:
This PR introduces the ability to add and edit nested facets in the admin form.
We decided to not do a full design for this feature, but provide a developer-readable solution as MVP. 

### Nested Facet
- Note the optional input that takes both Facet name and key (key on edit only), and the hint including instructions on subfacets, and example.
> Screenshot 2025-03-07 at 14 31 10

[Trello](https://trello.com/c/IG0Kb6XU/3372-nested-facets-edit-add-forms)


spec/helpers/facet_select_helper_spec.rb
```diff
@@ -63,4 +63,46 @@
       expect(select_options_for_facet(facets)).to eq expected_select_dropdown_values
     end
   end
+
+  describe "#admin_facet_value_from_allowed_values" do
```
ChrisBAshton, line 67: My first thought on reading the PR description was that blank lines feel somewhat brittle and unintuitive 🤔 I wondered if dashes (like a list) would make more sense? (And we could still allow blank lines for readability of input, but strip them out on submit)

Before:

```
Section 3(1) Graphical Representation {section-3-1-graphical-representation}
Is it a sign? {section-3-1-graphical-representation-is-it-a-sign}
Foo {section-3-1-foo}

Section 3(2) Something else {section-3-2-something-else}
Bar {section-3-2-bar}
Foo {section-3-2-foo}
```

After:

```
Section 3(1) Graphical Representation {section-3-1-graphical-representation}
- Is it a sign? {section-3-1-graphical-representation-is-it-a-sign}
- Foo {section-3-1-foo}

Section 3(2) Something else {section-3-2-something-else}
- Bar {section-3-2-bar}
- Foo {section-3-2-foo}
```

Advantages:
- Clearer at a glance which are the children and which is the parent
- Nicer code & regexes? (Anything with backslash `\r\n`s can be a bit of a nightmare)

app/models/facet.rb
```diff
@@ -36,8 +42,11 @@ def from_finder_admin_form_params(params)
       facet.preposition = nil_if_blank(params["preposition"])
       facet.display_as_result_metadata = params["display_as_result_metadata"]
       facet.filterable = params["filterable"]
-      facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"])
+      facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"], nil_if_false(params["nested_facet"]))
       facet.specialist_publisher_properties = facet_specialist_publisher_properties(params["type"], params["validations"])
+      facet.nested_facet = nil_if_false(params["nested_facet"])
```
ChrisBAshton: Hmmm these two fields do a similar thing, don't they?

The "nested_facet" radio button determines whether or not this is a nested facet.
And the "sub_facet" text input gives the name of the sub facet, which is a required property of a nested facet.

We could get into a confusing situation where someone provides a sub facet value, but chooses "No" on the radio. Or chooses "Yes" on the radio, but doesn't provide a sub facet value.

Shall we get rid of the radio and just make it dependent on the "sub facet" text input?

app/models/facet.rb
```diff
@@ -36,8 +42,11 @@ def from_finder_admin_form_params(params)
...
+      facet.sub_facet_name = extract_label_and_value(params["sub_facet"], "_").first if params["sub_facet"].present?
+      facet.sub_facet_key = facet_key(extract_label_and_value(params["sub_facet"], "_").last, facet.sub_facet_name) if params["sub_facet"].present?
```
ChrisBAshton, line 52: Does this work? 🙏 

```suggestion
      [facet.sub_facet_name, facet.sub_facet_key] = extract_label_and_value(params["sub_facet"], "_") if params["sub_facet"].present?
```

app/models/facet.rb
ChrisBAshton: (Would also mean we could delete the `nil_if_false` code)

app/views/admin/_facet_fields.html.erb
```diff
@@ -153,3 +153,23 @@
   value: facet["preposition"],
   hint: "Example: In Find funding for land or farms,  selecting an option in the filter ‘Area of interest’ displays the preposition ‘With’ before the selected option"
 } %>
+
+<%= render "govuk_publishing_components/components/radio", {
+  heading: "Should filter values be searchable?",
```
ChrisBAshton: I'm not clear what this does 🤔 (either from reading the commit or the PR description)

I don't see any searchable behaviour on the options in the dropdown on https://draft-origin.publishing.service.gov.uk/trademark-decisions?

spec/features/editing_the_trademark_decision_filters_and_options_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    expect(page).to have_selector("span", text: "Trademark Decision finder")
+    expect(page).to have_selector("h1", text: "Request change: Filters and options")
+
+    # Delete all but the first filter
+    all(".js-add-another__remove-button").last.click while all(".js-add-another__remove-button").count > 1
+    click_button "Add another filter"
+    within find("fieldset", text: "Filter 2") do
+      # 2 in "Filter 2" is the visible index of the newly inserted filter, since we deleted all but the first one, and now we're adding a new one
```
ChrisBAshton: I think the code is pretty much self-documenting, but we can move your comment above and shorten it to something like:

```suggestion
    # Fill in the newly created filter
    within find("fieldset", text: "Filter 2") do
```

spec/features/editing_the_trademark_decision_filters_and_options_spec.rb
```diff
@@ -27,15 +27,37 @@
     expect(page).to have_selector("span", text: "Trademark Decision finder")
     expect(page).to have_selector("h1", text: "Request change: Filters and options")
 
+    # Expect not to find the conditionally rendered fields on a non-filterable facet
```
ChrisBAshton: What happens if the user changes the facet to be filterable? I'm guessing the fields still aren't visible, until the 'filterable' change has been applied to the schema?

In which case, I'm not sure it's worth the overhead of showing/hiding these fields 🤔

lib/documents/schemas/trademark_decisions.json
```diff
@@ -274,12 +274,12 @@
       "key": "trademark_decision_date",
       "name": "Decision date",
       "preposition": "Decision date",
+      "short_name": "Date",
```
ChrisBAshton: Nice. I'd be tempted to pull the reorder commits out into a separate PR, since they're not connected to the nested facets work and are something that provide value in isolation.

spec/features/generating_a_new_finder_spec.rb
ChrisBAshton, line 1: Oops - [my bad](https://github.com/alphagov/specialist-publisher/blame/b2b06313d7d90932ff6b19d1a83d442e5c7757be/spec/features/generating_a_new_finder.rb)!!

app/models/facet.rb
```diff
@@ -106,7 +106,7 @@ def extract_nested_allowed_values(values_string)
     def extract_label_and_value(str, gsub_character)
       label = str.match(/^(.+){/)
       label = label.nil? ? str.strip : label[1].strip
-      value = str.match(/{(.+)}/)
+      value = str.truncate(500, omission: "}").match(/{(.+)}/)
```
ChrisBAshton, line 104: Nice - I'd not come across the `omission` arg before!

app/views/admin/_facet_fields.html.erb
ChrisBAshton: Unless you're referring to the search box at the top of the left hand side? In which case "search box for users to filter the list options" doesn't make sense (they're filtering results, rather than options)

ChrisBAshton (commented):
Well done for ploughing on with this. Some questions / suggestions below.

spec/features/editing_the_trademark_decision_filters_and_options_spec.rb
lauraghiorghisor-tw: Fair enough, will drop that commit, and keep the relevant bit of the test.

spec/helpers/facet_select_helper_spec.rb
lauraghiorghisor-tw: Nice, will do!

app/models/facet.rb
lauraghiorghisor-tw: Good shout.

app/views/admin/_facet_fields.html.erb
lauraghiorghisor-tw: It's this rather unrelated thing - but I can't pass the feature test without it, as the generated schema will include it.

Screenshot 2025-03-07 at 11 48 55

lauraghiorghisor-tw: I can add it as a separate PR as well, with minimal tests?

spec/features/editing_the_trademark_decision_filters_and_options_spec.rb
lauraghiorghisor-tw: It was the indices that were tripping me. Better off with no comment.

lib/documents/schemas/trademark_decisions.json
lauraghiorghisor-tw: DONE https://github.com/alphagov/specialist-publisher/pull/3042

app/models/facet.rb
lauraghiorghisor-tw: I don't think so, because I also want to call the `facet_key` method.

app/models/facet.rb
lauraghiorghisor-tw: Did not manage to remove the `nil_if_false` as used in the other PR that got merged prior, for the `show_option_select_filter`.

app/views/admin/_facet_fields.html.erb
lauraghiorghisor-tw: DONE here https://github.com/alphagov/specialist-publisher/pull/3043.

app/models/facet.rb
```diff
@@ -38,9 +44,12 @@ def from_finder_admin_form_params(params)
       facet.preposition = nil_if_blank(params["preposition"])
       facet.display_as_result_metadata = params["display_as_result_metadata"]
       facet.filterable = params["filterable"]
-      facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"])
+      facet.allowed_values = facet_allowed_values(params["allowed_values"], params["type"], nil_if_blank(params["sub_facet"]))
```
lauraghiorghisor-tw: I am calling this `nil_if_blank(params["sub_facet"])` twice when I could use the `facet.nested_facet` - but the code could get shuffled around so that initialisation does not happen anymore. Probably best to just check the `params`.

app/models/facet.rb
```diff
@@ -62,18 +71,47 @@ def facet_type(type)
...
+
+    def extract_allowed_values(values_string)
+      values_string.split("\n").map do |str|
+        label, value = extract_label_and_value(str, "-")
         { label:, value: }
       end
     end
 
+    def extract_nested_allowed_values(values_string)
+      values_string.split("\r\n\r\n").map do |facet_block|
```
lauraghiorghisor-tw: `"\r\n\r\n"` was the only thing that worked.

app/views/admin/_facet_fields.html.erb
```diff
@@ -69,10 +69,10 @@
     text: "Filter options ('Multiple select' or 'One option' only)",
     heading_size: "s",
   },
-  hint: sanitize("Put each option on a new line. The underlying name for existing (live) options will appear in curly braces: please don't edit these, and don't add them for new options (they'll be created automatically later on in the process). Example:<br><strong>Pre-existing value {pre-existing-value}</strong><br><strong>New value</strong>"),
+  hint: sanitize("Put each option on a new line. The underlying name for existing (live) options will appear in curly braces: please don't edit these, and don't add them for new options (they'll be created automatically later on in the process). Example:<br><strong>Pre-existing value {pre-existing-value}</strong><br><strong>New value</strong><br>If this is a nested facet, add nested options on subsequent lines, and leave a blank line between main facets:<br><strong>Main Facet Value</strong><br><strong>Sub Facet Value</strong><br><em>this is a blank line</em><br><strong>Another Main Facet Value</strong>"),
```
ChrisBAshton: I'd suggest that the "leave a blank line" instruction should be dropped. I think users should be welcome to add optional newline for formatting, but ultimately they get stripped when parsing the content.
And I imagine when we re-render the options in the form (from an existing finder) we'd lose the blank lines - and that's ok.

app/models/facet.rb
ChrisBAshton: If we're happy stripping the whitespace, new method would look something like:

```
def extract_nested_allowed_values(values_string)
  result = []
  current_main = nil

  values_string.each_line do |line|
    line.strip!
    if line.match(/^(.+?) \{(.+?)\}$/)
      label, value = [$1, $2]
      if line.start_with?('-')
        current_main[:sub_facets] << { label: label.sub('- ', ''), value: value } if current_main
      else
        current_main = { label: label, value: value, sub_facets: [] }
        result << current_main
      end
    end
  end

  result
end
```

Test:

```
values_string = """
Foo {foo}
- bar {foo-bar}
- baz {foo-baz}
something-else {something-else}
- banana {something-else-banana}
- orange {something-else-orange}
"""

puts extract_nested_allowed_values(values_string).inspect
```

app/views/admin/_facet_fields.html.erb
```diff
@@ -64,12 +64,22 @@
...
 } %>
 
+<%= render "govuk_publishing_components/components/input", {
+  label: {
+    text: "Sub-facet (Optional)",
+    heading_size: "s",
+  },
+  name: "facets[#{index}][sub_facet]",
+  value: facet["nested_facet"] ? "#{facet["sub_facet_name"]} {#{facet["sub_facet_key"]}}" : "",
+  hint: "The name of the sub-facet. Will show as 'name {key}'. Only relevant for nested facets. Only input the name on a new form, the key will be autogenerated. Do not edit the key."
```
ChrisBAshton:
```suggestion
  hint: "The name of the sub-facet (only relevant for nested facets). Once saved, a key will be generated alongside it in the format 'Name {auto-generated-key-name}'. Do not edit the key."
```

ChrisBAshton (approved):
LGTM 🎉 A couple of non-blocking suggestions below

app/models/facet.rb
lauraghiorghisor-tw: Pushed smth along these lines. It also helped get rid of one of the methods, so there's no `extract_nested_allowed_values` anymore, just `extract_allowed_values`.

ChrisBAshton (approved):
⭐

PR #3043, opened 2025-03-07
Include 'show_option_select_filter' attribute in schema generation

lauraghiorghisor-tw:
This commit adds `show_option_select_filter` to the finder admin UI, and thus generates it as part of the proposed schema for the Zendesk load. This is an option we only use for Licence finder and Trademark Decisions. But it is handy to have it included so that we don't get unnecessary diffs when editing those finders.

It's also handy to have it as a viable option on the admin UI. This option enables the kind of search seen for Appointed person/ hearing officer [here](https://draft-origin.publishing.service.gov.uk/trademark-decisions). 

In Specialist Publisher:
> Screenshot 2025-03-05 at 15 50 11

On Finder Frontend:
> Screenshot 2025-03-07 at 11 48 55

### DIFF
--- a/spec/models/facet_spec.rb
+++ b/spec/models/facet_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe "Facet" do
         "preposition" => "sector",
         "display_as_result_metadata" => "false",
         "filterable" => "true",
+        "show_option_select_filter" => "true",
         "allowed_values" => "existing value {existing-value}\nnew value",
       }
       facet = Facet.from_finder_admin_form_params(params)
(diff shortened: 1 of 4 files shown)


spec/models/facet_spec.rb
```diff
@@ -69,6 +71,16 @@
         facet = Facet.from_finder_admin_form_params({ "filterable" => "" })
         expect(facet.filterable).to eq(nil)
       end
+
+      it "converts 'true' to true for 'show_option_select_filter'" do
+        facet = Facet.from_finder_admin_form_params({ "show_option_select_filter" => "true" })
+        expect(facet.show_option_select_filter).to eq(true)
+      end
+
+      it "sets 'show_option_select_filter' to 'nil' if 'false' is passed" do
```
ChrisBAshton, line 80: Just to be clear, we're setting it to `nil` so that in the eventual diff, `show_option_select_filter` is omitted entirely? Whereas if we set it to `false`, it would be included?

ChrisBAshton (approved):
LGTM

spec/models/facet_spec.rb
lauraghiorghisor-tw: yep!

PR #3049, opened 2025-03-12
Nested facet - add custom nested type and change presenter logic

lauraghiorghisor-tw:
Changes:
- introduce custom "nested" type to replace `nested_facet` flag
- revert presenter logic so that we send a single cohesive "nested" load to Publishing API.

### Admin area changes
- New type value
> Screenshot 2025-03-13 at 10 37 18

- Existing Grounds section facet UI
> Screenshot 2025-03-13 at 10 32 24

- Existing grounds section facet generated schema in admin area (no change other than type)
> Screenshot 2025-03-12 at 13 18 16

- New facet added with nested multiple type - diff in admin area (no change other than type)
> Screenshot 2025-03-12 at 13 19 07

- New facet added with nested single type - diff in admin area (no change other than type)
> Screenshot 2025-03-13 at 10 36 28

[Trello](https://trello.com/c/EeZOA3iW/3534-spike-replicating-taxons-approach-for-nested-facets)

### DIFF
--- a/app/models/facet.rb
+++ b/app/models/facet.rb
@@ -65,7 +65,13 @@ class Facet
     end
 
     def facet_type(type)
-      facet_text_types.include?(type) ? "text" : type
+      if facet_text_types.include?(type)
+        "text"
+      elsif facet_nested_types.include?(type)
+        "nested"
+      else
+        type
+      end
     end
 
     def facet_allowed_values(values, type)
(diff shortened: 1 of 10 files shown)


app/models/facet.rb
```diff
@@ -115,7 +112,7 @@ def facet_specialist_publisher_properties(type, validations)
 
     def facet_specialist_publisher_properties_select(type)
       case type
-      when "enum_text_multiple"
```
minhngocd, line 118: nested facets could be used for both single select and multiselect fields, no?

lauraghiorghisor-tw: Ah, so add... two options here? Nested single and nested multiple?

lauraghiorghisor-tw: Pushed as a new commit (last commit).

lauraghiorghisor-tw: And added new screenshots.

minhngocd (approved):
LGTM - out of scope for this PR but maybe we need to revisit view logic splitting from models at some point as i had a little bit of trouble following the "type" changes.

PR #3076, opened 2025-03-28
Make blue metadata box optional

GDSNewt:
This change updates the backend to expose the show_metadata_block property to frontend applications.

The property is now defined at the schema level, providing more explicit control over its rendering. Previously, the metadata block was only hidden in specific cases where "important metadata" was absent.

Tests in this PR depend on example schemas from the Publishing API repository, which may cause CI failures until the related changes are merged.

**Associated PRs**

[Government-Frontend](https://github.com/alphagov/government-frontend/pull/3639)
[Publishing API](https://github.com/alphagov/publishing-api/pull/3241)

[Trello](https://trello.com/c/spUCgJtQ/3539-make-it-possible-to-hide-the-blue-box-on-specialist-documents)


ChrisBAshton (approved):
Looks good, Alex. Personally I'd have flipped the relationship (`hide_metadata_block: true`) and only updated one finder config instead of multiple, but it's not a blocker. Let's go with this 👍

PR #3084, opened 2025-04-07
Add feature flag for design system

minhngocd:
We are replicating the process we took with Whitehall to use user permissions as a feature flag to transition to the new design system. This will help us implement changes without breaking existing tests and flows while we build out new design system components. 

For testing locally, add `preview_design_system` permission to your local user. 
```
u = User.last
u.permissions << "preview_design_system"
u.save
```

https://trello.com/c/764XzLue/3419-design-system-add-new-document

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

Follow [these steps](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html) if you are doing a Rails upgrade.

PR #3088, opened 2025-04-09
New document design system

minhngocd:
## What

Transition the “Add new document” page template to the Design System. [Trello](https://trello.com/c/764XzLue/3419-design-system-add-new-document)

## Approach

This PR is the first step of a multi-step process. 
The first step covered here is to replace the existing fields with design system components and/or publishing gem components. 

1. Feature Flag was added in [the previous PR](https://github.com/alphagov/specialist-publisher/pull/3084) via user permissions to hide new design system work in progress while we transition pages. Layout and view renders are toggled within controllers. 
2. For any new view that is touched to change to the design system, a duplicate `*_legacy` version is created and moved to a legacy folder
3. Any associated tests are duplicated and tests for the new design systems are created as `*_design_system_spec` and log in via the helper methods such as `log_in_as_design_system_editor`. legacy tests are preserved for regression testing
4. We are only dealing with styling of new document page fields for this PR and creating a new page happy path. 
5. The following has been descoped into other tickets : 
  * Preview button working and looking correct
  * Retaining data previously entered in fields
  * validation & error messages
  * Add another full functionality

## Why

This is required as part of ongoing improvements to adopt Design System standards across the publishing apps. This will also enable us to implement new scheduling functionality for Specialist Publisher that is consistent with Whitehall

## Screenshots

With no `preview_design_system` permissions
image

With `preview_design_system` permissions
image

Add another component
image

Add another component restyled for Request new finder form
image

New replaced javascript length counter for summary
image


app/views/documents/new_legacy.html.erb
ChrisBAshton, line 1: In the interests of "optimising code for deletion", it might be cleaner to encapsulate these legacy templates in a subdirectory, i.e. `app/views/documents/legacy/new.html.erb`

app/views/documents/new.html.erb
```diff
@@ -1,8 +1,19 @@
...
+  <%= render "govuk_publishing_components/components/breadcrumbs", {
+    collapse_on_mobile: true,
+    breadcrumbs: [
+      {
+        title: "Your documents",
+        url: documents_path(current_format.admin_slug)
+      },
+      {
+        title: "New document",
+        url: documents_path(current_format.admin_slug)
```
ChrisBAshton: Shouldn't these `url` values be diffeent?

app/views/shared/_form_fields.html.erb
```diff
@@ -7,9 +7,21 @@
...
+    text: "Summary (required)",
+    heading_size: "m",
+  },
+  name: "#{format_name}[summary]",
+  class: "js-length-counter",
+  data: {
+    count_message_threshold:,
+    count_message_selector:,
+  },
+} do %>
```
ChrisBAshton: Is the character count JS functionality natively supported? I don't see an option for it on https://components.publishing.service.gov.uk/component-guide/textarea

Would be good to make clear in this commit ("Replace Summary text area") whether we're intending to retain the JS counter and what 'legacy' code from Specialist Publisher we're having to retain to support that.

spec/features/creating_a_new_document_design_system_spec.rb
```diff
@@ -75,14 +75,14 @@
...
 
-      schema.facets.each do |facet|
-        properties = facet["specialist_publisher_properties"] || {}
-        validations = properties["validations"] || {}
-
-        if validations["required"]
-          expect(page).to have_css("div.field_with_errors span.elements-error-message", text: /#{facet['key'].humanize} can't be blank|#{facet['name']} can't be blank/)
-        end
-      end
+      # schema.facets.each do |facet|
```
ChrisBAshton, line 79: Was this commenting out intentional? If so, good to make that clear in the commit message ("Replace the single select component" commit)

app/models/concerns/date_validation.rb
ChrisBAshton, line 1: Is this code - or any of the code in `app/views/shared/_datetime_fields.html.erb` - based on existing code elsewhere in alphagov? I suspect it is, as it's a lot of code to be added in one commit, and it looks pretty mature. If it is based on code elsewhere, I'd like to that existing code, in the "Replace date input" commit.

app/assets/stylesheets/application.scss
```diff
@@ -28,6 +28,5 @@ $govuk-page-width: 1140px;
 @import 'govuk_publishing_components/components/summary-card';
 @import 'govuk_publishing_components/components/textarea';
 
-@import "./components/add-another";
```
ChrisBAshton, line 30: I'd like to see a reason/"why" being added to the "Replace multiple select with Add Another" commit message, as it lacks detail at the moment. I'm also confused as to why this line of code is being removed 🤔

app/views/shared/_govspeak_help.html.erb
ChrisBAshton, line 1: I'd rename this commit from "Replace govspeak help component" to something like "Copy govspeak_help component into Specialist Publisher" so it's clearer we're not just swapping one component call for another pre-existing component.

ChrisBAshton (commented):
Nice one for persevering with this - it's a big job! ⭐

The PR would benefit from some before-and-after screenshots 🙏  Otherwise come comments below (mostly about maintaining a good audit trail in our commit messages)

app/views/documents/new.html.erb
minhngocd: I can pick this up as Matt's out - I think the "New Document" link shouldn't have a URL if we want to mimic the old functionality?

app/views/shared/_form_fields.html.erb
minhngocd: How do we make this call? Should I rope in Patrick?

spec/features/creating_a_new_document_design_system_spec.rb
minhngocd: Yes it is intentional as we've descoped validation as part of a separate card. I'll amend the commit message

app/models/concerns/date_validation.rb
minhngocd: It is, it's been ported over from Whitehall - i'll amend the commit message

app/assets/stylesheets/application.scss
minhngocd: Matt found that these style overrides were causing the add another components to look out of place on the form with some different margins and padding. Removing them makes them more consistent with how it is currently looking on Whitehall (used in fatality notices). We think these were put in when they were implemented for the admin change request pages. I'll also include some screenshots of those pages

app/views/shared/_form_fields.html.erb
ChrisBAshton: I'd probably raise it with Andrew at standup tomorrow. If it's not a supported feature of govuk_publishing_components then it feels like a non-standard feature to me and we're best off dropping it - but it's a PM call IMO.

lauraghiorghisor-tw:
It's already reviewed now, so not much in it, but I think it would have been good to have a breakdown of the methodology/approach in the PR description/initial commit. I did get a mental image of it all after looking through the commits, but it's a lot. The why of certain changes can maybe be captured on some of the larger commits (think Chris already said that elsewhere).
 
Perhaps it'll be useful to have the approach laid out, say, if someone else picks up the edit work, save them having to retrace all the thinking.

app/views/shared/_specialist_document_form.html.erb
```diff
@@ -8,9 +8,14 @@
     <%= render layout: "shared/date_fields", locals: { f: f, field: field, format: format_name, label: field_config["name"] } do %>
     <% end %>
   <% elsif !field_config["allowed_values"] %>
-    <%= render layout: "shared/form_group", locals: { f: f, field: field, label: field_config["name"] } do %>
-      <%= f.text_area field, class: 'form-control' %>
-    <% end %>
+    <%= render "govuk_publishing_components/components/textarea", {
```
lauraghiorghisor-tw, line 37: For these generic ones, can we set the row limit to lower (3 or so)? I think they were a tad smaller on the original views.

app/assets/stylesheets/application.scss
lauraghiorghisor-tw: I don't think there is a way around it, but I do find that the Add another component makes the form very busy and difficult to follow. I don't know if it's customary that this component lives at the bottom of the page, or if we can think up some way to delimit one select from another. If we have a document with three rounds of multiselect, the form will be an utter mess (and we probably do...).

Something to maybe consider when we finish migrating. Some UX/wrap-up card or a new card to tackle this.
I'd actually be in favour of using the other type of multi-select (that we have on the orgs in whitehall.)

minhngocd: There's another 2 cards in relation to "Add another" component in Ready to pick up - might be a good place for us to get together and discuss the component specifically. I would descope it from this PR - will pop in some time in the calendar for all of us (Chris, Matt, myself + Laura if you're interested) to figure out what we want to do with it

app/views/shared/_specialist_document_form.html.erb
minhngocd: I'll rope in Matt tomorrow

app/views/documents/new.html.erb
```diff
@@ -16,10 +16,6 @@
   } %>
 <% end %>
 
-<% content_for :document_ready do -%>
-  GOVUK.formChangeProtection.init($('.new_document'), 'You have unsaved changes that will be lost if you leave this page.');
-<% end -%>
```
minhngocd, line 10: We probably also need to have a chat around the removal of the javascript popup when navigating away from unsaved changes too. It's causing issues with the chrome browser tests, and Whitehall does not have this functionality.

app/models/concerns/date_validation.rb
minhngocd: Looks like we're not using the validation part of it for now... I've amended that information to the commit message as well

app/views/shared/_form_fields.html.erb
minhngocd: Need to remove the old JS and replace it with the new counter component

app/views/documents/new.html.erb
minhngocd: Decision to remove and see whether or not people ask for it back. Make it consistent with Whitehall, doesn't cause issues with feature specs

app/assets/stylesheets/application.scss
matthillco: Just chiming in here: As Minno said, we removed the styles as they looked wrong compared to Whitehall. They were also hooked into a `.js` class and were overriding a gem component, both of which are discouraged practices. Gem components are built aorund a concept of 'component isolation' which precludes this sort of ad-hoc CSS override. Any styling of a publishing gem component should be done in the gem, not within the apps.

Having said all that, I agree that the Add Another component has some strange UX and I'm not personally a fan. It would be good if we could get some UR on how users are interacting with it and where any pain points are. On a personal level I'd prefer to replace it with the Multiple variant of `select-with-search`.

lauraghiorghisor-tw (approved):
LGTM 

Sanity checked commits and changes ✅

PR #2715, opened 2024-08-05
Temporarily disable test

lauraghiorghisor-tw:
The CI is running on the latest version of chrome, which we are currently unable to pull in locally, to replicate the issue. All tests passing locally.
There are some changes we want to release today esp. around the OVA finder. Will be looking to fix the test presently.

[Related Trello](https://trello.com/c/HkndhHlm/2769-ova-changes)

PR #3091, opened 2025-04-10
Re-enable tests

ChrisBAshton:
These tests were temporarily disabled in https://github.com/alphagov/specialist-publisher/pull/2715.

Taking inspiration from https://github.com/alphagov/whitehall/pull/10074, we're adapting our CI so that we can re-enable the tests.

Similar problems affected other apps around the time the most recent ubuntu-latest image for GitHub Actions was released: https://github.com/actions/runner-images/pull/11761

In those cases, the fix was to re-install a different version of Chrome. Summary of changes:

- add `--no-sandbox` flag to Chromedriver startup (GovukTest will do this for us if we set the `GOVUK_TEST_CHROME_NO_SANDBOX` environment variable, but the theory is that making the change here keeps all these temporary fixes in one place)
- remove the version of Chrome that the actions/runner-images image bundles for us, so that Selenium doesn't try to use it
- install an old version of Chrome and Chromedriver

Trello: https://trello.com/c/7FTvIbPD/3589-re-enable-tests-disabled-due-to-unable-to-find-modal-dialog-error

---

### DIFF
--- a/.github/workflows/copy-pr-template-to-dependabot-prs.yml
+++ b/.github/workflows/copy-pr-template-to-dependabot-prs.yml
@@ -14,10 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     if: github.actor == 'dependabot[bot]'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Post PR template as a comment
-        uses: actions/github-script@v6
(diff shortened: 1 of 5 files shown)


ChrisBAshton:
Superseded by https://github.com/alphagov/specialist-publisher/pull/3108

PR #3095, opened 2025-04-14
New design system preview button

minhngocd:
Port over the GovSpeak Editor from Whitehall so we have a styled design system text area with a functional Preview button.

Removed some functionality that are currently not used in Specialist Publisher.
e.g. Rendering of Images and Attachment when rendering GovSpeak Editor preview. This is because Specialist Publisher does not support image uploading currently, and attachments are only available on the edit screens, which are not yet transitioning to the new design system. 

Removing the unused code instead of just leaving them there because the way it is implemented is different in Whitehall. The way we gather the image IDs and Attachment IDs are different and not currently implemented in Specialist Publisher. We can probably make that decision when we get to the point of transitioning Edit screens and Attachments and hence requiring to preview them. Another card has been created for Preview Button for Attachments.

https://trello.com/c/BvKeF7Vw/3641-design-system-add-new-document-preview-button

### DIFF
--- a/app/controllers/govspeak_controller.rb
+++ b/app/controllers/govspeak_controller.rb
@@ -2,7 +2,7 @@ class GovspeakController < ApplicationController
   def preview
     skip_authorization
 
-    json_attachments = JSON.parse(params["attachments"])
+    json_attachments = params["attachments"] ? JSON.parse(params["attachments"]) : []
 
     attachments = json_attachments.map do |attachment|
       Attachment.new(attachment)
--- a/app/views/shared/_form_fields.html.erb
+++ b/app/views/shared/_form_fields.html.erb
@@ -20,18 +20,22 @@
   maxlength: 280,
 } %>
 
-<%= render "govuk_publishing_components/components/textarea", {
+<%= render "shared/govspeak_editor", {
   label: {
     text: "Body (required)",
     heading_size: "m",
   },
+  id: "#{format_name}_body",
   name: "#{format_name}[body]",
   rows: 20,
+  # value: historical_account.body,
+  # error_items: errors_for(historical_account.errors, :body),
 } %>
 
 <% locale = @document.locale || "en" %>
 <%= render "govuk_publishing_components/components/select", {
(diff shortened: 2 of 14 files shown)


app/views/shared/_form_fields.html.erb
```diff
@@ -24,18 +24,22 @@
...
-<%= render "govuk_publishing_components/components/textarea", {
+<%= render "shared/govspeak_editor", {
   label: {
     text: "Body (required)",
     heading_size: "m",
   },
+  id: "#{format_name}_body",
   name: "#{format_name}[body]",
   rows: 20,
+  # value: historical_account.body,
```
lauraghiorghisor-tw, line 31: What is this value field meant to accomplish? Do you plan to bring it back?

I take it the errors bit is commented out because the validations are in a different card.

app/controllers/govspeak_controller.rb
```diff
@@ -2,7 +2,7 @@ class GovspeakController < ApplicationController
   def preview
     skip_authorization
 
-    json_attachments = JSON.parse(params["attachments"])
+    json_attachments = params["attachments"] ? JSON.parse(params["attachments"]) : []
```
lauraghiorghisor-tw, line 5: Very interesting 🤔

lauraghiorghisor-tw (approved):
LGTM

app/views/shared/_form_fields.html.erb
minhngocd: ah - it's a copy paste of the Whitehall rendering which contains values and errors - value is the retained value the user previously entered upon edit or failed save. It's out of scope but dealt with in the next ticket / PR https://github.com/alphagov/specialist-publisher/pull/3102 . Leaving it commented out as it does need to be filled out, but not just right now. 

Happy to remove or add a comment to the commit, but didn't see a critical need as it is short lived.

lauraghiorghisor-tw: Remove comment altogether if not explained in commit ;)

PR #3099, opened 2025-04-17
Changes to UKHSA

lauraghiorghisor-tw:
* Changed title to “Find out who is approved to access UKHSA data for research and other secondary uses”
* Org name - changed to a filterable set (with a placeholder value for now) - single select
* Alphabetical order for org type (but left other as last option)
* Changed classification values (add new, rename)
* Removed "protected" from summary
 

[Trello](https://trello.com/c/zubgVL1V/3569-specialist-finder-for-ukhsa-data-releases)

### DIFF
--- a/lib/documents/schemas/ukhsa_data_access_approvals.json
+++ b/lib/documents/schemas/ukhsa_data_access_approvals.json
@@ -64,13 +64,20 @@
       "type": "text"
     },
     {
+      "allowed_values": [
+        {
+          "label": "Placeholder organisation",
+          "value": "placeholder-organisation"
+        }
+      ],
       "display_as_result_metadata": true,
-      "filterable": false,
-      "key": "ukhsa_applicant_organisation_name",
+      "filterable": true,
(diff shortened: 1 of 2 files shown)


lib/documents/schemas/ukhsa_data_access_approvals.json
```diff
@@ -64,13 +64,20 @@
       "type": "text"
     },
     {
+      "allowed_values": [
```
minhngocd, line 67: ??? so this is a required dropdown with only one value??? i'm assuming there will be more populated later? is it worth making it non-mandatory for now? not sure what the value of changing this now is.

minhngocd (approved):

lib/documents/schemas/ukhsa_data_access_approvals.json
lauraghiorghisor-tw: I wanted to get the change over the line (they can use the admin area later or send the list of orgs via zendesk). 
The finder is in draft for now.

PR #3102, opened 2025-04-22
New design system retain data

minhngocd:
The design system form did not persist data when failing to save. We needed to read the value from the document model. Also did some refactorings to make the facet field in the form a view component to better navigate. 

https://trello.com/c/5JnIybdU/3637-design-system-add-new-document-persist-form-content-on-failed-save-draft


lauraghiorghisor-tw (approved):
LGTM

- I think the `input` bit in the names is a bit distracting, especially since input has other connotations in this space. I think `SpecialistDocumentFacetComponent` & co sound just fine. Could even drop the specialist document, if you ask me 🙈
- The textarea changes (value) can probably be moved from the last commit, as it seems to be more about multi and single select business, but not fussed.

minhngocd:
> LGTM
> 
> * I think the `input` bit in the names is a bit distracting, especially since input has other connotations in this space. I think `SpecialistDocumentFacetComponent` & co sound just fine. Could even drop the specialist document, if you ask me 🙈
> * The textarea changes (value) can probably be moved from the last commit, as it seems to be more about multi and single select business, but not fussed.

@lauraghiorghisor-tw the `input` in the name is to specify it as an input field view component. As opposed to, say when we want to display components or from any model naming. although we do have a directory namespace that i could play with so i'll see if i can come up with any alternative naming

2nd point - the commit is about all facet types retaining previously entered value (date, textarea, single and multi select). And the added test checks for that for every document. Which is why i think it's more complete to have them together as one commit.

PR #3117, opened 2025-05-01
Added a dartsass watch rake task which uses polling to detect SCSS changes

ryanb-gds:
The default dartsass watch command does not seem to work from inside GOV.UK docker, at least on MacOS machines. Using the polling option to poll the filesystem for changes works. It's a bit slow but beats using `rails precompile:assets` every time we make a change


patrickpatrickpatrick (approved):
Tested this locally and worked 👍

PR #3119, opened 2025-05-02
Refactor to finder resource

ryanb-gds:
Refactoring to use a resource route collection for all endpoints relating to finders should make leveraging Rails conventions easier, hopefully increasing our future productivity when working on Specialist Publisher.

This work is being done as part of https://trello.com/c/lRu3iVj3


app/views/finders/_facet_fields.html.erb
ChrisBAshton, line 1: Is there any way of restructuring the commit so that we effectively have a 'rename' of the existing https://github.com/alphagov/specialist-publisher/blob/cb937509d94a4244ca63c0e18fe4f7dd5bf3f054/app/views/admin/_facet_fields.html.erb file, rather than duplicating the file contents? It would be nice to know for certain that the file contents are identical (and where the code came from) rather than making the assumption as a reader.
E.g. by specifying template override in the controller until the final commit where we delete the admin controller - can then add one new 'final' commit that removes the override and renames the templates.

If not, perhaps clarify in the commit message that you'd copied the code over from that file, with a link to the sha for posterity.

ChrisBAshton (approved):
Looks good 👍 one non-blocking suggestion below, which would make it easier to review commit-by-commit

app/views/finders/_facet_fields.html.erb
ryanb-gds: Hadn't thought of that. I'll have a look

ryanb-gds: I have done this, but slightly stuffed up and accidentally merged the facets and metadata commits accidentally at some point. Not the end of the world, but rebasing is a tricky business 😅

PR #3131, opened 2025-05-09
Skip authorisation for finders index action

ryanb-gds:
There's no need to authorize the finders index action, as all users should be able to view the list of finders scoped to their permission level. The index method on the finder schmea policy was failing for non GDS editor users because it was attempting to check the schema organisations for the finder schema class, when no such method exists.

To avoid reoccurrence of such issues in future I have moved any behaviour in the application policy which was in fact specific to documents into the document policy class.

Trello: https://trello.com/c/lRu3iVj3

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -1,7 +1,7 @@
 class FindersController < ApplicationController
   layout "design_system"
   def index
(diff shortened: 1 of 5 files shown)


patrickpatrickpatrick (approved):
Looks good to me!

PR #3133, opened 2025-05-12
Provide design system implementation of documents index page

ryanb-gds:
Trello: https://trello.com/c/7cMAnBlD

## Screenshot

Screenshot 2025-05-12 at 13 32 38 (2)

One thing worth mentioning is that the GOV.UK Publishing Components library does not yet implement the full GOV.UK frontend pagination component. It would be preferable to implement the full pagination component if it becomes available in the future.

### DIFF
--- a/app/views/documents/index.html.erb
+++ b/app/views/documents/index.html.erb
@@ -1,57 +1,148 @@
-<%= content_for :page_title, current_format.title.pluralize %>
+<% content_for :page_title, "#{current_format.title} finder" %>
 
 <% content_for :breadcrumbs do %>
-  <li class='active'><%= current_format.title.pluralize %></li>
+  <%= render "govuk_publishing_components/components/breadcrumbs", {
+    collapse_on_mobile: true,
+    breadcrumbs: [
+      {
+        title: "All finders",
+        url: finders_path
+      },
+      {
+        title: "#{current_format.title} finder",
+        url: documents_path(current_format.admin_slug)
+      },
+    ]
+  } %>
 <% end %>
 
-<h1 class="page-header"><%= current_format.title.pluralize %></h1>
+<% content_for :title do %>
+  <%= "#{current_format.title} finder" %>
+<% end %>
 
-<div class="row">
-  <div class="col-md-12 text-right">
-    <p>
-      <% if current_format.target_stack == "draft" %>
-        <%= link_to "Preview draft (opens in new tab)", current_format.draft_url, target: "_blank" %>
-      <% elsif current_format.target_stack == "live" %>
-        <%= link_to "View on website (opens in new tab)", current_format.live_url, target: "_blank" %>
-      <% end %>
-    </p>
-  </div>
-</div>
+<%= render "govuk_publishing_components/components/inset_text", {} do %>
+  <% if current_format.target_stack == "draft" %>
+    <%= link_to "Preview draft (opens in new tab)", current_format.draft_url, class: "govuk-link", target: "_blank", rel: "noopener" %>
+  <% elsif current_format.target_stack == "live" %>
+    <%= link_to "View on website (opens in new tab)", current_format.live_url, class: "govuk-link", target: "_blank", rel: "noopener" %>
+  <% end %>
+<% end %>
 
-<div class="row">
-  <div class="sidebar col-md-3">
-    <%= link_to "Add another #{current_format.title}", new_document_path(current_format.admin_slug), class: 'action-link' %>
-    <hr>
-    <%= link_to "Finder setup", finder_path(current_format.admin_slug), class: 'action-link' %>
+<%= render "govuk_publishing_components/components/secondary_navigation", {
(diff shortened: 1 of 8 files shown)


patrickpatrickpatrick:
This isn't one specifically for this PR but I'm wondering if we should at some point move the facet containing component ('Filter by') into govuk_publishing_components. It's used in this application and in whitehall. I assume it will also be used in publisher also.

app/views/documents/index.html.erb
```diff
@@ -1,57 +1,145 @@
...
+          {
+            text: "Title"
+          },
+          {
+            text: "Updated",
+          },
+          {
+            text: "Status",
+          },
+          {},
```
patrickpatrickpatrick: It's good practice to have some hidden text here that describes what's in this column, [in whitehall we've used "View"](https://github.com/alphagov/whitehall/blob/8dc01eac6f46e9354d1debe305fdd82a9bb01b3c/app/views/admin/editions/_search_results.html.erb#L25).

app/views/documents/index.html.erb
```diff
@@ -1,57 +1,145 @@
...
+            },
+            {
+              text: "Updated #{time_ago_in_words(document.last_edited_at)} ago",
+            },
+            {
+              text: raw("<strong class='#{design_system_classes_for_frontend(document)}'>#{state_for_frontend(document).upcase}</strong>"),
+            },
+            {
+              text: link_to(
+                "View",
```
patrickpatrickpatrick: It's good to have [hidden text of the item that the user is viewing here for screen readers](https://github.com/alphagov/whitehall/blob/8dc01eac6f46e9354d1debe305fdd82a9bb01b3c/app/views/admin/featurable_editions/_search_results.html.erb#L36), otherwise it will just read 'View' which might be confusing if the user is navigating the page using links.

app/views/documents/index.html.erb
```diff
@@ -1,57 +1,145 @@
...
+        <%= render "govuk_publishing_components/components/select", {
+          id: "filter-organisations",
+          label: "Organisations",
+          options: organisation_options_for_design_system(selected_organisation_or_current(@organisation))
+        } %>
       <% end %>
 
-      <%= label_tag "query", "Search" %>
-      <%= text_field_tag("query", @query, class: "form-control add-bottom-margin") %>
+      <%= render "govuk_publishing_components/components/input", {
```
patrickpatrickpatrick, line 62: Should we make this a `type="search"`? I believe this is the functionality we have in whitehall.

app/views/documents/index.html.erb
```diff
@@ -1,57 +1,145 @@
...
+                "View",
+                document_path(current_format.admin_slug, content_id_and_locale: "#{document.content_id}:#{document.locale}"),
+                class: "govuk-link",
+                ),
+            },
+          ]
+        end
+      } %>
+
+      <%= render "govuk_publishing_components/components/previous_and_next_navigation", {
```
patrickpatrickpatrick, line 130: I think my instinct would be to move the logic determining the object being passed into the `previous_and_next_navigation` component to a helper but I appreciate that we are intending to add the pagination functionality to the component in the gem, so this would be superseded by that.

app/views/documents/index.html.erb
ryanb-gds: Good shout

app/views/documents/index.html.erb
ryanb-gds: My preference is to use helpers (or partials) only when there's a strong case for sharing code with other templates, which I don't think there is here yet. Otherwise it's just indirection for indirection's sake, imo

patrickpatrickpatrick: That makes sense, happy to leave it like that :)

app/views/documents/index.html.erb
```diff
@@ -1,57 +1,148 @@
...
+        rows: @paged_documents.map do |document|
+          [
+            {
+              text: document.title,
+            },
+            {
+              text: "Updated #{time_ago_in_words(document.last_edited_at)} ago",
+            },
+            {
+              text: raw("<strong class='#{design_system_classes_for_frontend(document)}'>#{state_for_frontend(document).upcase}</strong>"),
```
patrickpatrickpatrick: Sorry, just noticed this! What's the purpose of the `strong` here? Could this be a `tag.span` instead? Also if we're being consistent with the Design System, they don't use all upper case for text content in tags.

ryanb-gds: Have to confess I copied that from somewhere else, it absolutely could be a span. It's upcased in the design but I think if the design system doesn't use all caps then neither should we, will fix

patrickpatrickpatrick:
This looks better now. I think there's probably more rationalisation we can do cross application and in govuk_publishing_components but I think that come later.

patrickpatrickpatrick (approved):

PR #3135, opened 2025-05-12
New Document design-system views for non-conforming pages

matthillco:
## What

Replace legacy views and components with design system views and components for the New Document page template. This PR applies to non-conforming pages, which are specialist document types that don't use the common `specialist_document_form`.

## Why

We are updating Specialist Publisher to use the Design System.

## Key implementation details
- Added a new `InputComponent` to deal with input fields
- Added a new `OrganisationSingleSelectComponent` - to deal with the preselected organisation select that exists on the new document page
- Migrated 3 document types so far
- Fixed some rendering issue observed along the way (mainly around validation, as needed to pass tests)

## Approach for each non-conforming document type

- Port over the view (all components except multiple select)
- Fix schema ordering or model validation order to match pre-existing order
- Add feature test/model test - with a view to refactor these (unify them) in the future when we can form a better understanding 
- Decision: do not change any content - particularly labels for business finance support scheme were kept the same
- Decision: migrate single select to single select with blank (despite it having been preselected previously), specifically for  Statutory instruments (sifting status field)

## Descoped
- Research for development
- Licence transactions

The PR was getting too large, opted for early review. Plus, these two seem particularly convoluted.

## Testing

- Feature tests
- Manual testing of all doc types

Screenshots from testing (left: current changes/DS; right: production code/legacy)

- Statutory instruments - not the single select change "Sifting status" (blank is now selected) and the preselected "Primary Organisation"
> STAT INST preselected single select
- Statutory instruments - with withdrawn sifting status -> changed rendering order of withdrawn date (looks more usable)
> STAT INST WITHDRAWN
- Business finance support scheme -> note the "Will continue on" and "Continuation link" input components
> Business finance support scheme
- Flood and coastal erosion -> not the "Primary Organisation" preselected single select
> flood and coastal erosion

[Trello](https://trello.com/c/Bl6vFuT4/3656-design-system-add-new-document-non-conforming-pages)


app/components/facet_input_component/date_component.html.erb
```diff
@@ -5,6 +5,7 @@
   field_name: "#{@document_type}[#{@facet_key}]",
   error_items: @error_items,
   date_only: true,
+  date_hint: "For example, 2025 05 15",
```
minhngocd: The UI is DD MM YYYY, the example hint text is YYYY MM DD. Might be worth changing to match the input UI

app/helpers/organisations_helper.rb
```diff
@@ -1,18 +1,28 @@
...
+                               value: "all",
+                               selected: false,
+                             },
+                           ]
+                         when :blank
+                           [{}]
+                         when :none
+                           nil
+                         end
+    sorted_organisations = all_organisations.inject([]) { |options, organisation|
```
minhngocd: A couple of questions here: 
- what's the difference between :blank and :none? why do we need these scenarios?
- Can't you just pass a pre-populated value if you want the select to be preselected? otherwise they all default to the blank option
- If we do the above, then we can get rid of the switch case. Can handle it as if the select is prepopulated with an existing value (selected boolean checks for matching value)

app/views/metadata_fields/_flood_and_coastal_erosion_risk_management_research_reports.html.erb
```diff
@@ -1,30 +1,7 @@
 <%= render layout: "shared/specialist_document_form", locals: { f: f } do %>
 <% end %>
 
-<% # TODO: find a way of making organisation management schema-driven %>
-<%= render "shared/form_group", f: f, field: :primary_publishing_organisation, label: "Publishing organisation" do %>
```
minhngocd, line 5: Old label was Publishing Organisation, new label is Primary organisation. Should they be the same?

app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -1,55 +1,15 @@
-<%= render "shared/legacy/date_fields_legacy", f: f, field: :sift_end_date, format: :statutory_instrument %>
-<%= render "shared/legacy/date_fields_legacy", f: f, field: :laid_date, format: :statutory_instrument %>
-<%= render "shared/form_group", f: f, field: :sifting_status do %>
-  <%= f.select :sifting_status, select_options_for_facet(f.object.allowed_values(:sifting_status)), {}, { class: "form-control" } %>
+<%= render layout: "shared/specialist_document_form", locals: { f: f } do %>
 <% end %>
-<% # TODO: this bit of logic prevents us moving to the shared view %>
+
+<% format_name = f.object.class.document_type.to_sym %>
 <% withdrawn_date_display = f.object.withdrawn_sifting_status? ? "block" : "none" %>
```
minhngocd, line 9: I disagree with the re-ordering of the fields... I think Withdrawn Date needs to sit directly below Sifting Status as it was like in the old form. The reason being I believe it looks like it's connected, as the Sifting status affects whether or not Withdrawn date even appears: 

image

app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -1,55 +1,15 @@
-<%= render "shared/legacy/date_fields_legacy", f: f, field: :sift_end_date, format: :statutory_instrument %>
-<%= render "shared/legacy/date_fields_legacy", f: f, field: :laid_date, format: :statutory_instrument %>
```
minhngocd, line 2: This label is changing from Laid Date to Laid On - is that also intentional?

app/components/facet_input_component/date_component.html.erb
lauraghiorghisor-tw: Oh yeah lol. Changed ✅

app/views/metadata_fields/_flood_and_coastal_erosion_risk_management_research_reports.html.erb
lauraghiorghisor-tw: Yep 🦅

app/views/metadata_fields/_statutory_instruments.html.erb
lauraghiorghisor-tw: It was somewhat but will change it to what it was prior since I am making the withdrawn date change anyway.

app/views/metadata_fields/_statutory_instruments.html.erb
lauraghiorghisor-tw: Changed ✅
Screenshot 2025-05-28 at 15 46 30

app/helpers/organisations_helper.rb
lauraghiorghisor-tw: Currently we use the `organisation_select_options` in the following situations:
- with: :all - on `index.html` - it does require the "all organisations" option
- with: :none - used within the new document form (specifically in the single select component) - there is always a selected option, because the primary organisation will default to the logged in user's; so "none" means no additional blank option. We can call it something different, maybe`with: :preselected`?
- with: :blank - not sure we will need it in this way - currently used for rendering the multiselect for "Other associated organisations" on some of the non-conforming pages - which is not working at all at the minute 😬.

app/helpers/organisations_helper.rb
```diff
@@ -1,18 +1,26 @@
...
-        selected: false,
-      },
-    ] + all_organisations.sort_by { |org| org.title.downcase.strip }.map do |organisation|
-      {
-        text: organisation.title,
-        value: organisation.content_id,
-        selected: organisation.content_id == selected_organisation_content_id,
-      }
-    end
+  def organisation_select_options_with_all(selected_organisation: nil)
```
minhngocd, line 2: I don't know how DRY you would want to do it, but `organisation_select_options_with_all` could be something like 

```
organisation_select_options.prepend({ text: "All organisations", value: "all", selected: "all" ==  selected_value})
```

on that note - i also think the parameter passed in `selected_organisation` here could very well be "all" to retain the user-preselected-value

app/helpers/organisations_helper.rb
```diff
@@ -1,18 +1,26 @@
...
+    all_organisations
+      .sort_by { |org| org.title.downcase.strip }
+      .inject([{ text: "All organisations", value: "all", selected: false }]) do |options, organisation|
+        options << {
+          text: organisation.title,
+          value: organisation.content_id,
+          selected: organisation.content_id == selected_organisation,
```
minhngocd: this line here will mean "all" will never be preselected...?

app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -1,55 +1,19 @@
...
-<%= render "shared/form_group", f: f, field: :sifting_status do %>
-  <%= f.select :sifting_status, select_options_for_facet(f.object.allowed_values(:sifting_status)), {}, { class: "form-control" } %>
-<% end %>
-<% # TODO: this bit of logic prevents us moving to the shared view %>
```
minhngocd, line 6: Might be worth leaving this comment back in until we move over?

app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -1,55 +1,19 @@
...
-        multiple: true,
-        data: {
-          placeholder: "Select organisations"
-        }
-      }
-  %>
-<% end %>
+<%= render FacetInputComponent.new(@document, finder_schema.get_facet(:subject)) %>
+
+<%= render FacetInputComponent::OrganisationSingleSelectComponent.new(@document, format_name, :primary_publishing_organisation, "Primary organisation", selected_organisation_or_current(@document.primary_publishing_organisation)) %>
```
minhngocd: This one should also be "Publishing organisation"

minhngocd (approved):
Some suggestions and a label change.

app/helpers/organisations_helper.rb
lauraghiorghisor-tw: Oh yes, this explains why the filtering on the index page is broken - no change here from Ryan's original code though - can fix in the new "bug" card.

app/views/metadata_fields/_statutory_instruments.html.erb
lauraghiorghisor-tw: I think everything in that view prevents us from doing it right now, the js, the order of fields in the schema, the custom labels, you name it 😅 I can add to the commit message.

lauraghiorghisor-tw: Readded + added comment in commit

app/helpers/organisations_helper.rb
lauraghiorghisor-tw: Dried 🪦

Kept the "false" for now, will think it through in the bug fix.

PR #3147, opened 2025-05-16
Changes for review of service - Countryside stewardship grants

patrickpatrickpatrick:
## What

🚨 These changes should be made live on the 27th of June. ⏰

Update `countryside_stewardship_grants.json` to:
- rename service and path
- remove/update facets.

## Why 

This request is based on:
- a review of the service
- feedback received from users
- updated information

The service review changes (i.e. renaming the finder) are set up in this PR - https://github.com/alphagov/specialist-publisher/pull/3151.

## Screenshot from testing
Screenshot 2025-05-27 at 09 35 31
Screenshot 2025-05-27 at 09 35 39
Screenshot 2025-05-27 at 09 40 19
Screenshot 2025-05-27 at 09 36 12
Screenshot 2025-05-20 at 16 14 19

[Trello](https://trello.com/c/UXXwOxKQ/3728-specialist-finder-edit-request-countryside-stewardship-grants)
[Trello](https://trello.com/c/0g9CpeLQ/3727-specialist-finder-edit-request-countryside-stewardship-grants)

PR #3151, opened 2025-05-16
Decouple document type from document title

lauraghiorghisor-tw:
## What

These changes address these pre-existing behaviours in Specialist Publisher:
- The document title was used for more than just rendering on forms. The view files themselves were named and loaded based on the `document_type_slug`, passed as a parameter with the value of `admin_slug` (`title.parameterize.pluralize`), which is title based. That meant that we had a bunch of views that did not match our usual naming conventions (where the json and the view files are the pluralized model name). This also meant that if we needed to change the title, we had to rename files. 
- Specialist Publisher used to deduce the `document_type` value that we send to Publishing API from the corresponding model class name (`to_s.underscore`). Nonetheless, there are historical cases where this is not the case, which we are currently not including in generic testing (and not testing at all). It's easy to assume that the schema set `filter.format` is the "official" document type (the format does typically match our model names, but it is not used in the presenter; it is mainly used in Search API). Following the scenario from the previous point, if we did end up renaming views and also wanted to change the model names to match, we'd be unable to, because that would change the document type value. There is no way to rename files without causing downstream issues.
- Filter format - because we associate it with the document type - is then used to render views (as part of the `name[]` on forms) - which breaks non-conforming finders. This is because the `DocumentsController` permits params based on `document_type` not `filter.format`. 
- ESI Fund is the outlier where the naming conventions have not been respected (has a different `filter.format` value than what the model name). It is also currently broken and cannot be included in general testing. The issues are a mixture of `document_type` and `filter.format` used in the form rendering (pre-design system migration), and then the use of `filter.format` in the design system which is not permitted at the controller level.

| Esi Fund | |
| --- | --- |
| Legacy 'esi_fund[]' on form | Legacy 'european_str...[]' on form  |
| Screenshot 2025-05-18 at 14 40 08 | Screenshot 2025-05-18 at 15 01 58 |
| Closing date missing on form submission (legacy) | Design system enabled ESI fund is broken on submission |
| Screenshot 2025-05-18 at 15 20 41 | Screenshot 2025-05-18 at 14 39 28 |
- Overall improving general testing coverage (so that we can catch things like ESI funds breaking).

Fixes (more details in commits):
- Instead of using the `document_title` for naming views, use the `document_type` instead. On occasion, we may want to change the title of a finder (and therefore documents). We want to be able to show this new name across the publisher app. Nonetheless, we should not be required to change the already configured models/views/schema file names.
- Use `document_type` instead of `filter.format` for SP internals (views, forms etc.). The idea is that if we rename things, they're still internally consistent within SP. The `filter.format` value is strictly used for downstream/Search API business. For new finders we can always have the document type and format match (added this as guidance).
- Add the possibility to specify a `document_type` that does not match the model name. Open for discussion if this should be a default field on all schemas or only custom (currently only in the ESI Fund schema). This requires maintaining a list of mappings.

## Why
- Incoming [changes to Countryside stewardship finder](https://github.com/alphagov/specialist-publisher/pull/3147) - title changes - unveiled that we cannot just do it. I tried renaming the view files to match the new title, which worked, but that further broke whatever loose naming conventions we had in the repo. With the new implementation, we can make those changes and any future title changes/file renaming possible/easy.
- ESI fund was broken (and totally not working for Design system)
- Improve test coverage ("generic").


lauraghiorghisor-tw (commented):
For reviewer: 

As far as unblocking the renaming of Countryside Stewardship Grants finder goes, the first 3 commits achieve that. They also suffice for adding ESI fund to general testing and ensuring it's not broken anymore. 

The last 3 functional commits, excluding documentation:
- Use `downstream_document_type` across the repo
- Introduce a mapper for non-conforming document types
- Allow specifying `document_type` for downstream usage, in finder schema
are a little more opinionated and risky. Especially for the top/last one, I had little reliance on tests, had to basically search for occurrences of 'document_type` across the repo and manually test rake tasks. I am still not sure how to manually test the tagger. I think covering such gaps is out of scope for this, but have done my due diligence. 

Manual testing plan:
- legacy - selection of finders, including all excluded from general testing and ESI fund
- design system - some finders, including ESI fund (all exceptions from general testing are not yet migrated)
- rake tasks that make use of document_type (publish and republish all)

Issues: legacy `business_finance_support_scheme` I could not preview once published, though that seemed to be an error in staging as well. With design system enabled it's a 500.

app/views/metadata_fields/_ai_assurance_portfolio_techniques.html.erb
patrickpatrickpatrick, line 1: This makes sense to me! Maybe an irrelevant question (and not one to address in this PR), is there are a reason we didn't go [this route](https://stackoverflow.com/a/3559555) to check to see if a partial exists and then have a default one if it doesn't?

app/models/document.rb
```diff
@@ -6,7 +6,6 @@ class Document
   include ActionView::Helpers::TextHelper
   include DateHelper
   include PublishingHelper
-
```
patrickpatrickpatrick: Is whitespace change intentional or by a linter?

patrickpatrickpatrick:
After a discussion on Slack with Laura, from an outsider (not backend) perspective I thought that a migration in `publishing-api` could be a fix to the issue addressed by the `document_type` mapping (for the ESI Fund). However I appreciate that co-ordinating the migration (for an application which our team doesn't own ourselves, which could potentially be risky or impossible) might take some time. As a result a potential plan of action I can see is:

- split the PR into two, one which is the partial changes (which I think are sensible and fix the support issue), one which is the last 3-5 commits for the `document_type` mapping (for the ESI Fund)
- create a card for the support backlog to contact `publishing-api` team about possibility of migration to fix the `document_type` issue
- we merge both the PRs
- if the `publishing-api` team is open to migration, we can either revert the second PR with the mapping changes or keep the mapping functionality and remove ESI fund from the list of exceptions

app/views/metadata_fields/_ai_assurance_portfolio_techniques.html.erb
minhngocd: I like this - it means we'll only have custom views for truly custom views rather than a passthrough to the generic form

lauraghiorghisor-tw:
> After a discussion on Slack with Laura...

Also from our chats, argument for why keep the mapper, even if we did do a migration (yes, I am quoting myself).

There is value in still having the underlying changes (adding the `downstream_document_type` and maintaining a mapper) - a mapper with just ESI funds, or a blank mapper if we do migrate ESI funds - so that we have the possibility to rename  files in SP without accidentally causing issues downstream. I think that’s a risky coupling to have.

My thinking with introducing the `downstream_document_type` was “let’s make it obvious” - because it took me a long time to track this document type through the system to understand what it was doing.

app/models/finder_schema.rb
```diff
@@ -34,6 +34,7 @@ def self.document_models
   attribute :description
   attribute :document_noun
   attribute :document_title
+  attribute :downstream_document_type
```
minhngocd, line 37: Might I suggest `public_content_type`, sort of in the same vein of `public_updated_at_date`, to show the contrast being public facing, and internal system; rather than upstream / downstream?

minhngocd: I also wonder if having a doc within this repo for `Ubiquitous Language` might be helpful to clarify the differences between

- document type
- content type
- document model
- document noun 

etc.

spec/support/document_type_mapper.rb
```diff
@@ -0,0 +1,10 @@
+module DocumentTypeMapper
```
minhngocd, line 1: Why do we have a duplicate mapper just for specs? can the specs not import and use the real mapper? probably shouldn't have a mapper just for specs as they can diverge.

app/helpers/document_type_mapper.rb
```diff
@@ -0,0 +1,10 @@
+module DocumentTypeMapper
```
minhngocd, line 1: Might be better to hold a hash rather than switch case, being more performant. Additionally, if the document models themselves hold the public document types defined, then we probably should use its enum instead of having repeat strings across the codebase?

app/helpers/document_type_mapper.rb
```diff
@@ -7,4 +7,9 @@ def self.get_document_type(document_type)
       document_type
     end
   end
+
+  def self.all_document_types
+    Rails.application.eager_load!
```
minhngocd, line 12: surely this should have been done in an initializer somewhere? 
If not, maybe it could live more in the initializer configs - e.g. `preload_document_types` or something.

app/helpers/document_type_mapper.rb
```diff
@@ -7,4 +7,9 @@ def self.get_document_type(document_type)
...
+    Document.subclasses.tap { |a| a.delete(Attachment) }.map(&:downstream_document_type)
```
minhngocd, line 13: I'm not sure about the impact of `delete` - haven't really tested, but can `exclude` be used instead here? or reject? for excluding Attachment as a subclass

app/helpers/document_type_mapper.rb
minhngocd: I see - that line was an existing piece of code that lived `publisher` and `republisher`. I think those usages there since they lived in `app/lib` had to do eager load because it is running a rake task, and does not have the rails application loaded. In which case, would make sense to separate the preloading just for rake tasks. instead of having them wrapped within this helper class here, since if it is called within the rails app, it shouldn't need to call this line.

docs/naming-conventions.md
```diff
@@ -0,0 +1,38 @@
...
+
+The files typically required to create a finder follow these conventions. Exceptions to these rules require test customization in the codebase, and should be avoided.
+
+## Schema
+- Use the pluralised name of the document type, e.g. `cma_cases.json`.
+- Within the schema, the `filter.format` is typically set to the singularised name of the document type, e.g. `cma_case`, although it does not have to be. This is used in Search API. It should not be changed once set.
+
+## Model
+- The singular name of the document type, e.g. `cma_case.rb`.
+- The model name dictates the `document_type` that gets sent to Publishing API. It should not be changed once set. Historically divergent behaviour is captured in a `downstream_document_type` schema field.
```
minhngocd, line 11: might be worth making it explicit that this public_document_type/downstream_document_type field has an overriding behaviour that falls back to document type if not defined?

minhngocd:
I do like the separation of internal document type usage that is closely tied to code structure and public facing document type usage for API contracts - both could evolve in different ways for different requirements, and at the moment they're restricting each other from changing. Splitting out the usages feel like a reasonable way to go.

PR #3152, opened 2025-05-19
Fix some exceptions from generic testing

lauraghiorghisor-tw:
## Fix `marine_equipment`
Document type `marine_equipment_approved_recommendation` was excluded from generic testing because it was failing validations due to a custom text field that takes a year value.

Added a tiny bit of logic to deal with this case. This makes it more obvious that we're making use of such date variations (for future design system date component improvements), and it enables us to run the entire test suite for this document types, so we don't have gaps in coverage. 

## Note on `protected_food_drink_name`
`protected_food_drink_name` should have also been included in generic testing with a similar fix (it has a text field that takes a validated time value, similar to the above case). Nonetheless, for some reason it passes validation with no summary, thus breaking some tests. I tested in deployed environments too, it seems to be the case everywhere. Could not get to the bottom of it, so left it as an exception for now. Hope it gets fixed when migrating to DS.

Screenshot 2025-05-19 at 15 28 11

### DIFF
--- a/spec/features/creating_a_new_document_spec.rb
+++ b/spec/features/creating_a_new_document_spec.rb
@@ -6,7 +6,6 @@ EXCEPTIONS_TO_GENERAL_TESTING = %w[
   licence_transaction
   research_for_development_output
   statutory_instrument
-  marine_equipment_approved_recommendation
   protected_food_drink_name
 ].freeze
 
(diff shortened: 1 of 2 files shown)


lauraghiorghisor-tw:
The food and drink name bent my mind - could not for the life of me understand what's going on 😅 Help welcome.

GDSNewt (approved):
Having a 'Year' facet that is not a date smells a bit but it's not a dealbreaker for testing 😬

lauraghiorghisor-tw:
> Having a 'Year' facet that is not a date smells a bit but it's not a dealbreaker for testing 😬

Yeah, that's future work for us - to add support for these date varieties.

PR #3165, opened 2025-05-27
Decouple document title from document type (split)

lauraghiorghisor-tw:
## What

These changes address these pre-existing problematic behaviours in Specialist Publisher:
- The document title was used for more than just rendering the title on forms. The view files themselves were named and loaded based on the `document_type_slug`, passed as a parameter with the value of `admin_slug` (`title.parameterize.pluralize`), which is title based. That meant that we had a bunch of views that did not match our usual naming conventions (where the json and the view files are the pluralized model name). This also meant that if we needed to change the title, we had to rename view files. 
- `filter.format` is on occasion confused with `document_type` (which is derived from the model `to_s.underscore`) - and thus used to render views (as part of the name[] on forms) - which breaks non-conforming finders. This is because the `DocumentsController` permits params based on `document_type` not `filter.format`.
- ESI Fund is the outlier where the naming conventions have not been respected (has a different `filter.format` value than what the model name). It is currently broken (see screenshots) and cannot be included in general testing. The issues are a mixture of `document_type` and `filter.format` used in the form rendering (pre-design system migration), and then the use of `filter.format` in the design system which is not permitted at the controller level.

Fixes (more details in commits):
- Instead of using the `document_title` for naming views, use the `document_type` instead. On occasion, we may want to change the title of a finder (and therefore documents). We want to be able to show this new name across the publisher app. Nonetheless, we should not be required to change the already configured models/views/schema file names.
- Use the `document_title` and not the `document_type` for SP internals, such as rendering forms. 
- Include ESI funds in general testing.

## Why
- Incoming [changes to Countryside stewardship finder](https://github.com/alphagov/specialist-publisher/pull/3147) - title changes - unveiled that we cannot just do it. I tried renaming the view files to match the new title, which worked, but that further broke whatever loose naming conventions we had in the repo. With the new implementation, we can make those changes and any future title changes/file renaming possible/easy.
- ESI fund was broken (and totally not working for Design system)
- Improve test coverage ("generic").

| Esi Fund (previous) | |
| --- | --- |
| Legacy 'esi_fund[]' on form | Legacy 'european_str...[]' on form  |
| Screenshot 2025-05-18 at 14 40 08 | Screenshot 2025-05-18 at 15 01 58 |
| Closing date missing on form submission (legacy) | Design system enabled ESI fund is broken on submission |
| Screenshot 2025-05-18 at 15 20 41 | Screenshot 2025-05-18 at 14 39 28 |
| Esi Fund (after - from QA) | |
| ESI FUNDS DS | ESI FUNDS LEGACY |

## Testing plan

Legacy/non-design-system 
- Few standard types
- AI assurance 
- Licence transactions
- Product safety alerts
- Statutory instruments
- ESI funds - including date submission

Design System:
- as above

Unblocking Countryside Stewardship Grant change
- Tested [that branch ](https://github.com/alphagov/specialist-publisher/pull/3147)on top of this, ensuring it will work as intended.

---

Split out from https://github.com/alphagov/specialist-publisher/pull/3151.


lauraghiorghisor-tw:
Comments from [previous PR](https://github.com/alphagov/specialist-publisher/pull/3151/commits/ac97a2ca0ed8679d4cf0af103b43a6dd7e77e05d#r2099815371) @patrickpatrickpatrick @minhngocd 

Patrick
> This makes sense to me! Maybe an irrelevant question (and not one to address in this PR), is there are a reason we didn't go [this route](https://stackoverflow.com/a/3559555) to check to see if a partial exists and then have a default one if it doesn't?

Minno
> I like this - it means we'll only have custom views for truly custom views rather than a passthrough to the generic form

I don't think I want to go down that path, I don't see how it suits our case. Because our views **have to** exist, so it's not a case of falling back to some "default". But maybe I don't understand exactly how you'd use that. I imagined you'd use it inside the `new.html.erb`? 
```
<%= render partial: "metadata_fields/#{@document.document_type.pluralize}", locals: { f: f } %
```

docs/naming-conventions.md
```diff
@@ -0,0 +1,38 @@
...
+
+The files typically required to create a finder follow these conventions. Exceptions to these rules require test customization in the codebase, and should be avoided.
+
+## Schema
+- Use the pluralised name of the document type, e.g. `cma_cases.json`.
+- Within the schema, the `filter.format` is typically set to the singularised name of the document type, e.g. `cma_case`, although it does not have to be. This is used in Search API. It must not be changed once set.
+
+## Model
+- The singular name of the document type, e.g. `cma_case.rb`.
+- The model name dictates the `document_type` that gets sent to Publishing API. It must not be changed once set.
```
lauraghiorghisor-tw, line 11: Changed from the original (pre-split PR) to capture the fact that we cannot rename models. 

Removed all mentions of the `downstream_document_type`.

patrickpatrickpatrick:
> Comments from [previous PR](https://github.com/alphagov/specialist-publisher/pull/3151/commits/ac97a2ca0ed8679d4cf0af103b43a6dd7e77e05d#r2099815371) @patrickpatrickpatrick @minhngocd
> 
> Patrick
> 
> > This makes sense to me! Maybe an irrelevant question (and not one to address in this PR), is there are a reason we didn't go [this route](https://stackoverflow.com/a/3559555) to check to see if a partial exists and then have a default one if it doesn't?
> 
> Minno
> 
> > I like this - it means we'll only have custom views for truly custom views rather than a passthrough to the generic form
> 
> I don't think I want to go down that path, I don't see how it suits our case. Because our views **have to** exist, so it's not a case of falling back to some "default". But maybe I don't understand exactly how you'd use that. I imagined you'd use it inside the `new.html.erb`?
> 
> ```
> <%= render partial: "metadata_fields/#{@document.document_type.pluralize}", locals: { f: f } %
> ```

I just saw for a lot of the metadata partials, they were just rendering a shared form

`<%= render layout: "shared/specialist_document_form", locals: { f: f } do %>
<% end %>`

so thought you could do instead in `new.html.erb`:

```
<%= render "metadata_fields/#{@document.document_type.pluralize}" rescue render :partial => "shared/specialist_document_form" %>
```

Alternatively, it seems there's also `lookupcontext` and `template_exists?` which you could use to set the partial path in the controller if you wanted to avoid logic/branching in the view itself:

```
@metadata_partial_path = if template_exists?("metadata_fields/#{@document.document_type.pluralize}")
  "metadata_fields/#{@document.document_type.pluralize}"
else 
  "shared/specialist_document_form"
```

lauraghiorghisor-tw:
> I just saw for a lot of the metadata partials, they were just rendering a shared form

Oooh, OK, I think I am following - I can try that in a new PR ;) It feels like a refactor and a lot of file changes.
If nothing else, can this one be approved?

minhngocd (approved):

patrickpatrickpatrick:
> > I just saw for a lot of the metadata partials, they were just rendering a shared form
> 
> Oooh, OK, I think I am following - I can try that in a new PR ;) It feels like a refactor and a lot of file changes. If nothing else, can this one be approved?

Ah sure, that's probably for a later PR.

patrickpatrickpatrick (approved):

PR #3166, opened 2025-05-27
Decouple internal document type usage from downstream usage

lauraghiorghisor-tw:
## What

- Use `document_type` instead of `filter.format` for SP internals (views, forms etc.). The idea is that if we rename things, they're still internally consistent within SP. The `filter.format` value is strictly used for downstream/Search API business. For new finders we can always have the document type and format match (this has now been added as guidance, and included in documentation).
- Add the possibility to specify a `document_type` that does not match the model name. Open for discussion if this should be a default field on all schemas or only custom (currently only in the ESI Fund schema). This requires maintaining a list of mappings.

## Why

- Specialist Publisher used to deduce the `document_type` value that we send to Publishing API from the corresponding model class name (`to_s.underscore`). Nonetheless, there are historical cases where this is not the case, which can be confusing to track, as proven by the work originally done in https://github.com/alphagov/specialist-publisher/pull/3166. ESI Funds is the outlier where the naming conventions have not been respected (has a different `filter.format` value than what the model name). ESI fund has been used in this PR to prove the efficacy of the changes.
- It's easy to assume that the schema set `filter.format` is the "official" document type (the format does typically match our model names, but it is not used in the presenter; it is used in Search API). These changes make the usages of `document_type`/`downstream_document_type` more obvious. 
- We are currently unable to rename a model, because that would change the `document_type` value. There is no way to rename files without causing downstream issues. This PR aims to address this coupling.

## Testing

- Manually tested as part of the original PR
- Briefly retested after PR review changes

---

Split out from https://github.com/alphagov/specialist-publisher/pull/3151.


app/models/finder_schema.rb
```diff
@@ -34,6 +34,7 @@ def self.document_models
   attribute :description
   attribute :document_noun
   attribute :document_title
+  attribute :downstream_document_type
```
lauraghiorghisor-tw, line 37: From @minhngocd https://github.com/alphagov/specialist-publisher/pull/3151/commits/ea7452f74e2b1a5da56a1cf06026905f038badaf#r2100196978

> Might I suggest public_content_type, sort of in the same vein of public_updated_at_date, to show the contrast being public facing, and internal system; rather than upstream / downstream?

> I also wonder if having a doc within this repo for Ubiquitous Language might be helpful to clarify the differences between
- document type
- content type
- document model
- document noun
etc.


I feel like the `public_` terminology suggests user-facing functions, whilst this is a very backendy use-case. I am happy with downstream 😬

Would the changes in https://github.com/alphagov/specialist-publisher/pull/3165/commits/a4385c9d585ee4c411054c7640b041f1c2e0d9f5 + [the changes in this PR ](https://github.com/alphagov/specialist-publisher/pull/3166/commits/76f1077d1b2ea2e5bf0eb5335ea5741c58aeb98f)cover the documentation need?

app/helpers/document_type_mapper.rb
```diff
@@ -0,0 +1,9 @@
+module DocumentTypeMapper
+  OUTLIER_DOCUMENT_TYPES = {
```
lauraghiorghisor-tw, line 2: @minhngocd https://github.com/alphagov/specialist-publisher/pull/3151/commits/81aab5d132b0692dda566a2510b1a3db78e3e054#r2100218052

> Might be better to hold a hash rather than switch case, being more performant. Additionally, if the document models themselves hold the public document types defined, then we probably should use its enum instead of having repeat strings across the codebase?

This is the change. Does that work for you? Did not quite follow what you meant with the enum.

spec/fixtures/factories.rb
```diff
@@ -163,9 +163,10 @@
       merged_details = default_details.deep_stringify_keys.deep_merge(details.deep_stringify_keys)
       result = attributes.merge(details: merged_details)
       if document_type
+        specialist_publisher_document_type = DocumentTypeMapper.get_document_type(document_type)
```
lauraghiorghisor-tw, line 166: @minhngocd https://github.com/alphagov/specialist-publisher/pull/3151/commits/81aab5d132b0692dda566a2510b1a3db78e3e054#r2100211027

> Why do we have a duplicate mapper just for specs? can the specs not import and use the real mapper? probably shouldn't have a mapper just for specs as they can diverge.

Done ✅

app/lib/publisher.rb
```diff
@@ -49,11 +49,12 @@ def content_id_and_locale_pairs_for_document_type(document_type)
   end
 
   def document_types
+    Rails.application.eager_load!
     @document_types ||= all_document_types
   end
 
   def all_document_types
     Rails.application.eager_load!
```
lauraghiorghisor-tw, line 57: @minhngocd https://github.com/alphagov/specialist-publisher/pull/3151/commits/8b37a946ff502ad1519084aa32a00ca6bd739ab6#r2100223658

> surely this should have been done in an initializer somewhere?
If not, maybe it could live more in the initializer configs - e.g. preload_document_types or something.

> I see - that line was an existing piece of code that lived publisher and republisher. I think those usages there since they lived in app/lib had to do eager load because it is running a rake task, and does not have the rails application loaded. In which case, would make sense to separate the preloading just for rake tasks. instead of having them wrapped within this helper class here, since if it is called within the rails app, it shouldn't need to call this line.

Changed so that the eager loading is in the rake task where needed, and only the document fetching is in the Mapper ✅

app/lib/documents_tagger.rb
```diff
@@ -49,7 +45,7 @@ def editions_enum(params = {})
   def all_documents
```
lauraghiorghisor-tw: Reverted [changes](https://github.com/alphagov/specialist-publisher/pull/3151/commits/8b37a946ff502ad1519084aa32a00ca6bd739ab6#diff-a5dd699042305637e714bcf92725b4aa3c9dc56cfbf3fc7b761ec15eb9199d4f) here as no need to use mapper.

app/helpers/document_type_mapper.rb
```diff
@@ -6,4 +6,8 @@ module DocumentTypeMapper
   def self.get_document_type(document_type)
     OUTLIER_DOCUMENT_TYPES[document_type] || document_type
   end
+
+  def self.all_document_types
```
lauraghiorghisor-tw, line 10: @minhngocd https://github.com/alphagov/specialist-publisher/pull/3151/commits/8b37a946ff502ad1519084aa32a00ca6bd739ab6#r2100228492

> I'm not sure about the impact of delete - haven't really tested, but can exclude be used instead here? or reject? for excluding Attachment as a subclass

Was fine with delete but changed to reject ✅

lauraghiorghisor-tw:
Closing as decided it's not needed yet.

PR #3176, opened 2025-06-02
Migrate licence transactions to the design system

lauraghiorghisor-tw:
## What

Migrate non-conforming document type `LicenceTransaction` to the design system.

Testing:
- Replicated existing tests for design system.
- Custom model fields and validations are covered in custom model test `licence_transaction_spec.rb`.

View:
- Replicated existing view with DS components
- Opted to manually render fields, due to view structure (custom text in the middle, custom fields in the middle of schema fields)
- "Publishing organisation" field now rendered with the new `OrganisationSingleSelectComponent`
- "Organisations" field still needs migrating, once we pull in the multiple select
- Used a fieldset component to render the "How users apply" section.

Changed behaviour: 
- Previously the `base` -> `link_and_identifier_exists` error did not wrap the section relevant to it; now the relevant fields have been moved inside a fieldset on which the error is defined.
- Previously, on publication error "This document's URL is already used on GOV.UK. You can't publish it until you change the title.", the `link_and_identifier_exists` error would also show - because the check was simply for any :base errors. Now we check for :base + if it is that specific error.

`ErrorSummaryComponent` logic:
- Reusing the logic from `DocumentsController` - `error_messages` method - to render messages, rather than full messages, when the field is registered as "custom error message field"; currently only the LicenceTransactions have such custom fields.

--- 

## Screenshots

| Link or identifier error | |
| --- | --- | 
| Legacy | Design system (changed) | 
| legacy how users apply error | DS how users apply error |
| Guidance showing | |
| Legacy | Design system (unchanged) |
| legacy guidance showing when error is fixed | DS guidanc eshowing when error is fixed |
| Incorrect error behaviour (there is no field error, just a publishing error) | |
| Legacy | Design system (changed) |
| Legacy base error erratic behaviour when other base errors are present | Screenshot 2025-06-02 at 19 41 31 |

[Trello](https://trello.com/c/Bl6vFuT4/3656-design-system-add-new-document-non-conforming-pages)


patrickpatrickpatrick (approved):
LGTM!

PR #3180, opened 2025-06-03
Transition Document Summary page to the new Design System

minhngocd:
- In moving actions to the sidebar according to design, we are losing some useful text such as explanations where publishing will trigger email alerts, or that the Unpublishing button isn't available because there is an outstanding draft. I've introduced a temporary section below the actions box to display these useful Publishing and Unpublishing notifications. Will need to run them by product and design. (as indicated in screenshots below)
- Once we have confirmation pages for Publishing, we can move the publishing notification messages there
- More trouble with the Unpublishing notification messages about the Unpublishing button being unavailable due to an existing draft - we probably need to think about where to place it. One option is to always show Unpublish button if the document is published, and on the confirmation page warn users that there is a draft and they can't do it unless draft is published or discarded, but this is a waste of one user click to find out that an action is not actionable. Alternative is to keep this message displayed on summary page.
- None of the sidebar action buttons currently do anything. This is because they are all GETs which in the future should link to "confirm delete" or "confirm publish" type pages. It is more trouble to wire them up to do a POST to keep current functionality working. Since this behind a feature flag, we can descope these flows to future tickets.

New draft: 
Screenshot 2025-06-03 at 23 31 50

Published:
Screenshot 2025-06-03 at 23 32 49

Published with new draft:
Screenshot 2025-06-03 at 23 35 04

With preview permissions turned off (old design):
Screenshot 2025-06-03 at 23 44 36

https://trello.com/c/fiL2r13D/3717-design-system-document-show-page


minhngocd:
Update from standup discussion: 

- Publishing message : This can be moved to the confirm publish page for the next Trello ticket addressing the Publishing Flow
- Unpublish message : Style it to look more like an inset prompt in Whitehall. When the inset prompt from Whitehall is moved into Publishing Components, switch to using inset prompt.

minhngocd:
Restyled the notices section to look like the inset prompt on Whitehall

image

spec/features/design_system/viewing_the_admin_summary_for_cma_cases_design_system_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
```
lauraghiorghisor-tw, line 1: Out of curiosity was there much to change to pass these tests (in the first commit)?

app/views/layouts/design_system.html.erb
```diff
@@ -47,8 +47,8 @@
     <main class="govuk-main-wrapper<%= " govuk-main-wrapper--l" if yield(:back_link).blank? && yield(:breadcrumbs).blank? %>" id="main-content" role="main">
 
       <%= render "govuk_publishing_components/components/success_alert", {
-        message: flash[:notice]
-      } if flash[:notice] %>
+        message: flash[:notice] || flash[:success]
```
lauraghiorghisor-tw, line 50: Why is this changing - can't seem to track where this is being introduced (as of this commit anyway).

app/views/documents/_show_summary.html.erb
```diff
@@ -49,44 +49,34 @@
...
-</div>
+<section class="app-view-summary__section govuk-!-margin-top-8">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Attachments",
+    heading_level: 2,
+    font_size: "l",
+    margin_bottom: 3,
+  } %>
+
+  <% attachments = @document.attachments || [] %>
```
lauraghiorghisor-tw, line 60: I agree with skipping truncating logic.

app/components/sidebar_actions_component.rb
```diff
@@ -0,0 +1,78 @@
...
+      discard_draft_action,
+    ].compact
+  end
+
+  def notices
+    notices = []
+
+    notices << notice("Publishing", @presenter.publish_text) if @presenter.publish_button_visible?
+
+    notices << notice("Unpublishing", @presenter.unpublish_text)
```
lauraghiorghisor-tw, line 26: Oh nice little method `unpublish_text` - didn't know that existed.

app/components/sidebar_actions_component.rb
```diff
@@ -0,0 +1,78 @@
+class SidebarActionsComponent < ViewComponent::Base
```
lauraghiorghisor-tw, line 1: I am wondering if releasing this whilst we're working on other things is going to make manual testing/local debugging really annoying - in that you cannot actually action anything whilst on the DS permission. We're still to drag in the single select and some weird stuff around an `authors` field.  

I hope it's enough to save the document (which is a POST). Hmm...

lauraghiorghisor-tw (approved):
Nothing blocking - LGTM 🚚

Just wondering if we should release it now because of the missing actions in the sidebar - but oh well - cross that bridge when we get there.

spec/features/design_system/viewing_the_admin_summary_for_cma_cases_design_system_spec.rb
minhngocd: Yes, i had to fix the flash notice for the design system pages, because the tests were checking for success messages, and there were none that were being shown because flash messages were broken for design system

app/views/layouts/design_system.html.erb
minhngocd: Because we set flash[:success] in the controllers, not flash[:notice] - i suspect the flash[:notice] came from copying over from Whitehall. Maybe in the future we would want to save them as :success, but for now, to keep everything working, i'm supporting both success and notice

app/components/sidebar_actions_component.rb
minhngocd: They were the little texts that appeared in the Actions sub sections for each of the actions.

app/components/sidebar_actions_component.rb
minhngocd: I think it should be fine. The thing you could do is switch back to legacy view if you wanted to action something. e.g. i had run this constantly on my rails c console: 
```
u = User.last; u.permissions << "preview_design_system"; u.save;
```
```
u = User.last; u.permissions.delete "preview_design_system"; u.save;
```

It's likely we'll be picking up the other tickets quite soon anyways - i'm looking to pick them up right after this - they're in the backlog, just fleshing them out and linking them in the mural.

PR #3188, opened 2025-06-06
Replace add another with multiple select

lauraghiorghisor-tw:
Replace add another usage with select with search (multiple), in:
- Facets
- Custom/non-conforming `:organisations` field.

Screenshot 2025-06-06 at 13 27 56

Screenshot 2025-06-06 at 17 17 54

[Trello](https://trello.com/c/pTuNpxzZ/3742-design-system-add-new-document-replace-add-another-component-with-select-with-search-multiple-variant)

### DIFF
--- a/app/components/facet_input_component/organisation_multi_select_component.rb
+++ b/app/components/facet_input_component/organisation_multi_select_component.rb
@@ -1,0 +1,27 @@
+class FacetInputComponent::OrganisationMultiSelectComponent < ViewComponent::Base
+  include ErrorsHelper
+  include OrganisationsHelper
+
+  def initialize(document, document_type, facet_key, facet_name)
+    @document = document
+    @document_type = document_type
+    @facet_key = facet_key
+    @facet_name = facet_name
+    @error_items = errors_for(@document.errors, @facet_key)
+    @options = select_options
+  end
+
+  def select_options
+    selected_values = @document.send(@facet_key)
+
+    all_organisations
+      .sort_by { |org| org.title.downcase.strip }
+      .map do |organisation|
+      {
+        text: organisation.title,
+        value: organisation.content_id,
+        selected: selected_values&.include?(organisation.content_id),
+      }
+    end
+  end
+end
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "accessible-autocomplete-multiselect": "alphagov/accessible-autocomplete-multiselect",
+    "choices.js": "^11.1.0",
     "paste-html-to-govspeak": "^0.5.0"
   },
   "devDependencies": {
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source "https://rubygems.org"
+git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
 gem "rails", "~> 8.0.2"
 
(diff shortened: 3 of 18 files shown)


lib/documents/schemas/design_decisions.json
```diff
@@ -113,6 +113,6 @@
   ],
   "show_summaries": true,
   "show_metadata_block": true,
-  "summary": "Results of past design and design right hearing decisions issued since 2013. Decisions published between 1998 - 2012 are available on the <a href=\"https://webarchive.nationalarchives.gov.uk/ukgwa/20140603095013/http://www.ipo.gov.uk/pro-types/pro-design/d-decisions.htm\">National Archives</a>. Please note, no decisions were issued during 2003.\r\n\r\nYou can use the free search box below to search by decision number.",
+  "summary": "Results of past design and design right hearing decisions issued since 2013. Decisions published between 1998 - 2020 are available on the <a href=\"https://webarchive.nationalarchives.gov.uk/ukgwa/20140603095013/http://www.ipo.gov.uk/pro-types/pro-design/d-decisions.htm\">National Archives</a>. Please note, no decisions were issued during 2003.\r\n\r\nYou can use the free search box below to search by decision number or British Library number.",
```
ChrisBAshton: Following the link, decisions only go up to 2014...? Why did the year change to 2020 here?

package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "accessible-autocomplete-multiselect": "alphagov/accessible-autocomplete-multiselect",
+    "choices.js": "^11.1.0",
```
ChrisBAshton, line 36: Out of scope for this PR but shouldn't this have moved into the select-with-search component so that consuming apps don't need to know about it? 🤔 
Hopefully we can resolve that in the port to govuk_publishing_components 🙏

app/components/facet_input_component/organisation_multi_select_component.rb
```diff
@@ -0,0 +1,27 @@
+class FacetInputComponent::OrganisationMultiSelectComponent < ViewComponent::Base
```
ChrisBAshton, line 1: Seems a reasonable call to make right now 👍 
Do we have a backlog item to address the non-conforming organisations field?

ChrisBAshton (approved):
LGTM ⭐ non-blocking comments below.

package.json
lauraghiorghisor-tw: Hmm that's a fair shout - I would not have expected to need to do this yeah 🤔
I added a note to [that card](https://trello.com/c/GYs1LCI7/970-move-select-with-search-into-govukpublishingcomponents?filter=component).

lib/documents/schemas/design_decisions.json
lauraghiorghisor-tw: This commit had no business here (rebasing gone wrong?). In terms of why the year changed, they asked for it - "We’ve decided to migrate decisions from 2020 onwards [...]" This finder is not live yet so I guess it might be sorted by the time it's live.

app/components/facet_input_component/organisation_multi_select_component.rb
lauraghiorghisor-tw: I've captured what I know so far in https://trello.com/c/oN6jDHYd/948-unify-sp-views-wip-card.

ChrisBAshton: Thanks!

PR #3190, opened 2025-06-06
Replace autocomplete with select

lauraghiorghisor-tw:
## What

These changes remove that last (and only) instance of the gem in Specialist Publisher. It was only used in the admin area to render organisations. In its stead, we are now using the multiselect with search.

## Why

We are planning to decommission the autocomplete gem.

## Testing that existing behaviour is preserved

**NB:** This can be released now, although it is not behind a feature flag. The only use is in the admin area which was already DS. Tested in integration (see below) - and validated a few happy paths from random document types to ensure nothing was unintentionally deleted.

- Renders the existing field values
Screenshot 2025-06-09 at 18 26 14

- Allows selection of new values
Screenshot 2025-06-09 at 18 26 21

- Uses selected values in the summary
Screenshot 2025-06-09 at 18 26 32

- Values in the schema diff
Screenshot 2025-06-09 at 18 28 21

[Trello](https://trello.com/c/OptIzkrQ/3798-remove-accessible-autocomplete-multiselect-from-specialist-publisher)


app/views/components/docs/_select_with_search.yml
```diff
@@ -0,0 +1,204 @@
+name: Select with search
```
lauraghiorghisor-tw: @matthillco I realised WH had an example of this so I copied it over - would be good to see it run locally, no sure how to do test that.

matthillco (approved):
Looks good to me, no further comments.

app/views/components/docs/_select_with_search.yml
lauraghiorghisor-tw: All tested here ✅

PR #3194, opened 2025-06-10
Replace single drop down with select with search

lauraghiorghisor-tw:
Covered:
- Single select on "conforming" pages - introduced a new select with search as all selects on these forms must be searchable.
- "Publishing Organisation" field on non-conforming pages - also a single select with search, but in a custom component for organisations.
- Language drop down.

Decisions:
- Keep a non-search single select component for fields that should not be searcheable. There are a few of these on the non-conforming pages - see before/after table below.
- There are other usages of the no-search single select in the codebase, in the admin area. Because the components were defined primarily with the forms and fields/facets in mind, they are not very well suited for reuse in the admin area. It's more readable to just render the select straight from publishing components, rather than try to reuse our `SingleSelectComponent`.

| Before | After (not changed) |
| --- | --- |
| Screenshot 2025-06-10 at 22 19 27 | Screenshot 2025-06-10 at 22 19 33 |
| int | Screenshot 2025-06-11 at 08 53 05 |
| Screenshot 2025-06-10 at 22 03 10 | Screenshot 2025-06-10 at 22 03 15 |
| Screenshot 2025-06-12 at 15 35 11 | Screenshot 2025-06-12 at 15 36 15 |

[Trello](https://trello.com/c/6mkl64lT/3744-design-system-add-new-document-replace-single-drop-down-with-select-with-search)


matthillco:
@lauraghiorghisor-tw 

I've tested the three non-conforming pages amended in the PR and they behave as expected now. Nice work.

I note that the `select` elements for any Organisation dropdowns don't work locally (empty) but that's an expected behaviour.

I have a couple of other thoughts:

- Should the language select now use `select_with_search` too? It's quite a long list and might benefit from search.
- All the `select_with_search` instances are missing a visual indicator (down arrow) that would show how they will behave. However I don't think we should deal with that in this PR as it's likely a fundamental change in the component repo itself. Can be ignored for now, but I'll make a note of it as something extra to explore (backlog).

minhngocd (approved):
Seems fine - i'm of the opinion maybe we nuke single select and only support single select with search since the search is not a mandatory feature the user has to use. But I also am not opposed to having two components that are used for different purposes.

matthillco (approved):
If we're not adding this select to Languages, then it all looks good to me. Ship it! :)

lauraghiorghisor-tw:
> If we're not adding this select to Languages, then it all looks good to me. Ship it! :)

Pushed the language change + added alphabetical sorting. The sorting will go out for legacy too.

PR #3197, opened 2025-06-11
[Design System] Publishing Flow

minhngocd:
Added interstitial page for publishing a document. 

- Publishing page contains text depending on the state of the document
- This text has been moved from the sidebar into the interstitial page as discussed in [previous PR](https://github.com/alphagov/specialist-publisher/pull/3180).
- Figma Design has all the text on the same line, but I've added a break to separate the confirmation question from the explanation of the impact. 

Publishing confirmation for new draft: 
Screenshot 2025-06-11 at 10 26 47

Publishing confirmation for published with new draft (major update):
Screenshot 2025-06-11 at 10 20 00

Publishing confirmation for published with new draft (minor update):
Screenshot 2025-06-11 at 10 22 38

Sidebar: 
Screenshot 2025-06-11 at 10 27 03

QA Note 

- Even if the publishing button doesn't show (e.g. unpublished document, published document), there's nothing stopping the user from navigating to the `/confirm_publish` part of the document link and clicking publish. This is currently returning the following error so there's no broken flow, but is not super user friendly. As it's not a valid user flow, only happens if users force going to the URL, and it does not introduce any bugs, feel like we can descope handling of this case from this PR/ticket. There may be a wider discussion to have about how to handle more granular authorisation and status checks for controller methods in Specialist Publisher.

Screenshot 2025-06-11 at 11 03 46

https://trello.com/c/TnRzQDr5/3796-design-system-publishing-flow

### DIFF
--- a/app/presenters/actions_presenter.rb
+++ b/app/presenters/actions_presenter.rb
@@ -18,6 +18,25 @@ class ActionsPresenter
   end
 
   def publish_text
+    if update_type == "minor"
+      safe_join([
+        tag.p("You are about to publish a minor edit."),
+        tag.p("Are you sure you want to publish this document?"),
+      ])
+    elsif update_type == "major" && !document.first_draft?
+      safe_join([
+        tag.p("You are about to publish a major edit with a public change note. Publishing will email subscribers to #{klass_name}."),
+        tag.p("Are you sure you want to publish this document?"),
+      ])
+    else
+      safe_join([
+        tag.p("Publishing will email subscribers to #{klass_name}."),
+        tag.p("Are you sure you want to publish this document?"),
+      ])
+    end
+  end
+
+  def publish_text_legacy
     if state == "published"
       "There are no changes to publish."
     elsif state == "unpublished"
@@ -45,6 +64,10 @@ class ActionsPresenter
     else
       "Publishing will email subscribers to #{document.class.title.pluralize}. Continue?"
     end
+  end
+
+  def confirm_publish_path
+    confirm_publish_document_path(slug, document.content_id_and_locale)
   end
 
   def publish_path
--- a/spec/components/sidebar_actions_component_spec.rb
+++ b/spec/components/sidebar_actions_component_spec.rb
@@ -1,0 +1,187 @@
+require "rails_helper"
+
+RSpec.describe SidebarActionsComponent, type: :component do
+  describe "edit button" do
+    it "should show on draft document" do
+      content_item = FactoryBot.create(:cma_case, :draft, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to have_link("Edit document")
+    end
+
+    it "should show on published document" do
+      content_item = FactoryBot.create(:cma_case, :published, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to have_link("Edit document")
+    end
+
+    it "should show on published with draft document" do
+      content_item = FactoryBot.create(:cma_case, :redrafted, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to have_link("Edit document")
+    end
+
+    it "should show on unpublished document" do
+      content_item = FactoryBot.create(:cma_case, :unpublished, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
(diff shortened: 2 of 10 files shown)


spec/components/sidebar_actions_component_spec.rb
```diff
@@ -0,0 +1,187 @@
+require "rails_helper"
```
lauraghiorghisor-tw, line 1: Think the last cma test was also covering the "The document is unpublished. You need to create a new draft before it can be published."  - will that surface on subsequent PRs?

The commit message gives the impression we're about to remove some tests but the commit then adds to massive test files. Perhaps worth making it clear which feature test files we're "replicating" and have moved some coverage further down the pyramid for.

I couldn't quite figure out which integration test you are referring to - as far as I can tell you have already migrated that coverage into the new DS test from `publishing_a_previously_unpublished_cma_case_spec.rb`. Maybe worth mentioning in the commit which test and removing the "future" guidance.

lauraghiorghisor-tw (approved):
Nice! LGTM - nothing blocking.

Some thoughts:
First commit could use some styling (extra full stop on first line, weird case on last) and a bit more why.

I wonder if those lines that used to be around the actions (that are now in the confirmation page) could be styled as some type of "information" (callout too much maybe?), with the actual "are you sure" being more... alarming. The text is a bit homogenous right now, I don't feel either alarmed nor informed.

As I am going through the PRs (next actions), it does feel like maybe the attachments specific tests can be grouped into their own test (out of scope).

Publish text. So we've lost:
- "The document is unpublished. You need to create a new draft before it can be published." - can live without.
-  "There are no changes to publish." (now button just doesn't show)
- "You don't have permission to publish this document." -> I **feel like this one would still be useful on the interstitial page, no?** Surely WH also shows a permissions error on there.

spec/components/sidebar_actions_component_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+      content_item = FactoryBot.create(:cma_case, :unpublished, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to_not have_link("Unpublish document")
+      expect(page).to have_text("The document is already unpublished.")
+    end
+
+    it "should not show when user not allowed to publish" do
```
lauraghiorghisor-tw: UNpublish?

spec/components/sidebar_actions_component_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+    it "should not show on unpublished document" do
+      content_item = FactoryBot.create(:cma_case, :unpublished, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to_not have_link("Delete draft")
+    end
+
+    it "should not show when user not allowed to discard draft" do
```
lauraghiorghisor-tw, line 178: Can we include the  "You don't have permission to discard this draft." expectation?

app/presenters/actions_presenter.rb
```diff
@@ -18,6 +18,25 @@ def publish_button_visible?
   end
 
   def publish_text
+    if update_type == "minor"
```
lauraghiorghisor-tw, line 21: Should we not keep the permissions text here? Rather than not showing the button on the previous screen? 

I am just thinking that we do get some zendesk tickets around permissions and it might be good to be explicit about why they're not seeing something. I think on WH we show it on the interstitial.

minhngocd: I can add the permissions text as part of the notices on the document summary same as unpublishing, maybe - rather than on the interstitial page. Feels uneccessary to lead the user somewhere else to tell them they can't execute the action.

spec/components/sidebar_actions_component_spec.rb
minhngocd: what do you mean? is this not it?

minhngocd:
> Nice! LGTM - nothing blocking.
> 
> Some thoughts: First commit could use some styling (extra full stop on first line, weird case on last) and a bit more why.
> 
> I wonder if those lines that used to be around the actions (that are now in the confirmation page) could be styled as some type of "information" (callout too much maybe?), with the actual "are you sure" being more... alarming. The text is a bit homogenous right now, I don't feel either alarmed nor informed.
> 
> As I am going through the PRs (next actions), it does feel like maybe the attachments specific tests can be grouped into their own test (out of scope).
> 
> Publish text. So we've lost:
> 
> * "The document is unpublished. You need to create a new draft before it can be published." - can live without.
> * "There are no changes to publish." (now button just doesn't show)
> * "You don't have permission to publish this document." -> I **feel like this one would still be useful on the interstitial page, no?** Surely WH also shows a permissions error on there.

The styling of the text displayed on the interstitial page follows the designs on Figma. If we want to make any changes there I would suggest involving design folks later down the line.

minhngocd:
Discussed in standup - i will add the permissions text below the actions thing as a separate PR

PR #3198, opened 2025-06-11
Selects with a multiple attribute should show several options (no JS)

matthillco:
## What

When JS is unavailable, `select` elements with a `multiple` attribute should have their height increased to show more options than are currently displayed. [Trello](https://trello.com/c/6mkl64lT/3744-design-system-add-new-document-replace-single-drop-down-with-select-with-search)

## Why

Select elements are set to 2.5rem high, which is defined in the CSS of `govuk-frontend`. This has a detrimental affect on selects using the `multiple` attribute as they only display one option. It's even worse if the first option is blank as the select appears to be empty, and it's easy to miss the tiny scroll bar that reveals more options.

Setting the height of these variant selects to `auto` will ensure that they display several options within the scrolling area and therefore will provide an improved UX.

Note that in practice this only applies to the non-JS experience: selects with a `multiple` attribute are entirely replaced in the JS experience with markup generated by the third party plugin [choices.js](https://github.com/Choices-js/Choices).

## Before / After

| Before | After |
| --- | ----------- |
| image | image |

## Future consideration

This fix should eventually apply to all apps that might use `select_with_search` with multiple variants. Therefore it should be made available in a higher root dependency, for example the publishing gem or the design system.

-------------------------

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -41,3 +41,7 @@ $govuk-page-width: 1140px;
 .bg-light-grey {
   background-color: govuk-colour("light-grey");
 }
+
+.govuk-select[multiple] {
+  height: auto;
+}


lauraghiorghisor-tw (approved):
Nice one 💖

minhngocd (changes requested):
Thank you for fixing this @matthillco !! I think this problem exists in Whitehall as well - so i think we probably want to merge it into select-with-search repo rather than just in specialist publisher. 

after we merge the select with search repo, we might need to bundle install again on specialist publisher and whitehall to pull in latest version of select-with-search.

matthillco:
Yep, we could do it that way, but I figured that since all these selects are deriving from the design system component, it would make more sense to fix it in there -- then it would eventually trickle down into every app.

matthillco:
@minhngocd I thought abut it overnight and I'll follow your suggestion. I'll raise this same PR in [alphagov/select-with-search-component](https://github.com/alphagov/select-with-search-component) and close this one. We can decide later if we want to get this fix into the design system.

matthillco:
Closing this one as it's now covered by a similar PR in [select-with-search-component](https://github.com/alphagov/select-with-search-component/pull/25).

PR #3199, opened 2025-06-11
[Design system] Unpublishing flow

minhngocd:
Transitioning Unpublishing flow to the design system. 
Added interstitial page for unpublishing a document.

- In the old world, there used to be a permanent "Unpublishing" section which used to show a message of why the unpublish button wasn't showing or what the impact would be of unpublishing.
- Now that we have an interstital page, we are duplicating the information when the button is visible. Text from unpublishing notice relating to Unpublishing on the document summary sidebar has been moved to the confirmation page.
- Also removing some other scenarios and instances of the notice messages appearing such as: 
  - when document is already unpublished or when it is first draft. These can be seen on the publish status of the document. 
  - The only relevant useful message we should show users is when they expect to see an unpublishing button but it's not there. This is the case where they have a publish with draft, or they don't have permission. these messages will still be showing.

Screenshot 2025-06-11 at 16 19 16

Screenshot 2025-06-11 at 16 19 24

No unpublishing notice on Unpublished document:
Screenshot 2025-06-11 at 16 19 36

Unpublishing notice: existing draft
Screenshot 2025-06-11 at 16 20 59
Screenshot 2025-06-11 at 16 20 05

Unpublishing notice: no permission
Screenshot 2025-06-11 at 16 27 56

QA Note:

Even if the unpublishing button doesn't show (e.g. unpublished with draft), there's nothing stopping the user from navigating to the /confirm_unpublish part of the document link and clicking unpublish. This is currently returning the following error so there's no broken flow, but is not super user friendly. As it's not a valid user flow, only happens if users force going to the URL, and it does not introduce any bugs, feel like we can descope handling of this case from this PR/ticket. There may be a wider discussion to have about how to handle more granular authorisation and status checks for controller methods in Specialist Publisher. Similar case in https://github.com/alphagov/specialist-publisher/pull/3197 

image

https://trello.com/c/KHABJSW9/3801-design-system-unpublishing-flow

### DIFF
--- a/app/presenters/actions_presenter.rb
+++ b/app/presenters/actions_presenter.rb
@@ -83,6 +83,16 @@ class ActionsPresenter
   end
 
   def unpublish_text
+    return if document.first_draft?
+
+    if state == "draft"
+      "The document cannot be unpublished because it has a draft. You need to publish the draft first."
+    elsif !policy.unpublish?
+      "You don't have permission to unpublish this document."
+    end
+  end
+
+  def unpublish_text_legacy
     if document.first_draft?
       text = "The document has never been published."
     elsif state == "draft"
--- a/app/presenters/actions_presenter.rb
+++ b/app/presenters/actions_presenter.rb
@@ -70,6 +70,10 @@ class ActionsPresenter
     confirm_publish_document_path(slug, document.content_id_and_locale)
   end
 
+  def confirm_unpublish_path
+    confirm_unpublish_document_path(slug, document.content_id_and_locale)
+  end
+
   def publish_path
     publish_document_path(slug, document.content_id_and_locale)
   end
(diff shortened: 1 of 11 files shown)


lauraghiorghisor-tw (approved):
LGTM 🚚

app/presenters/actions_presenter.rb
```diff
@@ -79,6 +83,16 @@ def unpublish_button_visible?
   end
 
   def unpublish_text
+    return if document.first_draft?
+
+    if state == "draft"
+      "The document cannot be unpublished because it has a draft. You need to publish the draft first."
+    elsif !policy.unpublish?
+      "You don't have permission to unpublish this document."
```
lauraghiorghisor-tw, line 91: And for this one we do have the permissions thing on the interstitial (see my [previous PR](https://github.com/alphagov/specialist-publisher/pull/3197) comments). So maybe it would be good to be consistent.

minhngocd: This is not on the interstitial - it's on the document summary page. The notice that displays under the actions.

PR #3202, opened 2025-06-12
[Design system] Discard Draft flow

minhngocd:
Added an interstitial page to confirm draft deletion. 

This PR also includes: 
- Bug fixing discard draft for editions that are not EN. Currently on integration we are not able to discard non-EN drafts. 
- Bug fixing the rendering of an error message. Currently on integration, the message renders with the a href html

The error message appears when you get a publishing API error. Reproduceable by: 
- navigating to `/confirm_discard` when there are no drafts
- navigating to `/confirm_unpublish` when there is a draft
- navigateing to `confirm_publish` when document is not publishable
This behaviour is already discussed in previous PRs https://github.com/alphagov/specialist-publisher/pull/3199 and https://github.com/alphagov/specialist-publisher/pull/3197

image

Error message Before:
image

Error message After: 
image

https://trello.com/c/uX6Wsh6Z/3802-design-system-delete-draft-flow

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -7,7 +7,7 @@ class DocumentsController < ApplicationController
   include OrganisationsHelper
 
   layout :get_layout
-  DESIGN_SYSTEM_MIGRATED_ACTIONS = %w[new create index show confirm_publish confirm_unpublish].freeze
+  DESIGN_SYSTEM_MIGRATED_ACTIONS = %w[new create index show confirm_publish confirm_unpublish confirm_discard discard].freeze
   include DesignSystemHelper
 
   before_action :fetch_document, except: %i[index new create]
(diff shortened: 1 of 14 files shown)


app/controllers/documents_controller.rb
```diff
@@ -91,9 +91,16 @@ def unpublish
     redirect_to document_path(current_format.admin_slug, params[:content_id_and_locale])
   end
 
+  def confirm_discard; end
+
   def discard
     if @document.discard
-      flash[:success] = "Discarded draft of #{@document.title}"
+      message = if get_layout == "design_system"
+                  "The draft of '#{@document.title}' has been deleted"
```
lauraghiorghisor-tw, line 99: Very opinionated. Is this to match wh?

app/presenters/actions_presenter.rb
```diff
@@ -122,7 +117,7 @@ def discard_button_visible?
     policy.discard? && document.draft?
   end
 
-  def discard_text
+  def discard_text_legacy
```
lauraghiorghisor-tw, line 120: Same comment on the permission one as previous PRs.

lauraghiorghisor-tw (approved):
Nice one and nice bug catching! LGTM - although a bit unsure on having lost some of the permissions discard text 🤔

app/controllers/documents_controller.rb
minhngocd: Yes - but also as the terminology moved from Discard draft to Delete draft I thought to keep it consistent. Made the decision to match it from Whitehall as it wasn't part of the Figma designs

app/presenters/actions_presenter.rb
minhngocd: I think permissions difference is now just to not show the action. Feels a bit weird for the user to have to click through to find out they can't do a particular task. Can take it up at standup/sit down

PR #3206, opened 2025-06-13
Refactor views

lauraghiorghisor-tw:
## What

Refactor views so we don't have to maintain different view files of identical contents for conforming types. 

Previously, we maintained a view file for each document type. Most pages are what we call "conforming pages", in that their views are directly populated from the schema fields. These standard pages just render the `specialist_document_form` partial. The rest of the pages, non-conforming ones, must, for whatever reasons, implement a custom view with their own fields and sometimes custom logic.

These changes implement two controller helper methods to be used in the views, both DS and legacy variants, to help with computing which view to render. Using a `lookup_context` implementation, the methods look up a defined partial, and fall back on `specialist_document_form` if they cannot find the partial. 

The legacy method was added as to distinguish it from the DS helper method. This method should be deleted alongside the rest of the code when the DS is released.

The edit view has not yet been migrated to the Design System so it's still using the legacy partials.

## Why

We want to only ever have to define the custom views for non-conforming pages. This makes it more obvious which documents have custom views. It should help us better track how "generalised" the document types are and thus aid in our efforts to unify them so that everything is schema based.

## Technical considerations

This refactor was suggested by @patrickpatrickpatrick in a review comment [here](https://github.com/alphagov/specialist-publisher/pull/3165#issuecomment-2912176466).

I originally implemented the `rescue` suggestion, but I did not quite like the semantics of `rescue` in a code that was basically behaving as expected, so ended up going for `lookup_context`. 

These changes are a refactor as no behaviour has been changed.

## Testing

Tested locally with and without`preview_design_system` permissions.

### DIFF
--- a/app/controllers/documents_controller.rb
+++ b/app/controllers/documents_controller.rb
@@ -107,6 +107,15 @@ class DocumentsController < ApplicationController
     redirect_to documents_path(current_format.admin_slug)
   end
 
+  helper_method :computed_partial
+  def computed_partial
+    if lookup_context.exists?(@document.document_type.pluralize.to_s, %w[metadata_fields], true)
+      "metadata_fields/#{@document.document_type.pluralize}"
+    else
+      "shared/specialist_document_form"
+    end
+  end
+
 private
 
   def merge_nested_facet_fields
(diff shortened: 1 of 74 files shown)


lauraghiorghisor-tw:
@patrickpatrickpatrick / @minhngocd Feel free to merge / push any changes.

ChrisBAshton (approved):
Great improvement - thank you! 👏

PR #3209, opened 2025-06-16
Display permission notices for actions

minhngocd:
If actions are actionable but users are not permitted to action them, we now display a notice about this scenario in the notice box under the actions buttons in the sidebar. 

This is a direct follow-on from discussions on previous PRs. We don't want to lose the messaging should the user need to see them to explain why buttons are not there when they should be: 
- https://github.com/alphagov/specialist-publisher/pull/3197
- https://github.com/alphagov/specialist-publisher/pull/3199
- https://github.com/alphagov/specialist-publisher/pull/3202

Renamed the methods in actions_presenter that relate to this notice box to *_notice. This is to differentiate them from things like publishing_text where it is actually displayed on the interstitial page.

https://trello.com/c/5jxeMAlM/3815-design-system-edit-document-new-draft-published-unpublished-with-new-draft

Screenshot 2025-06-16 at 13 41 21

### DIFF
--- a/spec/components/sidebar_actions_component_spec.rb
+++ b/spec/components/sidebar_actions_component_spec.rb
@@ -94,7 +94,6 @@ RSpec.describe SidebarActionsComponent, type: :component do
       render_inline(described_class.new(document, user))
 
       expect(page).to_not have_link("Unpublish document")
-      expect(page).to_not have_selector(".app-view-summary__sidebar-notices")
     end
 
     it "should show on published document" do
@@ -104,7 +103,6 @@ RSpec.describe SidebarActionsComponent, type: :component do
       render_inline(described_class.new(document, user))
 
       expect(page).to have_link("Unpublish document")
-      expect(page).to_not have_selector(".app-view-summary__sidebar-notices")
     end
 
     it "should not show on published with draft document" do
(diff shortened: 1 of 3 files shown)


spec/components/sidebar_actions_component_spec.rb
```diff
@@ -184,4 +179,96 @@
...
+      expect(page).to_not have_selector(".app-view-summary__sidebar-notices")
+    end
+
+    it "should show unpublishing notice on published with draft document" do
+      content_item = FactoryBot.create(:cma_case, :redrafted, title: "Example CMA Case")
+      user = FactoryBot.create(:cma_editor)
+      document = DocumentBuilder.build(CmaCase, content_item)
+      render_inline(described_class.new(document, user))
+
+      expect(page).to have_text("The document cannot be unpublished because it has a draft. You need to publish the draft first.")
```
ryanb-gds, line 208: This seems a bit of a weird message - could we just not show anything? I don't know why a user would be expecting to be able to unpublish a draft

ryanb-gds (approved):
Looks good, just one thought about a possibly odd bit of behaviour from a user's perspective

spec/components/sidebar_actions_component_spec.rb
minhngocd: In this case this message will show on "published with draft", so user might expect to see an Unpublished button. 
If it's a first draft with no published edition, it won't show any messages.

PR #3210, opened 2025-06-16
[Design System] Edit Document flow

minhngocd:
Transitioning the Edit Document workflow to the new design system. 

Out of Scope: 
- Delete Attachment link. This is for a different ticket that will introduce a confirm delete page. For now the link does nothing.

Fixes: 
- This PR includes a number of fixes for the Edit flow to work. 
  - Drop down select options for nested facets
  - Date component loading existing value from document
  - Govspeak Help sidebar is made to be more consistent with Whitehall

Existing edit page on Integration: 
image

Existing sidebar (design system) on integration: 
Screenshot 2025-06-17 at 21 23 27

New edit page: 
Screenshot 2025-06-17 at 18 20 59

https://trello.com/c/5jxeMAlM/3815-design-system-edit-document-new-draft-published-unpublished-with-new-draft


app/views/attachments/_attachment_links.html.erb
```diff
@@ -7,30 +7,37 @@
...
+  <p class="govuk-body">
+    <%= link_to "Add attachments", new_document_attachment_path(params[:document_type_slug], @document.content_id_and_locale), class: "govuk-link govuk-link--no-visited-state" %>
+  </p>
 <% else %>
   <p class='govuk-body'>To add an attachment, please save the draft first.</p>
 <% end %>
+
 <% if @document.attachments %>
-    <ul class="attachments">
+    <div class="attachments">
```
ryanb-gds, line 18: Is there a reason you switched away from the list? I'd say the list is semantically a good choice.

app/helpers/facet_select_helper.rb
```diff
@@ -5,6 +5,15 @@ def select_options_for_facet(facet_values)
     }.reduce([], :concat)
   end
 
+  def admin_facet_value_from_allowed_values(allowed_values, nested_facet:)
+    values = allowed_values&.map do |value|
+      value_output = "#{value['label']} {#{value['value']}}"
+      nested_facet ? "#{value_output}\n#{value['sub_facets'].map { |sub_facet| "- #{sub_facet['label']} {#{sub_facet['value']}}" }&.join("\n")}" : value_output
+    end
+
+    nested_facet ? values&.join("\n\n") : values&.join("\n")
```
ryanb-gds: Probably shouldn't have to use the ampersand operator (must be a proper name for that) 3 times on the same value. Perhaps a guard clause at the top of the function would be a better choice?

app/components/facet_input_component/date_component.rb
```diff
@@ -1,11 +1,39 @@
...
   include ErrorsHelper
 
-  def initialize(document, document_type, facet_key, facet_name)
+  def initialize(document, document_type, facet_key, facet_name, params)
     @document = document
     @document_type = document_type
     @facet_key = facet_key
     @facet_name = facet_name
     @error_items = errors_for(document.errors, facet_key)
+    @params = params
```
ryanb-gds, line 10: I probably just don't understand, but shouldn't we be getting the values from the document model rather than the request parameters?

ryanb-gds (commented):
Just want to check the date state is being accessed in the approved Railsy way before approving

app/components/facet_input_component/date_component.rb
minhngocd: In the case where we are unable to parse the params into a single date, e.g. if the user enters: 
year: "not"
month: "a"
day: "date"

Then the "date" on the document model will be empty. That's why we prioritise the params value if it is passed in for rendering. I agree it's not ideal... i think it's because we're doing this back and forth parsing between the date field and the `1i`, `2i`, `3i` input fields on the form...

app/helpers/facet_select_helper.rb
minhngocd: ah yea - this is existing code moved from private to public section - but i can look at refactoring it.

app/views/attachments/_attachment_links.html.erb
minhngocd: I guess I did so because they're now blocks of elements that don't really look like lists: 
image

There's a lot of information and action links within the blocks, which I didn't feel were simple list behaviours.

app/components/facet_input_component/date_component.rb
ryanb-gds: Ah yep, just checked Whitehall and we had to do it there as well. I detest Rails sometimes.

app/views/attachments/_attachment_links.html.erb
ryanb-gds: I don't know if that matters really, but it's a minor point anyway

ryanb-gds (approved):
I think this is good to go :shipit:

PR #3223, opened 2025-06-23
[Design System] GovSpeak preview button with attachments preview

minhngocd:
To preview GovSpeak with rendered Inline Attachments, the document's list of attachments need to be made available to the call for preview. The rendering looks at both the GovSpeak body, and the list of attachments, and renders appropriate HTML. 

This PR does a number of notable things: 
- The show/hide toggling mechanism of the "Preview" and "Back to edit" buttons were broken due to CSS overrides. Fix this by making the selectors more specific
- Fix call to generate GovSpeak to be async. Previously it was a synchronous call which freezes up the UI. Users will see no activity and then the preview will be replaced all of a sudden. The preview generation could take over 10 seconds so is a little jarring. After async call fix, loading message displays properly, however it broke a feature test which has since been skipped.
- Render proper GovSpeak previews for bodies with inline attachments

Screenshot 2025-06-27 at 11 12 54

---

Sorry it's a bit cut off - had to be very quick with the screen grabbing even with 3G network throttling

Screenshot 2025-06-27 at 11 13 31

---

Screenshot 2025-06-27 at 11 13 01

https://trello.com/c/4pQulwQL/3833-design-system-edit-document-preview-button-render-attachment-links


spec/features/design_system/creating_a_research_for_development_output_spec.rb
```diff
@@ -72,7 +70,10 @@
       elsif properties["select"] == "one"
         select facet["allowed_values"].first["label"], from: facet["name"], match: :first
       elsif properties["select"] == "multiple"
-        select facet["allowed_values"].first["label"], from: "#{key}_0"
+        select facet["allowed_values"].first["label"], from: "#{document_type}[#{facet['key']}][]"
```
lauraghiorghisor-tw, line 73: Nice one!

spec/features/design_system/attachments_on_a_cma_case_design_system_spec.rb
```diff
@@ -0,0 +1,173 @@
+require "spec_helper"
```
lauraghiorghisor-tw, line 1: I see pub-api is already required in the spec helper https://github.com/alphagov/specialist-publisher/blob/7c55865e5350c7e3de07aeb2d91aa893317ef754/spec/spec_helper.rb#L34 - can probably just require the other two there.

spec/features/design_system/govspeak_design_system_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "spec_helper"
```
lauraghiorghisor-tw, line 1: Like the test split 👏

spec/features/design_system/govspeak_design_system_spec.rb
```diff
@@ -71,15 +71,15 @@
       stub_publishing_api_has_item(cma_case)
     end
 
-    scenario "previewing GovSpeak", js: true do
+    scenario "previewing GovSpeak", js: true, skip: "Test is stuck on Javascript loading message. Possibly Preview controller failing to return and ajax call is stuck. Unsure how to move forward from this. Scenario is tested manually and javascript is tested in Jasmine spec" do
```
lauraghiorghisor-tw, line 74: Alas - will keep it on my TODO list 📋

lauraghiorghisor-tw (approved):
LGTM 🚢
(except for those requires)

spec/features/design_system/attachments_on_a_cma_case_design_system_spec.rb
minhngocd: I feel it's also a little weird for publishing API to be a hidden require whereas the other two are explicit requires... I think Whitehall requires them all as in "all_adapters" from the spec helper, but unsure. Might be something to look at later down the line

lauraghiorghisor-tw: I would just require all in the spec helper - but up to you 👍🏻

minhngocd: I think i'll leave it like this for now for 3 reasons: 
- rebase is effort for small value
- adding the requires for asset manager and support api to spec helper requires them in all tests where they're not really needed
- removing just the publishing api require here makes it a little unbalanced since the include modules below it are also matching 3 in the same order. (purely aesthetics based). people might find it less confusing where the missing require is without having to dig into spec_helper

PR #3236, opened 2025-07-01
Ensure changenote timestamp is updated on document publish (WHIT-2263)

ChrisBAshton:
We used to rely on Publishing API to set this value for us, but there were two problems with that:

1. This weird email hack where we need to fetch the published document from Publishing API in order to fetch the 'public updated at' timestamp before manually calling Email Alert API.
2. More recently - and more pressing - we've been made aware of a bug whereby the latest changenote for a document takes its timestamp from the last time the document was edited _prior to publishing_. This means that if a document is drafted on 1st April and then published on 2nd April, the public changenote says 1st April, which can be misleading to users (it might suggest that the changes are effective from 1st April, rather than from 2nd April when the content was actually visible).

Setting the `public_updated_at` means:

1. No need to fetch the content item from Publishing API - we can proceed to our Email Alert API call immediately.
2. We explicitly tell Publishing API to use the `public_updated_at` in our payload, for its corresponding ChangeNote:  https://github.com/alphagov/publishing-api/blob/6143731ffad5db48476d7647a75413c42a5224fd/app/models/change_note.rb#L31

JIRA: https://gov-uk.atlassian.net/browse/WHIT-2263

---

### DIFF
--- a/app/services/document_publisher.rb
+++ b/app/services/document_publisher.rb
@@ -6,22 +6,15 @@ class DocumentPublisher
     if document.first_draft?
       document.change_note = "First published."
       document.update_type = "major"
-      document.save
     end
+    document.public_updated_at = Time.zone.now.strftime("%Y-%m-%dT%H:%M:%S%:z")
+    document.save
+
     Services.publishing_api.publish(document.content_id, nil, locale: document.locale)
-    published_document = document.class.find(document.content_id, document.locale)
 
     if document.send_email_on_publish?
-      # We don't have `public_updated_at` until the document is published, so we
-      # get it from the publishing-api and manually set it on the orignal document
-      # to preserve fields like `urgent` that are lost on publish.
-      # This special case will go away once specialist-publisher starts using the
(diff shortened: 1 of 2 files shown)


app/services/document_publisher.rb
```diff
@@ -6,22 +6,16 @@ def self.publish(document)
...
       document.change_note = "First published."
       document.update_type = "major"
-      document.save
     end
+    document.public_updated_at = Time.zone.now.strftime("%Y-%m-%dT%H:%M:%S%:z")
+    document.save
+
     Services.publishing_api.publish(document.content_id, nil, locale: document.locale)
-    published_document = document.class.find(document.content_id, document.locale)
+    document.class.find(document.content_id, document.locale)
```
lauraghiorghisor-tw: I don't imagine this line still does anything?

lauraghiorghisor-tw (approved):
LGTM

app/services/document_publisher.rb
ChrisBAshton: good spot!

PR #3250, opened 2025-07-10
[WHIT-2298] [Design System] Release QA - Change how we surface custom error messages

lauraghiorghisor-tw:
## Context

The issue we're trying to fix here is the rendering of custom error messages/fields. These are currently only defined for licence transactions in the `en.yml` I18n file. The `LicenceTransaction model` is marked as having custom fields through the `custom_error_message_fields` method. This is again consumed in the `DocumentsController` - `document_error_messages` method, to figure out the errors - either custom or full messages. All this controller logic is bypassed if we are on the design system.

In the previous implementation, we tapped into this logic in the `ErrorSummaryComponent` to render the correct errors for Licences. But we missed the fact that the validation messages on the fields themselves were not being done properly (they're full message mixed with custom, giving the rather unsightly `Licence transaction industry Select an industry`).

## Implementation

This PR fixes both the `ErrorSummary` and the field errors, as well as opening up the possibility of adding custom errors for subsequent models without having to go through the model or controller. The licence model and controller code will stay in place for legacy, but once we remove the DS toggle code, we will be able to remove that complexity (tested this locally ✨).

Alternatively, this also means we can remove the "customization of error messages" from licences/codebase by simply removing their definition from the yml file, if that's a preferable course of action. For now, we've decided to maintain content as is whilst going live with the Design System.

This is done by:
- moving the `get_text` method in the `ErrorsHelper` and reuse it for both fields and summary.
- make the `get_text` method tap into the yml file directly.
- we need to know what to search for in the yml file so we pass into `get_text`, the model "format" or document_type - which is how adding things to yml happens by convention.
- this has caused many changes where the `errors_for` or `errors_for_input` were used, but it's simply passing that param through.
- `errors_for` or `errors_for_input` - are mainly distinguished by the fact that they map to either an array, or an object with the `text` key in it, depending on what consumes them; for now we have to keep both.
- `errors_for` is used practically across all the components under `FacetInputComponent` so we could extract it at the top in the `FacetInputComponent` initializer - but I think we can live with the repetition.

### Screenshots
- licences before (weird inline error messages)
Screenshot 2025-07-10 at 13 30 59

- licences after 
> Screenshot 2025-07-10 at 13 32 21

- licences before (change note broken)
> Screenshot 2025-07-10 at 13 52 42

- licences after (change note ok)
> Screenshot 2025-07-10 at 13 52 56

- DS other finders' error showing correctly (no change) before (left) / after (right)
> Screenshot 2025-07-10 at 13 33 34
- DS this one has custom fields in the en.yml - still rendering as expected (no change) - before (left) / after (right)
Screenshot 2025-07-10 at 13 35 10

- Legacy - still displaying correctly (needs to still work)
> Screenshot 2025-07-10 at 13 47 21
> Screenshot 2025-07-10 at 13 48 00

- Attachments are not changed; they're actually not affected although we've changed their `errors_for` methods (the error is shown through flash not error summary)
> Screenshot 2025-07-10 at 13 56 44

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2298)

### DIFF
--- a/app/views/metadata_fields/_licence_transactions.html.erb
+++ b/app/views/metadata_fields/_licence_transactions.html.erb
@@ -15,7 +15,7 @@
   id: "#{format_name}_link_and_identifier_exists",
   error_message: errors_for_input(format_name, @document.errors, :base)&.include?(t("activemodel.errors.models.licence_transaction.attributes.base.link_and_identifier_exists")) ? errors_for_input(format_name, @document.errors, :base) : nil,
 } do %>
-  <% unless field_has_errors(@document, :base) %>
+  <% unless errors_for_input(format_name, @document.errors, :base) %>
     <p class="govuk-body govuk-!-margin-bottom-6">Enter either the website where users can apply for the licence, or the licence identifier.</p>
   <% end %>
 
--- a/app/views/metadata_fields/_licence_transactions.html.erb
+++ b/app/views/metadata_fields/_licence_transactions.html.erb
@@ -4,7 +4,7 @@
 <%= render FacetInputComponent.new(@document, schema.get_facet(:licence_transaction_industry), params, "Industry") %>
 
 <% # TODO: find a way of making organisation management schema-driven %>
-<% format_name = f.object.class.finder_schema.filter["format"].to_sym %>
+<% format_name = f.object.class.document_type.to_sym %>
 <%= render FacetInputComponent::OrganisationSingleSelectWithSearchComponent.new(@document, format_name, :primary_publishing_organisation, "Publishing organisation", selected_organisation_or_current(@document.primary_publishing_organisation)) %>
 <%= render FacetInputComponent::OrganisationMultiSelectWithSearchComponent.new(@document, format_name, :organisations, "Other associated organisations") %>
 
--- a/app/views/metadata_fields/_licence_transactions.html.erb
+++ b/app/views/metadata_fields/_licence_transactions.html.erb
@@ -13,7 +13,7 @@
   heading_level: 2,
   heading_size: "l",
   id: "#{format_name}_link_and_identifier_exists",
-  error_message: errors_for_input(@document.errors, :base)&.include?(t("activemodel.errors.models.licence_transaction.attributes.base.link_and_identifier_exists")) ? errors_for_input(@document.errors, :base) : nil,
+  error_message: errors_for_input(format_name, @document.errors, :base)&.include?(t("activemodel.errors.models.licence_transaction.attributes.base.link_and_identifier_exists")) ? errors_for_input(format_name, @document.errors, :base) : nil,
 } do %>
   <% unless field_has_errors(@document, :base) %>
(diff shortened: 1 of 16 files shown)


lauraghiorghisor-tw:
FYI @ChrisBAshton - curious to know what you'd make of this, I know you have strong opinions on not customizing anything.

spec/components/error_summary_component_spec.rb
```diff
@@ -89,40 +89,6 @@
     expect(page).to have_css(".gem-c-error-summary__list-item span", text: "This is a top level error that is agnostic of model level validations. It has probably been added by an updater service or a controller and does not link to an input.")
   end
 
-  it "renders errors when 'ActiveModel::Errors' are passed in" do
```
ChrisBAshton, line 92: ⭐ great

app/views/metadata_fields/_licence_transactions.html.erb
```diff
@@ -13,7 +13,7 @@
   heading_level: 2,
   heading_size: "l",
   id: "#{format_name}_link_and_identifier_exists",
-  error_message: errors_for_input(@document.errors, :base)&.include?(t("activemodel.errors.models.licence_transaction.attributes.base.link_and_identifier_exists")) ? errors_for_input(@document.errors, :base) : nil,
+  error_message: errors_for_input(format_name, @document.errors, :base)&.include?(t("activemodel.errors.models.licence_transaction.attributes.base.link_and_identifier_exists")) ? errors_for_input(format_name, @document.errors, :base) : nil,
```
ChrisBAshton, line 16: The one 'smell' in this PR is this line - perhaps because I'm misunderstanding, or because I'm extending scope.

My rough understanding of this commit is that you've built an abstraction that takes any document type and spits out any related custom errors. Therefore I wouldn't expect to see any more hardcoded references to `activemodel.errors.models.licence_transaction`? We still have the hardcoded view here, obviously, but can we more fully delegate to the ErrorSummaryComponent?

Also might be worth linking to the [existing errors for license transactions in en.yml](https://github.com/alphagov/specialist-publisher/blob/3f2544fdec98e2031e628a0f98b61b11fd9ec4c6/config/locales/en.yml#L50-L79) in the commit message, as I half-expected to see some custom YML added to support this feature and was surprised I couldn't find any.

ChrisBAshton (approved):
Looks like an improvement to me 👍

app/views/metadata_fields/_licence_transactions.html.erb
lauraghiorghisor-tw: Ah, that line of code... Sort of unrelated I'd say - it's just reading the string from the yml to check that the relevant error is on :base and not some other error (previously we were rendering THIS error for ANY :base errors 😰). 

So technically the bits that executes as part of the `errors_for_input` is using my abstraction correctly, just deciding whether or not to execute it here is a bit verbose.

lauraghiorghisor-tw: Will add the link to the commit 👍🏻

PR #3253, opened 2025-07-10
[WHIT-2297] [Design System] Design changes for DS release

matthillco:
On the back of testing for the design system release:

- Add classes to error page to match DS
- Documents index page: make status bold (700) + also update it on the summary page
- Summary page: Change the body font size / line height - it is unreadable at the moment
- Summary page: Publication state - badge - should be longer + ensure it works ok on the document index page after the change
- Update preview border to match border of other elements (both add new and edit pages)

## Screenshots

| Before | After |
| --- | --- |
| Error page | |
| Screenshot 2025-07-15 at 13 54 12 | Screenshot 2025-07-15 at 13 54 23 |
| Index page status badge | |
| Screenshot 2025-07-15 at 13 53 09 | Screenshot 2025-07-15 at 13 53 14 |
| Summary page status badge | |
| Screenshot 2025-07-15 at 13 53 40 | Screenshot 2025-07-15 at 13 53 48 |
| Summary page body - font size and padding | |
| Screenshot 2025-07-15 at 13 54 39 | Screenshot 2025-07-15 at 14 09 08 |
| Body preview border - should be a (one) pixel difference 🕵🏻‍♀️|
| Screenshot 2025-07-15 at 14 19 05 | Screenshot 2025-07-15 at 14 19 42 |

### DIFF
--- a/app/assets/stylesheets/_summary.scss
+++ b/app/assets/stylesheets/_summary.scss
@@ -1,5 +1,11 @@
 .app-view-summary__section {
   padding: govuk-spacing(3) 0;
+
+  .govspeak-help__pre {
+    padding: govuk-spacing(4);
+    @include govuk-font($size: 19);
+  }
+
 }
 
 .app-view-summary__table-link {
--- a/app/assets/stylesheets/admin/_overrides.scss
+++ b/app/assets/stylesheets/admin/_overrides.scss
@@ -1,0 +1,7 @@
+.govuk-tag--s {
+  @include govuk-font(16, $weight: bold, $line-height: 1);
+}
+
+.govuk-summary-list__value .govuk-tag {
+  max-width: none;
+}
--- a/app/assets/stylesheets/components/_govspeak-editor.scss
+++ b/app/assets/stylesheets/components/_govspeak-editor.scss
@@ -24,7 +24,7 @@
 }
 
 .app-c-govspeak-editor__preview {
-  border: 1px solid $govuk-border-colour;
+  border: $govuk-border-width-form-element solid $govuk-border-colour;
   padding: govuk-spacing(4);
 }
 
--- a/app/views/passthrough/error.html.erb
+++ b/app/views/passthrough/error.html.erb
@@ -1,2 +1,9 @@
-<h1>Sorry, you don't have permission to access this application</h1>
-<p>Please contact your main GDS contact if you need access.</p>
+<%= render "govuk_publishing_components/components/heading", {
+  text: "Permission required",
+  heading_level: 1,
+	font_size: "xl",
+	margin_bottom: 6,
+} %>
+
+<p class="govuk-body">Sorry, you don't have permission to access this application.</p>
+<p class="govuk-body">Please contact your main GDS contact if you need access.</p>
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -38,6 +38,8 @@ $govuk-page-width: 1140px;
 @import "./govspeak-help";
 @import "./summary";
 
+@import "./admin/overrides";
+
 .bg-light-grey {
   background-color: govuk-colour("light-grey");
 }
--- a/app/helpers/state_helper.rb
+++ b/app/helpers/state_helper.rb
@@ -6,9 +6,9 @@ module StateHelper
   end
 
   def design_system_classes_for_frontend(document)
-    return "govuk-tag govuk-tag--blue" if state_for_frontend(document) =~ /draft/
+    return "govuk-tag govuk-tag--s govuk-tag--blue" if state_for_frontend(document) =~ /draft/
 
-    "govuk-tag govuk-tag--green"
+    "govuk-tag govuk-tag--s govuk-tag--green"
   end
 
   def state_for_frontend(document)


app/assets/stylesheets/admin/_overrides.scss
```diff
@@ -0,0 +1,7 @@
+.govuk-tag--s {
```
lauraghiorghisor-tw, line 1: Not sure I understand what needs to happen in wh - if there is a change, can it be made and linked to?

app/assets/stylesheets/_summary.scss
```diff
@@ -1,5 +1,10 @@
 .app-view-summary__section {
   padding: govuk-spacing(3) 0;
+
+  .govspeak-help__pre {
+    @include govuk-font($size: 19);
```
lauraghiorghisor-tw, line 6: What a joy to look at it now!

Can we also have a bit more padding here? For example `.app-c-govspeak-editor__preview` has a 20px (4 spacing?) padding which looks nice.

app/assets/stylesheets/components/_govspeak-editor.scss
```diff
@@ -24,8 +24,8 @@
 }
 
 .app-c-govspeak-editor__preview {
-  border: 1px solid $govuk-border-colour;
   padding: govuk-spacing(4);
+  border: $govuk-border-width-form-element solid $govuk-border-colour;
```
lauraghiorghisor-tw, line 27: Width is now 2 which matches the rest but the colour comes out as `#b1b4b6` whereas other elements are `#0b0c0c`, still looking odd.

app/views/passthrough/error.html.erb
```diff
@@ -1,2 +1,8 @@
-<h1>Sorry, you don't have permission to access this application</h1>
-<p>Please contact your main GDS contact if you need access.</p>
+<%= render "govuk_publishing_components/components/heading", {
```
lauraghiorghisor-tw, line 1: Nice!

I don't know what the rules are on heading level 1 elements, but at "xl" size, the title wraps (unnecessarily), which it wouldn't do if we got the font down to "l". What do you think?

lauraghiorghisor-tw (changes requested):
Few bits to adjust - looking good 😌

Can we also have some screenshots in the PR summary please?

app/assets/stylesheets/admin/_overrides.scss
matthillco: I'll need to raise a Whitehall PR to update the minimum text size over there too. I've noted it as a separate piece of work and I'll sort it out after completing these ones.

app/assets/stylesheets/components/_govspeak-editor.scss
matthillco: So the reason I did this is to distinguish that it's a preview and not a form field. Setting it to black made it look like all the other interactive form fields, which is a bit misleading as it's a static preview at this point.I felt that the grey border helps show the difference. 

Personally I'm not sure if that's even enough of a distinction and wonder if it should have some extra styling for the preview...?

app/views/passthrough/error.html.erb
matthillco: H1s normally are rendered in XL, and for consistency I'd like to keep it the same as all the other H1s.

I think in this case it's actually quite odd that the H1 is so long. It would be better to make this a much shorter information title, and then put the full message in the body copy below, eg:

```
<h1>Permissions error</h1>
<p>Sorry, you don't have permission to access this application</p>
<p>Please contact your main GDS contact if you need access.</p>
```

However "Permissions error" as the title doesn't seem quite informative enough. Do you have thoughts on what the title could be, or is it one for the team to discuss?

app/assets/stylesheets/_summary.scss
matthillco: I can improve the padding but wonder if we need to do more.

Perhaps both this summary box and the preview box on the new/edit screen should have the same styling?

I'm not sure if that would be confusing, since the preview on the edit screen actually renders the markdown as HTML, while here it's just shown as raw markdown.

lauraghiorghisor-tw: I don't know that the consistency needs to be codified; I just think the padding on preview is a good amount of padding 😅

app/assets/stylesheets/components/_govspeak-editor.scss
lauraghiorghisor-tw: Oh, I see. I did not realize this distinction. In that case, perhaps keep it consistent with wh which is in fact 1px solid #b1b4b6. So revert this?

app/views/passthrough/error.html.erb
lauraghiorghisor-tw: As discussed with team - going for “Permission required”.


Wondering if we also need to adapt the url to match (currently going to /error).

matthillco: I've updated the H1 text. Happy to adjust the route, is there a convention for a 'permission required' route?

app/assets/stylesheets/components/_govspeak-editor.scss
matthillco: I think Keeping the 2px makes sense, it looks better with all the others -- but how about dashed instead of solid?

image

app/assets/stylesheets/_summary.scss
matthillco: OK this now has 20px padding

image

app/assets/stylesheets/components/_govspeak-editor.scss
lauraghiorghisor-tw: NO 😅

app/views/passthrough/error.html.erb
lauraghiorghisor-tw: Think we're bike shedding. Happy as is.

app/assets/stylesheets/components/_govspeak-editor.scss
matthillco: Harumph. OK, I've reverted it to a solid line, but kept it 2px.

lauraghiorghisor-tw (approved):

PR #3256, opened 2025-07-11
[WHIT-2297] [Design System] Fix the broken JS for the sifting status toggle

matthillco:
## What
Update the broken JS for the _Sifting Status_ field. This field can be found within the document type _EU Withdrawal Act 2018 statutory instrument_.

## Why
The JS broke after porting this document over to the design system, due to the legacy JS using jQuery and being generally incompatible with our current standards.

### DIFF
--- a/app/views/metadata_fields/_statutory_instruments.html.erb
+++ b/app/views/metadata_fields/_statutory_instruments.html.erb
@@ -1,33 +1,20 @@
 <% finder_schema = f.object.class.finder_schema %>
 <% format_name = f.object.class.document_type.to_sym %>
+
 <!-- TODO: to be standardized - added manually to maintain custom labels -->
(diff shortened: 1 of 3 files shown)


app/views/metadata_fields/_statutory_instruments.html.erb
```diff
@@ -1,33 +1,20 @@
 <% finder_schema = f.object.class.finder_schema %>
 <% format_name = f.object.class.document_type.to_sym %>
 <!-- TODO: to be standardized - added manually to maintain custom labels -->
```
lauraghiorghisor-tw: The comment on line 3 needs to go after the new line not before - as it refers to lines 5 and after.

lauraghiorghisor-tw (approved):
LGTM ⭐

PR #3260, opened 2025-07-15
[WHIT-2027] [Design system] Release design system to users

lauraghiorghisor-tw:
Added flipflop to facilitate the release.
Currently maintaining both DS and legacy tests.

🚨The flipflop is enabled by default so only merge this when go-live is confirmed.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2027)


app/controllers/application_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationController < ActionController::Base
   before_action :authenticate_user!
   before_action :set_authenticated_user_header
   after_action :verify_authorized
+  skip_after_action :verify_authorized, if: -> { self.class.name.include?("Flipflop") }
```
lauraghiorghisor-tw, line 11: It would break without it - not something I'd expect to have to do, but alas.

config/application.rb
```diff
@@ -21,6 +21,14 @@
 
 module SpecialistPublisher
   class Application < Rails::Application
+    # Before filter for Flipflop dashboard. Replace with a lambda or method name
+    # defined in ApplicationController to implement access control.
+    config.flipflop.dashboard_access_filter = -> { head :forbidden unless Rails.env.development? || %w[integration staging].any? { |environment| ENV.fetch("GOVUK_WEBSITE_ROOT", "").include?(environment) } }
```
lauraghiorghisor-tw, line 26: Copied from WH. Whether or not this `GOVUK_WEBSITE_ROOT` actually exists in the environment for SP, I do not know, but I will test in integration and hope to be able to see the dashboard.

I sort of thought we'd offer the dashboard in prod as well - but realized wh does not do it, and perhaps that's sensible.

config/features.rb
```diff
@@ -0,0 +1,26 @@
+Flipflop.configure do
+  # Strategies will be used in the order listed here.
```
lauraghiorghisor-tw, line 2: So we can't do an active record strategy for this as it doesn't have a db - but that's not really needed. I think we can manage just fine with the default. Bit constrained as we cannot play with it too much across environments (i.e. when the PR is merged, it's ON) - but I can live with it.

config/features.rb
```diff
@@ -0,0 +1,26 @@
...
+  # end
+
+  # Declare your features, e.g:
+  #
+  # feature :world_domination,
+  #   default: true,
+  #   description: "Take over the world."
+  feature :show_design_system,
+          description: "Enable design system functionality",
+          default: true
```
lauraghiorghisor-tw, line 25: All tests actually also pass with it off too, so if it's wiser to merge this in as false first and then open one simple PR to go live (where we turn it on), that can work too.

I am happy I've proven that all tests pass with DS and legacy simultaneously already.

EDIT: On second thought I am unsure why the tests pass with it off, I actually think DS tests should all fail with it off 😅
EDIT2: Think it might be because I am actively turning it on after each legacy test.
EDIT3: Confirmed that that's the case - when running all tests in parallel they all pass, but when running DS tests individually, they would fail. That to say, perhaps wait on this to merge properly as a go-live, so we don't introduce flaky tests into the mix.

spec/support/authentication.rb
```diff
@@ -26,18 +20,8 @@ def log_in_as(user)
...
-  def log_in_as_design_system(user)
-    user.permissions << "preview_design_system"
-    GDS::SSO.test_user = user
-  end
-
-  def log_in_as_editor(user_type)
-    log_in_as FactoryBot.create(user_type)
-  end
-
-  def log_in_as_design_system_editor(user_type, organisation_content_id = nil)
```
lauraghiorghisor-tw, line 38: I kept the organisation_content_id param as is used in some non conforming doc tests. It is default so not causing code changes to keep.

spec/controllers/legacy/attachments_controller_spec.rb
lauraghiorghisor-tw, line 1: Bit inconsistent that for feature tests we have all grouped under DS folder whilst here I grouped the ones for deletion under legacy - either way this PR has 80 file changes - I'll untangle them in the next one because they also require file renames 😰

lauraghiorghisor-tw (commented):
The way the PR is structured (to aid review) means some of the commits are not atomic (tests fail for legacy until the last commit).

config/application.rb
ChrisBAshton: Hmmm I also thought the dashboard would be visible on production (for Whitehall too). Seems wer'e not making the most of Flipflop otherwise (i.e. we can't quickly disable a problematic feature - we'd have to instead change the config in features.rb and then deploy the code 🤔 

That said, Flipflop's README [suggests not enabling it on prod](https://github.com/voormedia/flipflop?tab=readme-ov-file#dashboard-access-control), so... 🤷 let's leave this I guess.

app/controllers/application_controller.rb
ChrisBAshton: This is to make `/flipflop` visible, right? We're not weakening the permissions model for accessing _other_ pages on SP?

config/features.rb
ChrisBAshton: SGTM

ChrisBAshton (approved):
Looks good 👍 I can see why it's a bit uncomfortable adding all that boilerplate for the legacy tests, and for it to feel a bit reductive as we're going to soon delete the legacy views anyway - but the approach makes sense. Happy to approve!

config/application.rb
lauraghiorghisor-tw: If it comes to that you can still deploy a quick default: false so it's still better.

lauraghiorghisor-tw:
@ChrisBAshton 
> This is to make /flipflop visible, right? We're not weakening the permissions model for accessing other pages on SP?

Yes. One would expect not a lot of things to be named Flipflop anyway... 
Unless you're trying to google the flipflop documentation... then it's a problem 😅

> That said, Flipflop's README [suggests not enabling it on prod](https://github.com/voormedia/flipflop?tab=readme-ov-file#dashboard-access-control), so... 🤷 let's leave this I guess.

They say it should not be "public" - I think as long as we showed it to authorized users it'd be ok.

PR #3270, opened 2025-07-18
[WHIT-2287] [Design system] Remove legacy tests

lauraghiorghisor-tw:
Remove legacy tests and reorganise/rename remaining tests. 

**Testing**: I have not tested this branch independently (manually) - as it only removes tests. It is nonetheless the base branch of `remove-design-system-feature-toggle` (https://github.com/alphagov/specialist-publisher/pull/3274) and that's been tested in integration.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2287)

PR #3274, opened 2025-07-21
[WHIT-2287] [Design system] Remove design system feature toggle

lauraghiorghisor-tw:
This PR cleans up all "functional" logic related to the design system toggle. 
- Remove all toggle logic from controllers and helpers.
- Remove any custom "split" logic that maintained both design and legacy behaviours.
- Remove the main legacy views associated with the controllers.

The tests have been removed previously in #3270.
Upcoming work will clean up remaining legacy code #3275.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2287)


spec/features/publishing_workflow/publishing_a_document_spec.rb
```diff
@@ -219,4 +219,26 @@
       end
     end
   end
+
+  context "publishing-api returns error" do
```
lauraghiorghisor-tw, line 223: Since we still have flash handling logic in the `design_system.html.erb` layout - I imagine we want to continue using flash messages for unknown errors (see extensive commit message).

Nonetheless, as mentioned in the commit - these flash messages are not very well covered in tests. But at least we now have this publish test, as well as the [unpublish test](https://github.com/alphagov/specialist-publisher/blob/885579d3f59a1f5044b9564540077d2ab2464ce9/spec/features/design_system/unpublishing_a_cma_case_design_system_spec.rb#L107). These should really be controller tests. But I did not have the time to set them up. 

The logic there is a bit muddled, and the mock helpers can be a bit annoying - so it's fair to say I gave up. For example, I was trying to mock an error from pub-api but it would still append it to the @document.errors (for update action) - causing it to go on the other branch of the if/else. Unsure if this is something that needs further digging.

patrickpatrickpatrick (approved):
This looks good to me!

app/controllers/passthrough_controller.rb
```diff
@@ -1,23 +1,8 @@
 class PassthroughController < ApplicationController
```
lauraghiorghisor-tw, line 1: I do wonder if we can get the same behaviour from doing something smart in the routes now - this is but a shell of a controller now. Out of scope, though.

GDSNewt (approved):
LGTM :shipit:

PR #3275, opened 2025-07-21
[WHIT-1909] [Design System] Clean up legacy code

lauraghiorghisor-tw:
### What

The Design System feature toggle has been removed in https://github.com/alphagov/specialist-publisher/pull/3274. This PR now cleans up remaining views, js, and css.

I've used this opportunity to scan through the code and remove anything dead-looking. 
Next: removal of `govuk_admin_template` in https://github.com/alphagov/specialist-publisher/pull/3276.

### Testing

Tested manually in integration, following the test plan we had for [release here](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2027).

The only oddity noticed in testing was this overflown attachment title: 
| Before | After |
| --- | --- |
| Screenshot 2025-07-22 at 13 14 09 | Screenshot 2025-07-22 at 13 13 53 |
- As we can see that was the case before too. 
- Now fixed in commit `ce905d95`:
Screenshot 2025-07-22 at 17 57 05

[Jira](https://gov-uk.atlassian.net/browse/WHIT-1909)


app/presenters/actions_presenter.rb
```diff
@@ -129,24 +71,16 @@ def discard_button_visible?
...
-
-  def discard_alert_legacy
-    "Are you sure you want to discard this draft?"
-  end
-
   def discard_path
     discard_document_path(slug, document.content_id_and_locale)
   end
 
+  def discard_draft_notice
```
lauraghiorghisor-tw, line 78: Just moved down in the "discard" section.

app/assets/javascripts/legacy-application.js
```diff
@@ -1,79 +0,0 @@
-//= require jquery
-//= require select2
```
lauraghiorghisor-tw, line 2: I am not 100% what this select2 thing was, but pretty sure we don't need it, nor the jquery anymore.

app/assets/stylesheets/components/_govspeak-help.scss
lauraghiorghisor-tw, line 1: This refers to the side-bar "help" - so still needed, and should be under components.

app/views/attachments/_attachment_links.html.erb
```diff
@@ -20,12 +20,12 @@
           <div class="attachment govuk-!-margin-top-6">
             <p class="govuk-body">
               <strong>Title: </strong>
-              <span class="title attachment-line"><%= attachment.title %></span>
+              <span class="title"><%= attachment.title %></span>
```
lauraghiorghisor-tw: My thinking here is, this class was not available anyway for the DS views. There is no difference as far as I can see in the legacy vs DS. Tough there is an overflowing attachment title (see screenshot in the PR summary.)

Gemfile
```diff
@@ -5,7 +5,6 @@ gem "rails", "~> 8.0.2"
 
 gem "aws-sdk-s3"
 gem "bootsnap", require: false
-gem "bootstrap-kaminari-views"
```
lauraghiorghisor-tw, line 8: I am pretty sure all kaminari things refer to pagination. No breaking changes as far as I could see in testing.

Gemfile
```diff
@@ -5,7 +5,6 @@ gem "rails", "~> 8.0.2"
...
 gem "dartsass-rails"
 gem "diffy"
 gem "flipflop"
```
lauraghiorghisor-tw: Kept flipflop for now - although it has no features atm. Could also remove.

lauraghiorghisor-tw:
@reviewer - Input on what dependencies and imports are needed, or rather, which others can be removed, would be welcome. I noticed no breaking changes in testing, so I hope I didn't nuke anything that we still need 😅 But always happy to delete some more stuff if you spot anything!

app/assets/javascripts/legacy-application.js
matthillco: `select2` was the old third party plugin for the bootstrap `select-with-search` and `autocomplete`. We can remove it.

lauraghiorghisor-tw: I figured 👍

app/assets/stylesheets/application.scss
```diff
@@ -36,10 +36,9 @@ $govuk-page-width: 1140px;
...
 @import "./components/diff";
-
-@import "./govspeak-help";
+@import "components/govspeak-help";
 @import "./summary";
-
+@import "./attachments";
 @import "./admin/overrides";
 
 .bg-light-grey {
```
lauraghiorghisor-tw, line 44: Should this tiny bit of style move into its own file? This is used for the select box background.

patrickpatrickpatrick: We're aiming to move this into the GOV.UK Publishing Components eventually, so I think it could stay here for now since it's temporary.

Gemfile
patrickpatrickpatrick: Would there be anything else to remove apart from removing it from the `Gemfile`? Just that we'll still get dependabot checking it if it's still installed even if we don't use it.

lauraghiorghisor-tw: I had some secret plans to use it to toggle "live" or "draft" target stack in integration - because I find it annoying that I can't test finders "as live" in integration 👹
But not used outside of that. 

Removing it is gemfile + `features.rb` config.

Should I remove? And then if and when I get around to implementing the integration thing, it might be that I bring it back or use another way.

patrickpatrickpatrick: Yeah, I think that would be the best approach. If you were to remove in it's own commit, then you could always use `git revert` when you get around to doing that.

lauraghiorghisor-tw: Done.

patrickpatrickpatrick (approved):

PR #3286, opened 2025-07-27
[WHIT-1460] Create Rake Task to Bulk Import Design Decisions From Whitehall 

TonyGDS:
## What

A rake task that reads from a CSV of exported Design Decisions (including attachments) from Whitehall and creates corresponding records for Specialist Publisher. It uses the `design_decisions.json` schema to map the Hearing Officers against the allowed values. Any rows that fail to import due to missing/unmapped values are outputted so that the csv can be checked.

## Why

[[WHIT-1460] IPO Design decisions Finder](https://gov-uk.atlassian.net/browse/WHIT-1460) Involved the creation of a new finder for Design Decisions. IPO asked if the existing records could be bulk imported rather than manually created, due to the amount of existing decisions. The script could be extended to support imports for other types in future.

[WHIT-1460]: https://gov-uk.atlassian.net/browse/WHIT-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Testing in integration 
- Ran the rake task in `dry-run` mode to identify any data anomalies that would prevent a row being imported - using the [csv](https://docs.google.com/spreadsheets/d/1P9Ycu9IYRI1Lio-lVNakIddnVdXSjy0VD7b1J7bGpv0/edit?gid=108004737#gid=108004737) extracted by [the export rake task](https://github.com/alphagov/whitehall/pull/10453)
- We identified some missing notes, missing dates (just missing from summary), some missing litigants and officers (due to bad table formatting), hearing officers that were not previously in the schema, hearing officers that do not match to existing officers (but are very likely them), BL numbers that do not conform to the format (o/O), and incorrectly named attachments. There was one document that did not get imported due to missing associations in WH.
```
[DRY RUN] Imported: 163 document(s)
Skipped: 4 row(s)
Fix up: 1 row(s)
Skipped row details:
- Line 3: Missing design_decision_litigants, design_decision_hearing_officer (Title: Design hearing decision: O/0636/25)
- Line 75: Missing design_decision_date (Title: Designs hearing decision: O/652/21)
- Line 89: Missing design_decision_hearing_officer (Title: Designs hearing decision: O/409/20)
- Line 161: Missing design_decision_date (Title: Designs hearing decision: O/288/14)

Fix up row details:
- Line 2: Fix up note (Title: Design hearing decision: O/0646/25)
```

- We manually fixed some litigants, dates and hearing officers in the csv - see [corrected csv](https://docs.google.com/spreadsheets/d/17EH3L5IWLXqrJZRbfnbe_k72GDE87te6H0hvTsUQy2c/edit?gid=108004737#gid=108004737)
- Added missing hearing officers to the schema
- Compose the mapping file from the output of the dry run
- Run the rake task with the  provisory hearing officer mappings, csv corrections and schema changes
```
[DRY RUN] Imported: 167 document(s)
Skipped: 0 row(s)
Fix up: 1 row(s)

Fix up row details:
- Line 2: Fix up note (Title: Design hearing decision: O/0646/25)
```
- Identified that the order in which decisions save needs to be reverted so that oldest ones are created first
- Added a sleep instruction to ensure the order is correct
- Final test run gave no missing fields, no skipped rows, just one remaining fix up note that can be manually fixed by user. One document needs to be created manually. All documents were created in the order of the csv:
image

## Prod run 
This is pending on https://github.com/alphagov/specialist-publisher/pull/3320. The user needs to confirm the hearing officers mapping file, as well as the schema changes in #3320.

The process for running the rake task:
- add csv file to pod `kubectl cp -n <namespace> <source> <pod-name>:<path>` - in the /tmp folder that we have write access to.
- add mappings file to pod `kubectl cp -n <namespace> <source> <pod-name>:<path>` - in the /tmp folder that we have write access to.
- run rake task -  dry-run, then actual run.
```
bundle exec rake bulk_import_documents_from_csv["/tmp/corrected_publications_export.csv","/tmp/mapping_file.csv",true]
bundle exec rake bulk_import_documents_from_csv["/tmp/corrected_publications_export.csv","/tmp/mapping_file.csv",false]
```

### DIFF
--- a/lib/tasks/bulk_import_documents_from_csv.rake
+++ b/lib/tasks/bulk_import_documents_from_csv.rake
@@ -9,8 +9,12 @@ task :bulk_import_documents_from_csv, %i[csv_file_path] => :environment do |_, a
   end
 
   hearing_officer_allowed_values = get_hearing_officer_allowed_values
+  imported_count = 0
+  invalid_rows = []
 
-  CSV.foreach(csv_file_path, headers: true) do |row|
+  documents_to_import = []
+
+  CSV.foreach(csv_file_path, headers: true).with_index(2) do |row, line_number|
     title = row["title"]
     summary = row["summary"]&.delete('"')
     body = row["body"]&.delete('"')
@@ -20,27 +24,50 @@ task :bulk_import_documents_from_csv, %i[csv_file_path] => :environment do |_, a
     design_decision_date = get_design_decision_date(summary)
     note_body = get_note_body(body)
 
-    design_decision = DesignDecision.new(title: title,
-                                         summary: summary,
-                                         design_decision_litigants: design_decision_litigants,
-                                         design_decision_hearing_officer: design_decision_hearing_officer,
-                                         design_decision_british_library_number: design_decision_british_library_number,
-                                         design_decision_date: design_decision_date,
-                                         body: note_body)
+    required_fields = {
+      "title" => title,
+      "summary" => summary,
+      "body" => note_body,
+      "design_decision_litigants" => design_decision_litigants,
+      "design_decision_hearing_officer" => design_decision_hearing_officer,
+      "design_decision_british_library_number" => design_decision_british_library_number,
+      "design_decision_date" => design_decision_date,
+    }
 
-    if design_decision_has_attachment?(row)
+    if required_fields.values.any?(&:blank?)
+      invalid_rows << {
+        line: line_number,
+        missing_fields: required_fields.select { |_, v| v.blank? }.keys,
+        title: title,
+      }
+    else
+      documents_to_import << { row: row, attributes: required_fields }
+    end
+  end
+
+  if invalid_rows.any?
+    raise StandardError, "CSV import failed: #{invalid_rows.count} row(s) have missing required fields."
+  end
+
+  documents_to_import.each do |document_data|
+    design_decision = DesignDecision.new(**document_data[:attributes].deep_symbolize_keys)
+
+    if design_decision_has_attachment?(document_data[:row])
       design_decision.attachments.build(
-        title: row["attachment_title"],
-        filename: row["attachment_filename"],
-        url: row["attachment_url"],
+        title: document_data[:row]["attachment_title"],
+        filename: document_data[:row]["attachment_filename"],
+        url: document_data[:row]["attachment_url"],
         content_type: "application/pdf",
-        created_at: row["attachment_created_at"].present? ? Time.zone.parse(row["attachment_created_at"]) : nil,
-        updated_at: row["attachment_updated_at"].present? ? Time.zone.parse(row["attachment_updated_at"]) : nil,
+        created_at: document_data[:row]["attachment_created_at"].present? ? Time.zone.parse(document_data[:row]["attachment_created_at"]) : nil,
+        updated_at: document_data[:row]["attachment_updated_at"].present? ? Time.zone.parse(document_data[:row]["attachment_updated_at"]) : nil,
       )
     end
 
     design_decision.save
+    imported_count += 1
   end
+
+  report_import_result(imported_count, invalid_rows)
 end
 
 def get_hearing_officer_allowed_values
@@ -86,4 +113,16 @@ end
 def get_design_decision_date(summary)
   raw_date = summary[/\b\d{1,2} \w+ \d{4}\b/] if summary
   raw_date ? Date.parse(raw_date).strftime("%Y-%m-%d") : nil
+end
+
+def report_import_result(imported_count, skipped_rows)
+  puts "Imported: #{imported_count} document(s)"
+  puts "Skipped: #{skipped_rows.count} row(s)"
+
+  if skipped_rows.any?
+    puts "\nSkipped row details:"
+    skipped_rows.each do |row|
+      puts "- Line #{row[:line]}: Missing #{row[:missing_fields].join(', ')} (Title: #{row[:title] || 'N/A'})"
+    end
+  end
 end
--- a/lib/tasks/bulk_import_documents_from_csv.rake
+++ b/lib/tasks/bulk_import_documents_from_csv.rake
@@ -13,6 +13,7 @@ task :bulk_import_documents_from_csv, %i[csv_file_path mapping_file_path dry_run
   hearing_officers = get_hearing_officers_from_schema.merge(get_hearing_officers_from_csv(mapping_file_path))
   imported_count = 0
   invalid_rows = []
+  fix_up_rows = []
 
   documents_to_import = []
 
@@ -29,7 +30,7 @@ task :bulk_import_documents_from_csv, %i[csv_file_path mapping_file_path dry_run
     required_fields = {
       "title" => title,
       "summary" => summary,
-      "body" => note_body,
+      "body" => note_body || "Missing note body",
       "design_decision_litigants" => design_decision_litigants,
       "design_decision_hearing_officer" => design_decision_hearing_officer,
       "design_decision_british_library_number" => design_decision_british_library_number,
@@ -43,6 +44,23 @@ task :bulk_import_documents_from_csv, %i[csv_file_path mapping_file_path dry_run
         title: title,
       }
     else
+      fixup_fields = []
+      unless design_decision_has_attachment?(row)
+        fixup_fields << "attachment"
+      end
+
+      unless note_body
+        fixup_fields << "note"
+      end
+
+      if fixup_fields.any?
+        fix_up_rows << {
+          line: line_number,
+          fixup_fields: fixup_fields,
+          title: title,
+        }
+      end
+
       documents_to_import << { row: row, attributes: required_fields }
     end
   end
@@ -69,7 +87,7 @@ task :bulk_import_documents_from_csv, %i[csv_file_path mapping_file_path dry_run
     imported_count += 1
   end
 
-  report_import_result(imported_count, invalid_rows, dry_run)
+  report_import_result(imported_count, invalid_rows, fix_up_rows, dry_run)
 end
 
 def get_hearing_officers_from_schema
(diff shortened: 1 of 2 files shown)


spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -1,10 +1,74 @@
...
+  end
+
+  # it "it exits if csv file not found" do
+  #   expect { Rake.application.invoke_task "bulk_import_documents_from_csv[non_existent_csv.csv]" }.to output("CSV file not found").to_stdout
+  # end
+  #
+  # it "handles an empty CSV file gracefully" do
+  #   CSV.open(csv_path, 'w') {} # Create an empty CSV file
+  #   expect { Rake.application.invoke_task "bulk_import_documents_from_csv[#{csv_path.to_s}]" }.to output("No data to import\n").to_stdout
+  # end
```
ChrisBAshton: I'd omit this commented out code until it's ready.

EDIT: you remove it altogether in commit 3. So yeah, let's not commit it in the first place.

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -7,8 +7,13 @@ task :bulk_import_documents_from_csv, %i[csv_file_path] => :environment do |_, a
   end
 
   CSV.foreach(csv_file_path, headers: true) do |row|
+    body = row["body"]&.delete('"')
+    design_decision_litigants = body[/\|\s*.*?litigants.*?\s*\|\s*(.+?)\s*\|/i, 1]
+    design_decision_hearing_officer = body[/\|\s*.*?hearing\s*officer.*?\s*\|\s*(.+?)\s*\|/i, 1]
```
ChrisBAshton: Test case for if there isn't a match?

spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -64,4 +60,6 @@
 $A')
     task.execute(csv_file_path: csv_path.to_s)
   end
+  # test for mapping Appointed Person to design_decision_hearing_officer
+  # test for managing different permutations of note
```
ChrisBAshton: These placeholder comments aren't particularly helpful at the review stage (which is why I double-checked this PR is ready for review, or whether you were still intending to tidy up the commit history beforehand). We tend to review commit by commit and are sticklers for [atomic commits as per the GDS Way](https://gds-way.digital.cabinet-office.gov.uk/standards/source-code/working-with-git.html#atomic-commits) 🙏

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -7,13 +7,18 @@ task :bulk_import_documents_from_csv, %i[csv_file_path] => :environment do |_, a
   end
 
   CSV.foreach(csv_file_path, headers: true) do |row|
+    title = row["title"]
+    design_decision_british_library_number = title[/Design hearing decision:\s*(\S+)/i, 1]
     body = row["body"]&.delete('"')
     design_decision_litigants = body[/\|\s*.*?litigants.*?\s*\|\s*(.+?)\s*\|/i, 1]
     design_decision_hearing_officer = body[/\|\s*.*?hearing\s*officer.*?\s*\|\s*(.+?)\s*\|/i, 1]
-    DesignDecision.new(title: row["title"],
+    start_index = body.lines.find_index { |line| line.match?(/^.*Every effort.*$/i) }
```
ChrisBAshton: This is starting to get unwieldy and would probably benefit from each field being a call-out to its own parser method.

EDIT: I see you refactor these in a commit at the end, but it's worth refactoring as you go along, i.e. at each commit, ask yourself if this was the final commit in the PR would you be happy or would you implement it differently?

spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -57,9 +58,10 @@
 Newport,
 South Wales
 NP10 8QQ
-$A')
+$A').and_return(double(save: true))
```
ChrisBAshton: Why this change? Unsure from the commit message. Probably wouldn't have noticed if this was squashed into one of the earlier commits.

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -1,3 +1,5 @@
+desc "Import design decisions from a CSV file and save them with optional attachments"
```
ChrisBAshton, line 1: Querying the commit message here: `Skip rows in the csv where required fields are missing`

Can you elaborate? Wouldn't we want to raise an exception if a required field is missing?

Is it worth doing a full validation of all the CSV contents before attempting the migration, so that we don't have to deal with semi-successful imports?

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -1,13 +1,20 @@
...
+
   csv_file_path = args[:csv_file_path]
 
   unless File.exist?(csv_file_path)
     puts "CSV file not found"
     exit
   end
 
+  schema_path = Rails.root.join("lib/documents/schemas/design_decisions.json")
+  schema = JSON.parse(File.read(schema_path))
```
ChrisBAshton: We should use the `FinderSchema.load_from_schema(type)` method here rather than hard-code the filepath and parse the schema manually.

spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -39,6 +39,10 @@
   let(:task) { Rake::Task["bulk_import_documents_from_csv"] }
   let(:csv_path) { Rails.root.join("publications_export.csv") }
 
+  before(:each) do
+    allow($stdout).to receive(:write)
```
ChrisBAshton, line 51: Worth adding an inline comment saying why you're adding this (presumably to reduce noise in the output)

spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -291,38 +291,46 @@
...
 
   it "maps hearing officer to schema value and skips invalid officers" do
-    CSV.open(csv_path, 'w') do |csv|
+    CSV.open(csv_path, "w") do |csv|
       csv << %w[title summary body attachment_title attachment_filename attachment_url attachment_created_at attachment_updated_at]
 
       # Valid officer
       csv << [
-        'Design hearing decision: O/1111/25',
+        "Design hearing decision: O/1111/25",
```
ChrisBAshton: These lint fixes should be fixed up to the commit where the code was introduced

spec/lib/tasks/bulk_import_documents_from_csv_spec.rb
```diff
@@ -37,33 +37,23 @@
...
-        attachment_filename
-        attachment_url
-        attachment_created_at
-        attachment_updated_at
-      ]
-      csv << [
-        "Design hearing decision: O/0567/25",
-        '"Outcome of request to invalidate, hearing held on 24 June 2025."',
-        '"| **Litigants** | Caesar Commerce Ltd v Huizhou New Road Cosmetics Company Limited |
+    csv_data = [
```
ChrisBAshton: I've no objection to the `Stub CSV IO in tests` changes but feel this should be 'fixed up' so that right from the start it is the approach taken for CSV testing.

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -88,7 +88,7 @@ def get_hearing_officer_mapping
 end
 
 def get_design_decision_british_library_number(title)
-  title[/Design hearing decision:\s*(\S+)/i, 1] if title
+  title[/([O0]\/\d+\/\d+)/i, 1] if title
```
ChrisBAshton: Fix up to commit where this is introduced (and add appropriate test so that if we reverted this change, something fails)

ChrisBAshton (commented):
Hey - this is shaping up nicely. Quite a few comments about commit structure, and one fairly substantial question: is it worth doing a full validation of all the CSV contents _before_ attempting the migration, so that we don't have to deal with semi-successful imports? Are we _expecting_ to have dodgy output in the CSV from Whitehall or are we preparing for a use case that we're not actually expecting to happen in the real migration?

E.g. if Whitehall is referencing an officer that doesn't exist in Specialist Publisher, does that indicate a problem with Whitehall's data or is there some up front work needed in SP to expand its list of officer enums?

I'd lean towards an "up front validation of the CSV" and only attempting the import if all the data is correct. That should simplify the code a bit, and potentially bring the headache 'forward' rather than kicking the can down the road for the dev to look at later 😅 

Haven't given much thought to genericising this yet, but we know we're going to want to do a CSV import for another finder soon too, so worth considering what kind of abstraction we may want to put in front of the CSV stuff, if any. Happy to give that more of a think in a second review!

lib/tasks/bulk_import_documents_from_csv.rake
TonyGDS: Yes, I do follow TDD principles and refactor as I go but i'm experimenting with AI first with TDD so I've been leaving it a bit longer before the refactoring step and I want to make sure I make frequent commits, however with the focus on the clean commit history i'll keep this in mind.

EDIT: I have re-written the commit history to take this into account

lib/tasks/bulk_import_documents_from_csv.rake
TonyGDS: Oh yes, I must have missed that helper

lib/tasks/bulk_import_documents_from_csv.rake
TonyGDS: As discussed I have added the ability to do a dry run, we don't really know all the issues with the data until we try to import. If dry run is not selected the task will raise an error and not save any Design Decisions

lib/tasks/bulk_import_documents_from_csv.rake
TonyGDS: when there are no matches they will be reported instead

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -0,0 +1,132 @@
...
+  csv_file_path = args[:csv_file_path]
+  mapping_file_path = args[:mapping_file_path]
+  dry_run = args[:dry_run].to_s.downcase == "true"
+
+  unless File.exist?(csv_file_path)
+    puts "CSV file not found"
+    exit
+  end
+
+  hearing_officers = get_hearing_officers_from_schema.merge(get_hearing_officers_from_csv(mapping_file_path))
```
lauraghiorghisor-tw, line 13: I am pretty sure this won't work unless we actually add the new officers to the schema, which at least means we don't need this logic here 😅
*It will fail validation when attempting to save in pub-api.

lib/tasks/bulk_import_documents_from_csv.rake
```diff
@@ -0,0 +1,132 @@
...
+    title = row["title"]
+    summary = row["summary"]&.delete('"')
+    body = row["body"]&.delete('"')
+    design_decision_litigants = get_design_decision_litigants(body)
+    design_decision_hearing_officer = get_design_decision_hearing_officer(body, hearing_officers)
+    design_decision_british_library_number = get_design_decision_british_library_number(title)
+    design_decision_date = get_design_decision_date(summary)
+    note_body = get_note_body(body)
+
+    required_fields = {
```
lauraghiorghisor-tw, line 30: I am inclined to consider attachments as required - but perhaps save without them and report which items did not have an attachment, if any.

lauraghiorghisor-tw (approved):
LGTM

PR #3290, opened 2025-07-29
Change finder organisation form control from multiselect with search to select with search plus add another [WHIT-1975]

ryanb-gds:
We require the order of the user's selection to be respected in the schema changes sent to Zendesk, as the order controls the order in which the organisations are listed on the finder content item when it is rendered on GOV.UK.

The select with search component in multiselect mode does not respect the order because it uses the native browser select element to store the selected values. Most browsers simply submit the selections in the order they appear in the select options list.

By using the add another component, we can allow the user to control the order in which the organisations will appear on GOV.UK.

### Before

Screenshot 2025-08-04 at 11 26 50

###  After

Note that there is a defect in the "add another" button positioning, which is too close to the "delete" button. I will look at fixing this in govuk publishing components

Screenshot 2025-08-04 at 11 42 43

JIRA: https://gov-uk.atlassian.net/browse/WHIT-1975

### DIFF
--- a/app/controllers/finders_controller.rb
+++ b/app/controllers/finders_controller.rb
@@ -87,7 +87,7 @@ private
       :show_table_of_contents,
       :document_title,
       :document_noun,
-      organisations: [],
+      organisations: %i[id _destroy],
       related: [],
     )
   end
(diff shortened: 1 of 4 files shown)


lauraghiorghisor-tw (approved):
We've been quite consistent in this codebase with extracting more complex view chunks  into components. There's no reuse to justify it, but I do find that it makes it obvious how many ways of doing things we have, if nothing else. Maybe you'd consider adding this is a component wrapper.

I might misremember, and sorry I could not test this locally as I don't have the orgs set up anymore, but was there some issue with deleting the last item in the list or something? Did you happen to test that in integration? 

LGTM otherwise.

PR #3297, opened 2025-07-30
[WHIT-2341] Add explanation on reslug for the base_path:edit rake task

lauraghiorghisor-tw:
Related to this support ticket: [Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2341)

### DIFF
--- a/lib/tasks/base_path.rake
+++ b/lib/tasks/base_path.rake
@@ -1,4 +1,8 @@
 namespace :base_path do
+  # This is equivalent to other Publishing applications' "reslugging" tasks.
+  # Once run, it will create a new draft document with the new `base_path`. The published document remains at the old base_path at this point. Your edition is now in a "published with new draft" state.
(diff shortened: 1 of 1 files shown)


lib/tasks/base_path.rake
```diff
@@ -1,4 +1,8 @@
 namespace :base_path do
+  # This is technically a reslug task.
```
ryanb-gds:
```suggestion
  # This is equivalent to other Publishing applications' "reslugging" tasks.
```
Not sure "reslug" actually is an official term we use

ryanb-gds (approved):
One suggestion

lib/tasks/base_path.rake
lauraghiorghisor-tw: I like this.

PR #3324, opened 2025-08-13
[WHIT-2320] Changes to UKHSA Chemical Hazards finder

lauraghiorghisor-tw:
Changes to chemical hazards finder

- change name, summary, noun, document name
- change email subscription
- remove and add new product category facet values
- add a new "reviewed date" non-filterable facet
- make product category not required

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2320)

### DIFF
--- a/lib/documents/schemas/ukhsa_chemical_hazards.json
+++ b/lib/documents/schemas/ukhsa_chemical_hazards.json
@@ -3,10 +3,10 @@
   "content_id": "326a05d8-0bce-40a4-b3fc-7aab7ad204b7",
   "description": "Find guidance on chemicals, chemical products and their health risks.",
   "document_noun": "item",
-  "document_title": "UKHSA Chemical Hazard",
+  "document_title": "UKHSA Chemical Risk",
   "email_filter_options": {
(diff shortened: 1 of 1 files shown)


ryanb-gds (approved):
I think this is fine. I'm not 100% sure there isn't any risk to changing the email filter by value - might just be worth having a little think about that if we haven't already. Suspect it'll be fine though as I don't think email alert API will care about that value

lauraghiorghisor-tw:
> I think this is fine. I'm not 100% sure there isn't any risk to changing the email filter by value - might just be worth having a little think about that if we haven't already. Suspect it'll be fine though as I don't think email alert API will care about that value

Actually you do have a point - we do need to usually open a PR in email alert api when creating a new finder - for the `all_` option we must have added all the facets as valid tags. If we only need one, then no need to have all in there. But not sure the user will like this email option actually 😅 - as I think they wanted multiple facets, but we have that limitation that we can only configure one in this manner -, so keeping that as is for now.

PR #3335, opened 2025-08-15
[WHIT-2320] Add option to override target_stack for testing

lauraghiorghisor-tw:
Introduce a feature toggle option that overrides the schema value for `target_stack`. This is only available in integration and staging.

This is to facilitate "live" testing and to avoid accidental publishing of finders after testing.

[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2320)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem "aws-sdk-s3"
 gem "bootsnap", require: false
 gem "dartsass-rails"
 gem "diffy"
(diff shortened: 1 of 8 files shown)


lauraghiorghisor-tw:
Changed approach - done here https://github.com/alphagov/specialist-publisher/pull/3377.

PR #3377, opened 2025-09-17
Add integration target stack override

lauraghiorghisor-tw:
In order to facilitate testing, we introduce a target stack override which only takes effect in integration.

This is helpful to:
- Developers who typically want to deploy the config and test it end to end, with all the functionality present. This typically requires setting the "target_stack" to "live" for testing, then back to "draft" to merge the PR, which is prone to error.
- Developers who want to test the finders locally following instructions in https://docs.publishing.service.gov.uk/repos/govuk-docker/how-tos/finder-setup.html#content
- Departments who want to test the search functionality which is not quite working at the minute in prod (i.e. draft documents do not appear on the draft finder; published documents will, but that's typically undesirable; it is also quite confusing to have to explain users this rather intricate flow to users).

An alternative approach trying to use a flipflop toggle was attempted here https://github.com/alphagov/specialist-publisher/pull/3335, but that wasn't very successful as we still operate with rake task for publishing finders. Since SP does not have its own DB, it cannot store the value of the flipflop with the active_record strategy, so the rake task cannot tap into the value of the flag set in the dashboard. Current implementation is probably simpler than using the toggle, but the flag approach could be revived if we build finder publishing into the UI, as part of future work to make SP more self-serve. Also tried using the `redis` strategy, but not successful (out of the box). Could explore other gems, but the schema based will do for now.

### DIFF
--- a/app/lib/publishing_api_finder_publisher.rb
+++ b/app/lib/publishing_api_finder_publisher.rb
@@ -18,7 +18,13 @@ private
   attr_reader :finders, :logger
 
   def should_deploy_to_live?(finder)
-    finder[:file]["target_stack"] == "live"
+    target_stack = finder[:file]["target_stack"]
+
+    if %w[development integration].include?(GovukEnvironment.current) && (integration_target_stack = finder[:file]["integration_target_stack"])
+      target_stack = integration_target_stack
+    end
+
+    target_stack == "live"
   end
 
   def export_finder(finder)
@@ -31,8 +37,7 @@ private
       finder[:file],
     )
 
-    target_stack = finder[:file]["target_stack"]
-    logger.info("Publishing '#{finder[:file]['name']}' to #{target_stack} stack")
+    logger.info("Publishing '#{finder[:file]['name']}' to #{target_stack_name(finder)} stack")
 
     update_draft(finder_payload)
     publish(finder_payload, links_payload) if should_deploy_to_live?(finder)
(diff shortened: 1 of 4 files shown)


app/lib/publishing_api_finder_publisher.rb
```diff
@@ -18,7 +18,9 @@ def call
   attr_reader :finders, :logger
 
   def should_deploy_to_live?(finder)
-    finder[:file]["target_stack"] == "live"
+    return finder[:file]["target_stack"] == "live" unless %w[integration development].include?(GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment)
```
ChrisBAshton: Please use the `GovukEnvironment.current` method for this [as per the docs](https://docs.publishing.service.gov.uk/manual/environments.html#determining-your-environment).

app/lib/publishing_api_finder_publisher.rb
```diff
@@ -18,7 +18,9 @@ def call
...
+
+    (finder[:file]["integration_target_stack_override"] || finder[:file]["target_stack"]) == "live"
```
ChrisBAshton: Not loving the double `finder[:file]["target_stack"] == "live"` in this method. Could it be changed to...?

```ruby
  target_stack = finder[:file]["target_stack"]
  if (%w[local integration].include?(GovukEnvironment.current) && finder[:file]["integration_target_stack"])
    target_stack = finder[:file]["integration_target_stack"]
  end
  target_stack
```

I think we could also rename the var to `integration_target_stack` rather than `integration_target_stack_override`, as it's self explanatory that we're overriding integration's target stack when setting this 🤔

ChrisBAshton (approved):
LGTM, with some suggestions below.

app/lib/publishing_api_finder_publisher.rb
lauraghiorghisor-tw: Nice one!

app/lib/publishing_api_finder_publisher.rb
lauraghiorghisor-tw: It's still giving me "development" rather than "local" locally so used that instead ;)

PR #3434, opened 2025-11-07
[WHIT-2422] Add schema for digital market regime requirements

lauraghiorghisor-tw:
[Jira](https://gov-uk.atlassian.net/jira/software/c/projects/WHIT/boards/401?selectedIssue=WHIT-2422)

Screenshots from testing
Doc form
doc summary
Designation now single select
Company now single select
publisher index view
finder index view
filter 3
filter 1
filter 2
emails 1
emails 2

Also tested emails. Because we only set up the most generic option, you get an email for any sort of document, regardless of the facet.
tested email alert
test 1
test 2

### DIFF
--- a/spec/features/publishing_workflow/creating_a_new_document_spec.rb
+++ b/spec/features/publishing_workflow/creating_a_new_document_spec.rb
@@ -8,6 +8,7 @@ EXCEPTIONS_TO_GENERAL_TESTING = %w[
   statutory_instrument
   protected_food_drink_name
   trademark_decision
+  digital_markets_regime_requirement
 ].freeze
 
 RSpec.feature "Creating a new document", type: :feature do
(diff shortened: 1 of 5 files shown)


spec/features/publishing_workflow/creating_a_new_document_spec.rb
```diff
@@ -8,6 +8,7 @@
   statutory_instrument
   protected_food_drink_name
   trademark_decision
+  digital_markets_regime_requirement
```
lauraghiorghisor-tw: Oh missed this, no longer needing to exclude as we are not using nested facets!

GDSNewt (approved):
Looks good, I assume the breaking CI is because we haven't added a valid schema to pub api?

PR #3495, opened 2026-01-09
[WHIT-2899] Update date validation to match Publishing API

TonyGDS:
This fixes an issue where a user had input a year (20206) that was outside the range specified by Publishing API (^[1-9][0-9]{3}-(0[1-9]|1[0-2])-(0[1-9]|[12][0-9]|3[0-1])$ === 1000-01-01 to 9999-12-31). The validator has been updated to check the range for the year. Also the parse operation has been removed because the `strptime` function we use already performs stricter validation.

The unit tests were also passing in the dates in the wrong format. The tests still passed by coincidence rather than intention, so these have also been fixed.

[Jira](https://gov-uk.atlassian.net/browse/WHIT-2899)

### DIFF
--- a/app/validators/date_validator.rb
+++ b/app/validators/date_validator.rb
@@ -2,8 +2,10 @@ class DateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.blank?
 
-    Date.parse(value)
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
⭐

PR #3589, opened 2026-03-25
Allow specialist documents to opt out of search engines indexing [WHIT-3180]

kashifatcha:
We're now supporting the ability to add a `noindex` meta tag to specialist document pages.
This is to prevent search engines from indexing certain specialist documents, such as tribunal decisions.

Message from support request:
> HMCTS publishes tribunal decision documents to GOV.UK  which contain the names of individuals involved in tribunal cases – for example, [residential property tribunal decisions](https://www.gov.uk/residential-property-tribunal-decisions) and [employment tribunal decisions](https://www.gov.uk/employment-tribunal-decisions).
These documents are publicly available – therefore, search engines (e.g. Google) index them and include them in search results. They are likely to feature prominently in search results due to the authority of the GOV.UK  website.

#### New Finder Form 
image

#### Finder Summary Page
image

#### Edit Finder Confirmation Page
image

Related PRs:
- https://github.com/alphagov/publishing-api/pull/3979 
- https://github.com/alphagov/publishing-api/pull/3987
- https://github.com/alphagov/frontend/pull/5442

---

Jira: https://gov-uk.atlassian.net/browse/WHIT-3180


spec/features/finders_admin/generating_a_new_finder_spec.rb
```diff
@@ -61,6 +61,18 @@
     expect(page).to have_checked_field("show_table_of_contents", with: "true")
   end
 
+  scenario "setting index_documents_in_search_engines value and generating a schema" do
+    visit new_finder_path
+
+    fill_in("Title of the finder", with: "Finder with contents list")
+    choose("No", name: "index_documents_in_search_engines")
```
lauraghiorghisor-tw: Just because the default is false, can we set "yes" here to ensure we can override the default and that the test is not passing due to the default?

app/presenters/finder_content_item_presenter.rb
```diff
@@ -51,6 +51,7 @@ def details
       show_summaries: file.fetch("show_summaries", false),
       show_metadata_block: file.fetch("show_metadata_block", false),
       show_table_of_contents: file.fetch("show_table_of_contents", false),
+      index_documents_in_search_engines: file.fetch("index_documents_in_search_engines", true),
```
lauraghiorghisor-tw: Inconsistent with the other settings, and its own default in the FinderSchema. 

The settings above, for example, default to `false`, though those settings are usually true except for a few finders (specifically `show_metadata_blocks`, `show_table_of_contents`). I suppose it would reads as "if it hasn't been set, consider it not true"? And then we must actively set them to true in the jsons.

In the attributes definition in the finder schema you made this fall back to false (i.e. not set = false), which should be consistent with the presenter behaviour, otherwise you end up sending `true` downstream for something intended to be "false" by omission from schema.

app/views/finders/_metadata_form_fields.html.erb
```diff
@@ -146,6 +146,26 @@
   ]
 } %>
 
+<%= render "govuk_publishing_components/components/radio", {
+  heading: "Should the documents be indexed by search engines?",
+  heading_size: "s",
+  hint: "Allow this finder's documents to be indexed by search engines such as Google. Finders like Residential Property Tribunal Decision that could contain personally identifiable information are set to 'No' to prevent search engines from indexing the documents.",
```
lauraghiorghisor-tw: Suggest changing to: "Allow this finder's documents to be indexed by search engines such as Google. Usually set to true. Disabled for documents such as 'Residential Property Tribunal Decision' that could contain personally identifiable information."

spec/features/finders_admin/editing_a_finder_spec.rb
```diff
@@ -39,6 +39,7 @@
     fill_in "document_title", with: "Changed document title"
     fill_in "document_noun", with: "Changed document noun"
     choose "email_alert_type", option: "no"
+    choose("No", name: "index_documents_in_search_engines")
```
lauraghiorghisor-tw, line 42: Just because the default is false, can we set "yes" here to ensure we can override the default and that the test is not passing due to the default? 😅

lauraghiorghisor-tw (changes requested):
LGTM but change the presenter default to false.

You will likely need to republish documents for the link expansion to work, but am not sure, might be the that republishing the finder suffices.

app/models/finder_schema.rb
```diff
@@ -38,6 +38,7 @@ def self.document_models
   attribute :email_filter_options
   attribute :facets, default: []
   attribute :filter
+  attribute :index_documents_in_search_engines, :boolean, default: false
```
lauraghiorghisor-tw, line 41: I had a think about this, as it really bugs me that the default is false for these settings that are _usually_ true. The `FinderSchema` is mainly used in the admin view we were looking at earlier today.  On the `finders/new` form it would be nice if this were preselected as "Yes", but then for the edit view, if the option is not set in the schema, it would be rendered as true which is incorrect 😅

Probably best left as is, though these options must tediously be set to true in each json (which is nice as it is expressive, but annoying that it is a MUST), and it is misleading that the default option is "false" when creating a new finder.

Alternatively you can amend the logic in the `new` action in the `FindersController` to set some of these to true:
```
[...]
    @proposed_schema.facets = []
    @proposed_schema.index_documents_in_search_engines ||= true
    @proposed_schema.show_metadata_block = true
    @proposed_schema.show_table_of_contents = true
```
and that would ensure that (only) when creating a new finder the "favoured" options are preselected.

lib/documents/schemas/employment_tribunal_decisions.json
```diff
@@ -315,6 +315,7 @@
   "phase": "beta",
   "show_table_of_contents": true,
   "show_metadata_block": true,
+  "index_documents_in_search_engines": false,
```
lauraghiorghisor-tw: I'd extract out the two false settings into their own commit to make it clear that they are excluded, and clearly reference the two relevant finders in the commit message and PR description. Also, remove the "." at the end of your commit message.

app/presenters/finder_content_item_presenter.rb
eYinka: Good spot! That wasn't the original intention. I've corrected it.

app/models/finder_schema.rb
eYinka: Nice idea! I think I'll defer it to another PR though

spec/features/finders_admin/editing_a_finder_spec.rb
```diff
@@ -39,7 +39,7 @@
     fill_in "document_title", with: "Changed document title"
     fill_in "document_noun", with: "Changed document noun"
     choose "email_alert_type", option: "no"
-    choose("Yes", name: "index_documents_in_search_engines")
+    choose("No", name: "index_documents_in_search_engines")
```
lauraghiorghisor-tw, line 42: This is being put back to No in the second commit.

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -24,6 +24,7 @@ See [example PR here](https://github.com/alphagov/publishing-api/pull/3026/files
 1. Add the [document type](./naming-conventions.md#publishing-api) to [allowed document types list](https://github.com/alphagov/publishing-api/blob/main/content_schemas/allowed_document_types.yml).
 2. Add any new field definitions to [this file](https://github.com/alphagov/publishing-api/blob/main/content_schemas/formats/shared/definitions/_specialist_document.jsonnet).
 3. Run `bundle exec rake build_schemas` to regenerate schemas.
+4. If you are adding new properties to the finder's `links` section, you may need to update the [expansion rules](https://github.com/alphagov/publishing-api/blob/main/lib/expansion_rules.rb) — see [Updating expansion rules for new finder-level properties](#updating-expansion-rules-for-new-finder-level-properties).
```
lauraghiorghisor-tw: I'd add this as a new paragraph under the numbered list (`Note: If you are blah...`), rather than numbering the entry (4). Or maybe add an `4. [OPTIONAL] If...` tag to it so people don't go wondering if they have to always do this step (I am people, I get confused by the steps I wrote few months ago 😅).

And then maybe change to:
"If you need to surface finder level settings to individual documents, you can do so by [updating the expansion rules](#updating-expansion-rules-for-new-finder-level-properties)."

docs/creating-editing-and-removing-specialist-document-types-and-finders.md
```diff
@@ -263,6 +264,12 @@ Specific values for fields of type array are now defined only in the `specialist
 
 The [schema](https://github.com/alphagov/specialist-publisher/tree/main/lib/documents/schemas) files that define a specialist document, are also used to configure that document's specialist finder. See [this](https://github.com/alphagov/specialist-publisher/pull/1899/commits/925abc689119138a0e04e17d3610f8ae276773dd) commit for an example.
 
+## Updating expansion rules for new finder-level properties
+
+If you add a new property to the finder's `links` section (i.e., a property that needs to be available to frontend apps through the Publishing API content store), you may also need to update the [expansion rules](https://github.com/alphagov/publishing-api/blob/main/lib/expansion_rules.rb) in `publishing-api`. Specifically, you'll need to add the new property to the `FINDER_FIELDS` constant, which controls which detail fields of a finder content item are included when the finder is expanded as a linked item.
```
lauraghiorghisor-tw: Let's call the section "Updating expansion rules for new finder-level **fields**" because that is what we seem to be calling things.

You don't really add the field to the links section, publishing API does it for you. Also, this is not really an "exceptional" behaviour, it is THE behaviour, as even "facets" are available on the document in this manner in order to render the blue box at the top (whether or not that is a great way to model this, well, that's a question for sure 😅). Also no need to be so detailed about the PR and tests. So let's rephrase to smth like:


> The [frontend](<link to fe>) application rendering a specialist document also needs to render finder-level information, for example in the metadata section. We typically provide this via the expanded `finder` link in the document content item. To instruct `publishing-api` to expand a field, update the [expansion rules](https://github.com/alphagov/publishing-api/blob/main/lib/expansion_rules.rb). Specifically, you'll need to add the new field to the `FINDER_FIELDS` constant, which controls which fields are included when the finder is expanded as a linked item.
> 
> This is a common pattern for the `facets` themselves, as well as additional fields such as `show_metadata_block`, `show_table_of_contents` and `index_documents_in_search_engines`. See [example PR](https://github.com/alphagov/publishing-api/pull/3987).
> 
> In the current implementation, the schema jsons in the specialist publisher repo define behaviours for both the finders and their documents, but the specialist document schemas in Publishing API only allow an additional `metadata` field based on the `facets`.  Making use of the link expansion feature to surface finder data in the document content item, keeps the individual document schemas minimal in Publishing API, while ensuring all documents "inherit" the behaviour.

lauraghiorghisor-tw (approved):

spec/features/finders_admin/editing_a_finder_spec.rb
eYinka: Yes because the behaviour is different in the second commit and the test will fail.

In the initial commit, `index_documents_in_search_engines` is not set on `cma_cases.json` (the schema being used in the test), which implicitly means the summary diff won't pick it up if the value has not changed.
In the newer commit, `index_documents_in_search_engines` has been added to `cma_cases` schema and set to true, so we need to flip the test in order to simulate finding the change on the summary diff.

PR #3592, opened 2026-03-26
Remove licence link or identifier validation [WHIT-3275]

ryanb-gds:
There are licence types that do not require either a link to apply for the licence or a licence identifier, so we can remove this validation.

[JIRA](https://gov-uk.atlassian.net/browse/WHIT-3275)


lauraghiorghisor-tw (approved):
LGTM

PR #3733, opened 2026-06-19
[WHIT-3667] Change veterans finder base_path and title

TonyGDS:
## What

- Move the veterans support finder from `/support-for-veterans` to `/veteran-support-organisations`.
- Change the finder title from "Find support for veterans and their families" to "Find organisations that support veterans and their families".
- Add an optional `publish` argument to the `base_path:edit` rake task so a reslug can be completed (draft + publish) in one step.

## Why

[Jira](https://gov-uk.atlassian.net/browse/WHIT-3667)

## Deploy / post-merge steps

```
publishing_api:publish_finder[veterans_support_organisations]
```

republish/represent each finder document

```
require "rake"
Rails.application.load_tasks
task = Rake::Task["base_path:edit"]

OLD = "/support-for-veterans"
NEW = "/veteran-support-organisations"

published = []; drafted = []; skipped = []

VeteransSupportOrganisation.find_each do |summary|
  content_id = summary.content_id
  locale     = summary.locale

  item      = Services.publishing_api.get_content(content_id, locale:).to_h
  base_path = item["base_path"]
  ordered   = item["state_history"].sort_by { |k, _| Integer(k) }.map(&:last)
  latest    = ordered.last
  previous  = ordered[-2]
  label     = (previous && latest == "draft") ? "#{previous} with new draft" : latest

  unless base_path&.start_with?("#{OLD}/")
    skipped << "#{base_path} (#{label})"
    next
  end

  new_base_path = base_path.sub(OLD, NEW)
  task.reenable

  if latest == "published"
    task.invoke(content_id, locale, new_base_path, "true")
    published << new_base_path
  elsif latest == "draft" && previous != "unpublished"
    task.invoke(content_id, locale, new_base_path)
    drafted << new_base_path
  else
    skipped << "#{base_path} (#{label})"
  end
end

puts "Published #{published.size}:"; puts published
puts "Drafted #{drafted.size}:";     puts drafted
puts "Skipped #{skipped.size}:";     puts skipped
```

### DIFF
--- a/lib/tasks/base_path.rake
+++ b/lib/tasks/base_path.rake
@@ -2,16 +2,22 @@ namespace :base_path do
   # This is equivalent to other Publishing applications' "reslugging" tasks.
   # Once run, it will create a new draft document with the new `base_path`. The published document remains at the old base_path at this point. Your edition is now in a "published with new draft" state.
   # To finish the reslug, you will need to publish the new draft document. This will cause the old document to redirect to the new base_path.
+  # Pass `true` as the final argument to publish the new draft automatically and complete the reslug in one step.
 
   desc "Edits the base_path of a document, given its content_id and the new base_path"
-  task :edit, %i[content_id locale new_base_path] => %i[environment] do |_t, args|
+  task :edit, %i[content_id locale new_base_path publish] => %i[environment] do |_t, args|
     document = Document.find(args[:content_id], args[:locale])
     old_base_path = document.base_path
 
     document.base_path = args[:new_base_path]
     document.update_type = "minor"
-    document.save
+    abort "Failed to save: #{document.errors.full_messages.join(', ')}" unless document.save
 
     puts "#{old_base_path} -> #{document.base_path}"
+
+    if args[:publish] == "true"
+      abort "Failed to publish: #{document.errors.full_messages.join(', ')}" unless document.publish
+      puts "Published #{document.base_path}"
+    end
   end
 end
--- a/spec/lib/tasks/base_path_spec.rb
+++ b/spec/lib/tasks/base_path_spec.rb
@@ -1,0 +1,85 @@
+require "rails_helper"
+
+RSpec.describe "base_path rake tasks", type: :task do
+  describe "base_path:edit" do
+    let(:output) { StringIO.new }
+    let(:task) { Rake::Task["base_path:edit"] }
+    let(:content_id) { SecureRandom.uuid }
+    let(:locale) { "en" }
+    let(:new_base_path) { "/new-finder/example-document" }
+
+    let(:document) do
+      instance_double(
+        Document,
+        base_path: "/old-finder/example-document",
+        :base_path= => nil,
+        :update_type= => nil,
+        save: true,
+        publish: true,
+      )
+    end
+
+    before do
+      $stdout = output
+      task.reenable
+    end
+
+    after { $stdout = STDOUT }
+
+    it "creates a draft at the new base_path" do
+      allow(Document).to receive(:find).with(content_id, locale).and_return(document)
+
+      task.invoke(content_id, locale, new_base_path)
+
+      expect(document).to have_received(:base_path=).with(new_base_path)
+      expect(document).to have_received(:update_type=).with("minor")
+      expect(document).to have_received(:save)
+    end
+
+    it "does not publish when no publish argument is given" do
+      allow(Document).to receive(:find).with(content_id, locale).and_return(document)
+
+      task.invoke(content_id, locale, new_base_path)
+
+      expect(document).not_to have_received(:publish)
+    end
+
+    it "publishes the new draft when publish is 'true'" do
+      allow(Document).to receive(:find).with(content_id, locale).and_return(document)
+
+      task.invoke(content_id, locale, new_base_path, "true")
+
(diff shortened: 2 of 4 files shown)


lauraghiorghisor-tw (changes requested):
I don't think you need to do the second commit. 

If republishing the finder does not republish all docs by default - which i half think it should - try to run  the existing republish. 
```
desc "republish many documents (space separated list of content_id:locale strings)"
  task :many, [:content_ids_and_locales] => :environment do |_, args|
    Republisher.republish_many(
      args.content_ids_and_locales.split(" ")
        .map { |id_and_locale| id_and_locale.split(":") },
    )
  end
```

As a rule, never attempt to re-think republish logic 😅 PERILS AWAIT YOU.

TonyGDS:
Unfortunately this won't work because of how we derive `base_path` in `Document`

```
  def base_path
    if first_draft?
      @base_path = "#{finder_schema.base_path}/#{title.to_url}"
    else
      @base_path
    end
    truncate(@base_path, length: 250, omission: "")
  end
```

It only derives a fresh path on first draft. So we need to run `base_path:edit` which sets the base path directly:

```
document.base_path = args[:new_base_path]
```

Just republishing would publish to the existing path.

As far as the associated documents go, when we change the finder base path, publishing-api calculates the redirects here: https://github.com/alphagov/publishing-api/blob/7e2bf9ef7e7721067a1fc2ef7c2b2ad8aa411c61/app/services/redirect_service.rb#L36-L45

```
  def redirects_for(routes, old_base_path, new_base_path)
    routes.map do |route|
      {
        path: route[:path],
        type: route[:type],
        destination: route[:path].gsub(old_base_path, new_base_path),
      }
    end
  end
```

the routes are for the finder (sent from FinderContentItemPresenter), which are these:

```
def routes
    [
      {
        path: base_path,
        type: "exact",
      },
      {
        path: "#{base_path}.json",
        type: "exact",
      },
      {
        path: "#{base_path}.atom",
        type: "exact",
      },
    ]
  end
```

Only these three routes are updated, nothing beyond the `base_path` i.e. `/support-for-veterans/<doc>`

lib/tasks/base_path.rake
```diff
@@ -13,5 +14,10 @@ namespace :base_path do
     document.save
 
     puts "#{old_base_path} -> #{document.base_path}"
+
+    if args[:publish] == "true"
+      document.publish
```
davidsauntson: Can this fail? If so do you need to do anything here?

davidsauntson (approved):
The changes make sense to me

I'm not sure what the ordering/previous/latest bits are about in the rake task in the PR message, but I'm assuming you've run it ok in integration.

lib/tasks/base_path.rake
TonyGDS: Yes, good pick up, the error will actually get swallowed. I'll add some handling here.

lauraghiorghisor-tw:
> Unfortunately this won't work because of how we derive `base_path` in `Document`

I can see the code, I can understand the point... and pub api may have changed in the meantime... But I swear I have made base path changes and the docs "just worked". 

Commit: `407df181` - just worked https://specialist-publisher.publishing.service.gov.uk/farming-grants

Have you tried it out - just republish a single doc with the finder changes deployed, and after having published the finder?

spec/lib/tasks/base_path_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    it "does not publish when no publish argument is given" do
+      allow(Document).to receive(:find).with(content_id, locale).and_return(document)
+
+      task.invoke(content_id, locale, new_base_path)
+
+      expect(document).not_to have_received(:publish)
+    end
+
+    it "publishes the new draft when publish is 'true'" do
```
lauraghiorghisor-tw, line 47: Can we add an expectation that the update type is minor - this feels dangerous but if minor and no change note I hope it's fine ⭐️

TonyGDS:
> Have you tried it out - just republish a single doc with the finder changes deployed, and after having published the finder?

yes, the republish does not update the base path:

1. Before:
```
> Services.publishing_api.get_content(<content_id>, locale: "en").to_h["base_path"]
=> "/support-for-veterans/op-nova"
```

2. Execute Rake task:
```
bundle exec rake 'republish:many[<content_id>:en]'
```

3. After:
```
> Services.publishing_api.get_content(<content_id>, locale: "en").to_h["base_path"]
=> "/support-for-veterans/op-nova"
```

lauraghiorghisor-tw (approved):
LGTM 

If this is the only way to reslug a finder and its docs, is the script not worth committing as a rake task, as well? Not too bothered about too much refactoring, just inline what you did in the edit rake task.

Slugs are so annoying. So that [base_path](https://github.com/alphagov/specialist-publisher/blob/7c05220879eb9aeb0f0d74fcd050a4b1be217aa0/app/models/document.rb#L98) method basically updates slug all the time when drafting (or state history one or shorter?????), otherwise pushes onto the same path? 🤔 I guess that's in line with slug behaviours elsewhere 😮‍💨

[Continues feeling gaslighted by SP, because I thought this worked in the past. Maybe we just happened to have short histories on documents we ran republish for, or maybe we just changed base path while the finder was still in draft.]

PR #3742, opened 2026-06-25
[WHIT-3667] Add task to reslug all documents of a type

TonyGDS:
## What

Adds a `base_path:edit_all[document_type]` rake task that reslugs every document of a given type so it sits under its finder's live `base_path` — for example after a finder's `base_path` changes.

[Jira](https://gov-uk.atlassian.net/browse/WHIT-3667)

## How it works

- Extracted into a `DocumentReslugger` service; the rake task owns the preconditions, the confirmation prompt, and printing the report.
- **Precondition:** the task reads the live finder and only proceeds when it's published **to the live stack** at the `base_path` in its schema. Otherwise it stops with a specific message (finder missing, only on the draft stack, or published at the wrong path). So deploy the schema change and publish the finder first.
- Prompts for `y/n` confirmation before running.
- Each document's slug is preserved: new path = `<finder base_path>/<basename of current path>`.
- Documents are handled by state:
  - **published** → reslugged and re-published (completing the redirect from the old path)
  - **draft / published-with-new-draft** → reslugged but kept in draft, so an in-progress edit isn't pushed live but will publish to the correct path later
  - **unpublished**, **brand-new first drafts**, or anything already under the finder's base_path → left untouched
- Failures don't halt the run: each is recorded as `content_id (locale) - error` and printed in the end-of-run summary alongside the published/drafted/skipped tallies.

Safe to re-run — already-moved documents are skipped.

## Usage

```
rake 'base_path:edit_all[veterans_support_organisation]'
```

### DIFF
--- a/app/services/document_reslugger.rb
+++ b/app/services/document_reslugger.rb
@@ -1,0 +1,95 @@
+class DocumentReslugger
+  Report = Struct.new(:published, :drafted, :skipped, :failed) do
+    def to_s
+      {
+        "Published" => published,
+        "Drafted" => drafted,
+        "Skipped" => skipped,
+        "Failed" => failed,
+      }.map { |label, paths| ["#{label} #{paths.size}:", *paths].join("\n") }.join("\n")
+    end
+  end
+
+  def initialize(document_type, finder_base_path)
+    @document_type = document_type
+    @finder_base_path = finder_base_path
+    @published = []
+    @drafted = []
+    @skipped = []
+    @failed = []
+  end
+
+  def reslug_all
+    klass.find_each do |document|
+      reslug(document)
+    end
+
+    Report.new(@published, @drafted, @skipped, @failed)
+  end
+
+private
+
+  def klass
+    @klass ||= begin
+      Rails.application.eager_load!
+      Document.subclasses.find { |subclass| subclass.document_type == @document_type } ||
+        raise(ArgumentError, "Unknown document_type: #{@document_type}")
+    end
+  end
+
+  # Reslugs a single document according to its state:
+  #   - already under the finder's base_path -> skipped
+  #   - unpublished -> skipped
+  #   - published -> moved and re-published, so the old path redirects
+  #   - draft (incl. published-with-new-draft) -> moved, kept in draft
+  #   - a failed save/publish or unexpected error -> recorded as a failure
+  def reslug(document)
+    if should_reslug?(document)
+      new_base_path = target_base_path(document)
+      was_reslugged = move_to_new_base_path(document, new_base_path)
+
+      if was_reslugged
+        outcome_for(document) << new_base_path
+      else
+        record_failure(document)
+      end
+    else
+      skip(document)
+    end
+  rescue StandardError => e
+    @failed << "#{document.content_id} (#{document.locale}) - #{e.message}"
+  end
+
+  def should_reslug?(document)
+    !already_reslugged?(document) && (document.published? || document.draft?)
+  end
+
+  def already_reslugged?(document)
+    target = target_base_path(document)
+    target.nil? || target == document.base_path
+  end
+
+  def move_to_new_base_path(document, new_base_path)
+    document.base_path = new_base_path
+    document.update_type = "minor"
+    document.save && (document.draft? || document.publish)
+  end
+
+  def outcome_for(document)
+    document.published? ? @published : @drafted
+  end
+
+  def target_base_path(document)
+    return unless document.base_path
+
+    "#{@finder_base_path}/#{File.basename(document.base_path)}"
+  end
+
+  def skip(document)
+    @skipped << "#{document.base_path} (#{document.publication_state})"
+  end
+
+  def record_failure(document)
+    @failed << "#{document.content_id} (#{document.locale}) - #{document.errors.full_messages.join(', ')}"
+  end
+end
--- a/lib/tasks/base_path.rake
+++ b/lib/tasks/base_path.rake
@@ -20,4 +20,55 @@ namespace :base_path do
       puts "Published #{document.base_path}"
     end
   end
+
+  # Reslugs every document of a type so it sits under its finder's live base_path:
+  # published documents are moved and re-published (the old path then redirects),
+  # drafts are moved but kept in draft, and unpublished documents or ones already
+  # under the finder's base_path are skipped. It prints a report of each outcome.
+  #
+  # Only runs when the finder is published to the live stack at the base_path in
+  # its schema, so deploy the schema change and run publishing_api:publish_finder
+  # first.
+  desc "Reslugs a type's documents under its finder's live base_path (published redirect, drafts stay draft)"
+  task :edit_all, %i[document_type] => %i[environment] do |_t, args|
+    document_type = args[:document_type]
+    shell = Thor::Shell::Basic.new
+    finder_schema = FinderSchema.load_from_schema(document_type.pluralize)
+    schema_base_path = finder_schema.base_path
+
+    begin
+      finder = Services.publishing_api.get_content(finder_schema.content_id).to_h
+    rescue GdsApi::HTTPNotFound
+      shell.say_error "The #{document_type} finder does not exist."
+      next
+    end
+
+    unless finder["publication_state"] == "published"
+      shell.say_error "The #{document_type} finder is not published to the live stack " \
+                      "(currently #{finder['publication_state']}); this task only reslugs " \
+                      "documents for finders on the live stack."
+      next
+    end
+
+    unless finder["base_path"] == schema_base_path
+      shell.say_error "The #{document_type} finder must be published at #{schema_base_path} " \
+                      "before reslugging its documents (currently published at #{finder['base_path']})."
+      next
+    end
+
+    message = <<~CONFIRMATION
+      This will reslug every #{document_type} document to sit under #{schema_base_path}.
+      Published documents are moved and re-published (redirecting the old path);
+      documents with a draft are moved but kept in draft.
+    CONFIRMATION
+
+    unless shell.yes?("#{message}Proceed? (yes/no)")
+      shell.say_error "Aborted"
+      next
+    end
+
+    report = DocumentReslugger.new(document_type, schema_base_path).reslug_all
+
+    puts report
+  end
 end
--- a/spec/lib/tasks/base_path_spec.rb
+++ b/spec/lib/tasks/base_path_spec.rb
@@ -82,4 +82,92 @@ RSpec.describe "base_path rake tasks", type: :task do
       expect { task.invoke(content_id, locale, new_base_path, "true") }.to raise_error(SystemExit)
     end
   end
+
+  describe "base_path:edit_all" do
+    let(:output) { StringIO.new }
+    let(:task) { Rake::Task["base_path:edit_all"] }
+    let(:document_type) { "veterans_support_organisation" }
+    let(:schema_base_path) { "/veteran-support-organisations" }
+    let(:finder_content_id) { "finder-content-id" }
+    let(:shell) { instance_double(Thor::Shell::Basic, yes?: true, say_error: nil) }
+    let(:publishing_api) { double("publishing_api") }
+    let(:report) do
+      DocumentReslugger::Report.new(["/veteran-support-organisations/foo"], [], [], [])
+    end
+    let(:reslugger) { instance_double(DocumentReslugger, reslug_all: report) }
+
+    before do
+      $stdout = output
+      task.reenable
+      allow(Thor::Shell::Basic).to receive(:new).and_return(shell)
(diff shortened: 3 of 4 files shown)


lib/tasks/base_path.rake
```diff
@@ -20,4 +22,82 @@ namespace :base_path do
...
+      if publish
+        published << document_base_path
+      else
+        drafted << document_base_path
+      end
+    rescue StandardError => e
+      failed << "#{content_id} (#{locale}) #{base_path} - #{e.message}"
+    end
+
+    puts "Published #{published.size}:"
```
lauraghiorghisor-tw: Not too bothered, but is it in any way more efficient/beneficial to have these printed at the end and not while the thing is running? If you started seeing failures, would you ever try to exit early? Not sure.

lib/tasks/base_path.rake
```diff
@@ -20,4 +22,82 @@ namespace :base_path do
...
+  # Reslugs every document of a type so it sits under its finder's current base_path.
+  # The target base_path is read from the finder schema, so deploy the schema change
+  # (and publish the finder) before running this.
+  desc "Reslugs every document of a type to match its finder's current base_path"
+  task :edit_all, %i[document_type] => %i[environment] do |_t, args|
+    Rails.application.eager_load!
+    klass = Document.subclasses.find { |subclass| subclass.document_type == args[:document_type] }
+    raise ArgumentError, "Unknown document_type: #{args[:document_type]}" unless klass
+
+    new_base_path = klass.finder_schema.base_path
```
lauraghiorghisor-tw: Can we use thor (I think we already have it in this repo) to print out a message - basically what you put in that comment on the file - and force the user to acknowledge y/n before proceeding - it feels like a risky enough rake task.

lauraghiorghisor-tw: Can we call this finder base path

spec/lib/tasks/base_path_spec.rb
```diff
@@ -82,4 +82,154 @@
...
+      allow(Document).to receive(:find).with("doc-3", "en").and_return(document)
+
+      task.invoke(document_type)
+
+      expect(document).to have_received(:base_path=).with("#{new_base_path}/baz")
+      expect(document).to have_received(:save)
+      expect(document).not_to have_received(:publish)
+    end
+
+    it "skips an unpublished document" do
```
lauraghiorghisor-tw: Just to check then - we don't reslug unpublished. So when a user re-drafts an unpublished doc, the base_path method in the presenter will still return draft? true (though that method is called first draft which this is not, but whatever), and thus use the new finder base path, and it will be fine?

spec/lib/tasks/base_path_spec.rb
```diff
@@ -82,4 +82,154 @@
...
+      task.reenable
+      allow(Rails.application).to receive(:eager_load!)
+      allow(Document).to receive(:subclasses).and_return([klass])
+      allow(Services).to receive(:publishing_api).and_return(publishing_api)
+      allow(klass).to receive(:finder_schema).and_return(double(base_path: new_base_path))
+    end
+
+    after { $stdout = STDOUT }
+
+    def stub_summary(content_id:, base_path:, state_history:)
```
lauraghiorghisor-tw: Is this not a stub document?  Why call it summary? Call it pub_api_document or payload or even document summary, if you need to distinguish it from the in memory document build, if I am getting this right?

spec/lib/tasks/base_path_spec.rb
```diff
@@ -82,4 +82,154 @@
...
+      summary = instance_double(Document, content_id:, locale: "en")
+      allow(publishing_api).to receive(:get_content).with(content_id, locale: "en").and_return(
+        double(to_h: { "base_path" => base_path, "state_history" => state_history }),
+      )
+      summary
+    end
+
+    def editable_document(base_path:)
```
lauraghiorghisor-tw: This is the the "built"/in memory?/edit**ed**/submitted/about to be written document? 

Not sure what to name this.

spec/lib/tasks/base_path_spec.rb
```diff
@@ -82,4 +82,154 @@
...
+
+      task.invoke(document_type)
+
+      expect(document).to have_received(:base_path=).with("#{new_base_path}/baz")
+      expect(document).to have_received(:save)
+      expect(document).not_to have_received(:publish)
+    end
+
+    it "skips an unpublished document" do
+      summary = stub_summary(content_id: "doc-4", base_path: "#{old_base_path}/qux", state_history: { "1" => "published", "2" => "unpublished" })
```
lauraghiorghisor-tw: Just to be pedantic, the first state should likely be superseded when the second is unpublished

spec/lib/tasks/base_path_spec.rb
```diff
@@ -82,4 +82,154 @@
...
+      task.invoke(document_type)
+
+      expect(document).to have_received(:base_path=).with("#{new_base_path}/foo")
+      expect(document).to have_received(:update_type=).with("minor")
+      expect(document).to have_received(:save)
+      expect(document).to have_received(:publish)
+    end
+
+    it "reslugs but keeps a draft document in draft" do
+      summary = stub_summary(content_id: "doc-2", base_path: "#{old_base_path}/bar", state_history: { "1" => "draft" })
```
lauraghiorghisor-tw: Name of summary in tests could be changed.

lauraghiorghisor-tw (approved):
A lot of logic in here, unfortunately. LGTM with some minor changes and comments.

TonyGDS:
Reworked since the last review.

lib/tasks/base_path.rake
TonyGDS: I felt probably better to let it run, it's idempotent so safe to re-run

spec/lib/tasks/base_path_spec.rb
TonyGDS: see my comment after the re-write

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    @skipped = []
+    @failed = []
+  end
+
+  def finder_base_path
+    @finder_base_path ||= live_finder_base_path
+  end
+
+  def call
+    klass.find_each do |listed_document|
```
lauraghiorghisor-tw: What is "listed" 😅 These are the result of Document.build at this point?

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    @klass ||= begin
+      Rails.application.eager_load!
+      Document.subclasses.find { |subclass| subclass.document_type == document_type } ||
+        raise(ArgumentError, "Unknown document_type: #{document_type}")
+    end
+  end
+
+  def live_finder_base_path
+    finder = Services.publishing_api.get_content(klass.finder_schema.content_id).to_h
+    return finder["base_path"] if finder["publication_state"] == "published"
```
lauraghiorghisor-tw: Just to check, if the finder is in draft, then we assume all docs are also in draft so any changes to paths should apply without issue? You "republish" the finder (when in draft that means you're pushing it to draft stack only) + run the republish docs of type and it works? (I think we've also discovered that just republishing the finder fixes children due to pub API magic, but not 100% sure)

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+  rescue StandardError => e
+    failed << "#{content_id} (#{locale}) - #{e.message}"
+  end
+
+  def already_reslugged?(document)
+    target = target_base_path(document)
+    target.nil? || target == document.base_path
+  end
+
+  def move(document)
```
lauraghiorghisor-tw: "move", really 😅

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    publish = document.published?
```
lauraghiorghisor-tw: let's call this should_publish

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    old_base_path = document.base_path
```
lauraghiorghisor-tw: There is smth about this code here that got me confused yesterday as well, my brain went a bit in circles. 

Because all the logic flavours in the `base_path` method still apply, right?
When the presenter logic runs, this runs:
```
  def base_path
    if first_draft?
      @base_path = "#{finder_schema.base_path}/#{title.to_url}"
    else
      @base_path
    end
    truncate(@base_path, length: 250, omission: "")
  end
```

The whole thing is a bit compounded. 
So first drafts, even though you might "set"/assign their base path here, they still send their finder base path in the presenter, which is the same in this case.

Then everything else uses the in memory @base_path you just assigned, and that is what gets sent downstream. And even if we might have tried to set a longer base_path above, we end up truncating it. 

So we're basically pushing a new base_path in all cases now, regardless of state.

Is my understanding correct?

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    new_base_path = target_base_path(document)
+    document.base_path = new_base_path
+    document.update_type = "minor"
+
+    return record_failure(document, old_base_path) unless document.save
```
lauraghiorghisor-tw: This sequence of save - publish might read better in a try catch block (_pedantic_).

app/services/document_reslugger.rb
```diff
@@ -0,0 +1,91 @@
...
+    return record_failure(document, old_base_path) if publish && !document.publish
+
+    (publish ? published : drafted) << new_base_path
+  end
+
+  def target_base_path(document)
```
lauraghiorghisor-tw, line 82: The behaviour is a normal redirect in pub-api? Old urls of docs will redirect to new?

spec/services/document_reslugger_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+      listed = instance_double(Document, content_id: "doc-7", locale: "en")
+      allow(Document).to receive(:find).with("doc-7", "en").and_raise(StandardError, "boom")
+      allow(klass).to receive(:find_each).and_yield(listed)
+
+      reslugger.call
+
+      expect(reslugger.failed).to eq(["doc-7 (en) - boom"])
+    end
+
+    it "returns itself so the caller can read the results" do
```
lauraghiorghisor-tw: Oh, I see why it returned self. Not sure if that's a common pattern, it feels like you're potentially exposing the instance unnecessarily, when you can just compose a return payload.

lauraghiorghisor-tw (approved):

app/services/document_reslugger.rb
TonyGDS: Got rid of this, the `find_each` contains all we need so can call the variable `document`

app/services/document_reslugger.rb
TonyGDS: The task won't run against a draft finder by design, it now does the finder check itself and stops unless the finder is published on the live stack at the schema base_path.

> I think we've also discovered that just republishing the finder fixes children due to pub API magic, but not 100% sure

No, we need to run this task once the updated finder has been published

app/services/document_reslugger.rb
TonyGDS: `move_to_new_base_path` now, in the case of a draft it just saves, but for a published document it will republish. I don't want to overuse the `reslug` naming.

app/services/document_reslugger.rb
TonyGDS: that's gone now

app/services/document_reslugger.rb
TonyGDS: Genuine first drafts are actually skipped in this task. For a `first_draft?` doc the `base_path` getter recomputes `finder_schema.base_path/title.to_url` on every read, so `already_reslugged?` sees `target == document.base_path`

app/services/document_reslugger.rb
TonyGDS: save/publish return `false` and set errors on the document rather than raising so we can't handle with exceptions

spec/services/document_reslugger_spec.rb
TonyGDS: This now returns a Report value object. Previous design was using a command pattern style, but not as nice tbf.

app/services/document_reslugger.rb
TonyGDS: Yes:
- Published docs we re-publish → redirect goes live                                                                                                                                                                                                                                                                                                    
- Published-with-new-draft we keep in draft → creates the redirect, but as a draft, so you'll see it on the draft preview while the old path stays live until the new draft is published.                                                                                                                                        
- Never-published drafts / first drafts → the doc just moves to the new path.

app/services/document_reslugger.rb
lauraghiorghisor-tw: Sorry I meant for draft finders.

lauraghiorghisor-tw (approved):

PR #3827, opened 2026-08-25
Remove DocumentDB dependency [WHIT-3578]

ChrisBAshton:
## What

Removes Specialist Publisher's dependency on Mongo / DocumentDB, using an in-memory solution instead via some custom middleware. This PR swaps out the logic to avoid writing to a Users table, instead having User as a plain in-memory object that implements the exact interface that gds-sso requires. It stores the full user snapshot (uid, name, permissions, org) inside the same signed cookie rather than a database row.

This draft PR has some knock-on effects:

1. Signon has two push callbacks (`PUT /auth/gds/api/users/:uid` for instant permission updates, and `POST /auth/gds/api/users/:uid/reauth` for forced sign-out). These have no database to write to anymore, and can no longer authenticate their own bearer token, so now would silently fail closed, meaning a user's permissions or disabled state would now only take effect when they next sign-in or after the `auth_valid_for` expires ([20 hours](https://github.com/alphagov/gds-sso/commit/8733067b64342b74397e498dd6648634f3b7314a)). This is potentially a blocker for merge.
2. We lose the `govuk:user-created-at` GA4 metadata, since there is no local Users table anymore so no way of checking the creation date of the user. It's unclear how important that is.

Sister PR: https://github.com/alphagov/govuk-docker/pull/1030

## Why

Specialist Publisher is on a shared DocumentDB instance, whose version needs updating - made complicated by the fact it's shared with multiple tenants. [RFC-143](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md) agreed that all apps should have their own dedicated database instances. If we do nothing else, the current intention is to spin up a new DocumentDB instance, point Specialist Publisher to that, and then upgrade DocumentDB in place.

However: Specialist Publisher is unique insofar as it doesn't need a database for any of its data - it stores everything in Publishing API. Nevertheless, it currently needs a database for the gds-sso integration, which expects a database and a Users table. So in theory, migrating to a more well known database such as Postgres should be fairly straightforward.

That said, it seems a shame to have to spin up and pay for a database when we're using it so little and where the data is relatively disposable. This PR explores the pros and cons of taking this approach.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3578

---


ChrisBAshton:
This branch went well and it worked, but led to an unacceptable loss of security: the ability for us to remotely sign out users from Signon.

There are ways around it - building a small cache in Redis to track per-user state, writing to it from two different endpoints, treating that as the source of truth ahead of the cookie. By this point we’ve effectively built a lightweight database ourselves. So in conclusion I think the small infra cost of a MySQL / PostgreSQL database is worth it, and should lead to a much lighter swap-out of code in Specialist Publisher without any of the drawbacks.

PR #3828, opened 2026-08-26
Migrate from Mongo to MySQL [WHIT-3578]

ChrisBAshton:
## What

Migrates Specialist Publisher from Mongo to MySQL. No data migration is required since the database is only used for  a Users table required by the gds-sso gem (for Signon integration).

## Why

Specialist Publisher's Mongo database is in a shared DocumentDB instance, in contravention of [RFC-143](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-143-split-database-instances.md) which agrees that every app should have its own dedicated database instance.

The DocumentDB version requires upgrading, which is made difficult as it is multi-tenanted. We're using the opportunity to move to a different database engine altogether, rather than spin up another DocumentDB instance and have to unpick which data is relevant to Specialist Publisher compared to the other tenants.

Whilst Postgres tends to be the most used engine on GOV.UK, we've opted for MySQL here because Specialist Publisher is ultimately intended to be consolidated into Whitehall, which itself uses MySQL.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3578

---


leenagupte (approved):
Tried this out locally alongside https://github.com/alphagov/govuk-docker/pull/1031 and on integration.
Works great! 🥇

